PR #3, opened 2025-04-23
Imported trainer sdk from kubeflow/trainer

szaher:


szaher:
closing this one as it didn't honor git history.

PR #15, opened 2025-06-05
Consume Trainer models from external package kubeflow_trainer_api

### LINKED ISSUE #9: Consume Trainer models from external package

As we plan to have kubeflow components manage their own models independently, we need to update SDK to import models from `kubeflow_trainer_api` instead of maintaining its own copies.
- Initially import models from trainer repo as a git dependency if not yet on PyPi
- Remove duplicate model definitions from SDK repo
- Update all references to use imported models

### Why is this needed?

Currently, SDK repo contains duplicate model definitions that will become out of sync with the trainer repo when APIs change.

### Love this feature?

kramaranya:
This PR replaces duplicate model definitions in the SDK with imports from the `kubeflow_trainer_api` package.
Added `kubeflow_trainer_api` as a git dependency since it's not yet published to PyPI and removed duplicate models.

Fixes #9 

**Checklist:**

### DIFF
--- a/python/kubeflow/trainer/api/trainer_client.py
+++ b/python/kubeflow/trainer/api/trainer_client.py
@@ -20,7 +20,7 @@ import string
 import uuid
 from typing import Dict, List, Optional
 
-import kubeflow_trainer_api.models as models
+from kubeflow_trainer_api import models
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
 from kubeflow.trainer.utils import utils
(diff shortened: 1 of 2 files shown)


python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -20,7 +20,7 @@
 import uuid
 from typing import Dict, List, Optional
 
-import kubeflow.trainer.models as models
+import kubeflow_trainer_api.models as models
```
eoinfennessy: I don't think there's any need to create an alias here.

```suggestion
from kubeflow_trainer_api import models
```

And the same suggestion for other imports.

eoinfennessy (approved):
LGTM

andreyvelich (commented):
Thank you for this @kramaranya!
/lgtm
/approve

PR #28, opened 2025-06-17
fix: Expose BuiltinTrainer API to users

Electronic-Waste:

This PR exposes the `BuiltinTrainer` API to users, and also resolve some bugs:

1. Missing retrieving logic of `ACCESS_TOKEN`
2. Wrong torchtune config overrides

And the testing scripts and results can be found at: https://github.com/kubeflow/trainer/pull/2675

/cc @kubeflow/wg-training-leads @astefanutti 

Fixes #

**Checklist:**

### DIFF
--- a/python/kubeflow/trainer/api/trainer_client.py
+++ b/python/kubeflow/trainer/api/trainer_client.py
@@ -18,7 +18,7 @@ import queue
 import random
 import string
 import uuid
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Union
 
 from kubeflow_trainer_api import models
 from kubeflow.trainer.constants import constants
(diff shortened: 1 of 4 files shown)


python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -153,20 +153,22 @@ def train(
...
+            the post-training logic, requiring only parameter adjustments, e.g. `BuiltinTrainer`.
 
         Args:
             runtime (`types.Runtime`): Reference to one of existing Runtimes.
             initializer (`Optional[types.Initializer]`):
                 Configuration for the dataset and model initializers.
-            trainer (`Optional[types.CustomTrainer]`):
-                Configuration for Custom Training Task.
+            trainer (`Optional[types.CustomTrainer, types.BuiltinTrainer]`):
+                Configuration for Custom Training Task or Config-driven Task with Existing Trainer.
```
andreyvelich: Shall we call it builtin Trainer ?

```suggestion
                Configuration for Custom Training Task or Config-driven Task with Builtin Trainer.
```

python/kubeflow/trainer/utils/utils.py
```diff
@@ -493,7 +498,13 @@ def get_dataset_initializer(
             dataset.storage_uri
             if dataset.storage_uri.startswith("hf://")
             else "hf://" + dataset.storage_uri
-        )
+        ),
+        env=[
+            models.IoK8sApiCoreV1EnvVar(
+                name=constants.INITIALIZER_ENV_ACCESS_TOKEN,
+                value=dataset.access_token,
```
andreyvelich, line 505: What if users don't set `access_token` ? Will it throw exception ?

Electronic-Waste: Will fail to download models/datasets that need `access_token`. But it won't throw exception. I discovered this error because the model-initializer pod failed to download Llama-3.2-1B-Instruct.

Electronic-Waste:
Hey, I've addressed your comments. Is there any other suggestions? @andreyvelich 

/cc @kubeflow/wg-training-leads @astefanutti

python/kubeflow/trainer/utils/utils.py
andreyvelich: I just meant that if in the future we will support models that don't require access_token (e.g. Qwen3), and user doesn't set the `dataset.access_token`, it will set env `ACCESS_TOKEN=None`.
Should we just add this env for Dataset and Model initializer if `dataset.access_token != None` ?

Electronic-Waste: Sure! I've updated the code. Does it look good to you now? @andreyvelich

andreyvelich:
Thank you @Electronic-Waste!
/lgtm
/approve

Electronic-Waste:
@andreyvelich Thanks for your detailed review and quick response. I'll start working on https://github.com/kubeflow/trainer/pull/2670#discussion_r2175009367 soon:)

Issue #29, opened 2025-06-18
Support custom Docker images for MPI training runtimes


jskswamy:
### What happened?

The SDK currently has a hardcoded mapping of Docker image names to trainer configurations in `ALL_TRAINERS`. When using custom Docker images that aren't in this predefined mapping, the SDK falls back to `DEFAULT_TRAINER` which uses `torchrun` instead of `mpirun`, causing MPI training jobs to fail.

**Root Cause:**
In `sdk/python/kubeflow/trainer/utils/utils.py`:
```python
# Extract image name from the container image to get appropriate Trainer.
image_name = trainer_container.image.split(":")[0]
trainer = types.ALL_TRAINERS.get(image_name, types.DEFAULT_TRAINER)
```

The `ALL_TRAINERS` dictionary in `types.py` only includes predefined image names like:
- `ghcr.io/kubeflow/trainer/deepspeed-runtime`
- `ghcr.io/kubeflow/trainer/mlx-runtime`
- `pytorch/pytorch`

When using custom images like `custom-registry.example.com/my-deepspeed-runtime`, the SDK falls back to `DEFAULT_TRAINER` which uses `torchrun` entrypoint instead of `mpirun`.

**Current Behavior:**
- Custom images are not recognized in the predefined mapping
- SDK falls back to `DEFAULT_TRAINER` with `torchrun` entrypoint
- MPI training commands are dropped from the generated TrainJob
- Distributed training fails because `mpirun` is missing

**Additional Context:**
There's already a TODO comment in the code suggesting this should be addressed:
```python
# TODO (andreyvelich): We should allow user to overrides the default image names.
```

**Proposed Solutions:**
1. **Add configuration support**: Allow users to specify custom image-to-trainer mappings
2. **Image pattern matching**: Support wildcard/regex patterns for image names
3. **Runtime-based detection**: Use runtime metadata to determine trainer type instead of relying solely on image names
4. **Fallback logic**: Improve fallback behavior for unrecognized images

This affects users who need to use custom Docker images for their training workloads, particularly in enterprise environments where standard images may not be available or suitable.

### What did you expect to happen?

- Custom images should be properly detected for MPI training
- The SDK should support configurable image-to-trainer mappings
- MPI entrypoints should be preserved for custom DeepSpeed images
- The generated TrainJob should include the correct `mpirun` command regardless of image name

### Environment

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

Name: kubeflow
[...]

Issue #20, opened 2025-02-28
Explore `uv` project manager for Kubeflow Python SDK


andreyvelich:
### What you would like to be added?

We should explore the uv project manager for the Kubeflow Python SDK.
It is faster than other tools, and many Python libraries have started adopting it.

In particular, I am interested in the Workspace concept and whether it can help manage multiple libraries under the unified Kubeflow SDK.
https://docs.astral.sh/uv/concepts/projects/workspaces/

Currently, we use `hatch` for [the Kubeflow SDK](https://github.com/kubeflow/trainer/blob/master/sdk/pyproject.toml#L42-L49).

/cc @kubeflow/wg-training-leads @Electronic-Waste @shravan-achar @akshaychitneni @astefanutti @franciscojavierarceo @seanlaii @spolti 

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

briangallagher:
@andreyvelich Can we proceed with the use of UV? 

I've investigated this also, similar to Saad project. See project [here](https://github.com/briangallagher/uvworkspacetest/tree/main) 

**Proposal** (as demonstrated in repo above)
Use uv to provide:
- separate workspace per sub component
- individual project.toml files
- single uv.lock file 
And Use optional dependancies to allow a subset of components to be installed e.g. pip install sdk[trainer,katib]

The issue that Saad encountered i.e. having to replicate dependancies at the root level can be overcome if the optional dependancies are resolvable/published to pypi. Again, you can see this is my repo.  I think this would be preferable over duplicating dependancies in the root toml file? Possibly not ideal to have all sub components published but we can clarify preferred usage of the sdk through documentation. 

What do you think, If ok with this I can implement?

andreyvelich:
We've migrated to `uv` for Kubeflow SDK 🎉

PR #17, opened 2025-06-06
add unit test for trainer sdk

briangallagher:
**What this PR does / why we need it**:
Add some initial unit tests for trainer_client.py

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #16 

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

PR #31, opened 2025-06-18
Fix trainer detection for custom Docker images with regex pattern matching

jskswamy:
## Problem
Custom DeepSpeed Docker images were losing the `mpirun` command and falling back to `torchrun` instead. The `get_runtime_trainer` function only used a hardcoded `ALL_TRAINERS` mapping, so any custom images not in this mapping would default to the PyTorch trainer configuration.

## Solution
Enhanced trainer detection with regex-based pattern matching as a fallback mechanism:

1. **First priority**: Check existing `ALL_TRAINERS` mapping for exact matches
2. **Second priority**: Use regex patterns to detect framework from image names:
   - DeepSpeed: `(?i)deepspeed` (case-insensitive)
   - MLX: `(?i)mlx` 
   - TorchTune: `(?i)torchtune`
   - PyTorch: `(?i)(pytorch|torch)` (but not torchtune)
3. **Fallback**: Default to PyTorch trainer if no patterns match

## Key Changes
- Added `_detect_trainer_from_image_patterns()` function with case-insensitive regex matching
- Modified `_detect_trainer()` to use pattern matching as fallback
- Added `copy.deepcopy()` to prevent shared state issues between trainer configurations
- Comprehensive test suite with 76 test cases covering various image name formats

## Testing
- ✅ Known images from `ALL_TRAINERS` mapping
- ✅ Custom images with various case formats (lowercase, uppercase, mixed case)
- ✅ Images with registry prefixes, ports, and complex paths
- ✅ Edge cases and fallback scenarios
- ✅ Accelerator count logic with ML policies
- ✅ State isolation between test runs

This ensures custom DeepSpeed images like `my-org/deepspeed-custom:latest` correctly use `mpirun` instead of falling back to `torchrun`.

This fixes the issue #29


python/tests/test_utils.py
```diff
@@ -0,0 +1,403 @@
...
+    ):
+        """Test trainer detection using image pattern matching with various case scenarios."""
+        trainer = utils._detect_trainer_from_image_patterns(image_name)
+        if expected_framework == types.TrainerFramework.TORCH and trainer is None:
+            # For unknown images, the _detect_trainer function should return default
+            # but _detect_trainer_from_image_patterns returns None
+            assert trainer is None
+        else:
+            assert trainer is not None
+            assert trainer.framework.value == expected_framework.value
```
eoinfennessy: Small suggestion to replace `expected_framework` with `None` for no-match cases. I think this make the behavior of the function being tested clearer for readers.

```suggestion
            # Edge cases - no match
            ("unknown-image:latest", None),
            ("", None),
            ("nginx:latest", None),
            ("ubuntu:20.04", None),
        ],
    )
    def test_trainer_detection_from_image_patterns(
        self, image_name, expected_framework
    ):
        """Test trainer detection using image pattern matching with various case scenarios."""
        trainer = utils._detect_trainer_from_image_patterns(image_name)
        if expected_framework is None:
            # For unknown images _detect_trainer_from_image_patterns returns None
            assert trainer is None
        else:
            assert trainer is not None
            assert trainer.framework.value == expected_framework.value
```

python/kubeflow/trainer/types/types.py
```diff
@@ -21,6 +21,15 @@
 from kubeflow.trainer.constants import constants
 
 
+# Trainer framework constants for easy reference
+class TrainerFramework(Enum):
+    """Trainer framework constants."""
+    TORCH = "torch"
+    DEEPSPEED = "deepspeed"
+    MLX = "mlx"
+    TORCHTUNE = "torchtune"
```
eoinfennessy: This is exactly the same as the `Framework` enum. Can we delete this and use `Framework` instead?

eoinfennessy (commented):
@jskswamy thank you for this contribution! And thank you for also adding tests for `get_container_devices`.

It looks great to me -- I just added two small suggestions.

eoinfennessy (approved):
@jskswamy LGTM! Thank you!

python/uv.lock
```diff
@@ -0,0 +1,438 @@
+version = 1
```
andreyvelich: Please can we add support for `uv` in the separate PR ? 
We have an open issue for it: https://github.com/kubeflow/sdk/issues/20
cc @briangallagher @szaher

python/tests/test_utils.py
```diff
@@ -0,0 +1,397 @@
+import pytest
+from unittest.mock import Mock, patch
+from kubeflow.trainer.utils import utils
+from kubeflow.trainer.types import types
+from kubeflow.trainer.constants import constants
+
+
+class TestTrainerDetection:
```
andreyvelich: Thank you for adding unit tests!
Please can you move them under `kubeflow/trainer/types/types_test.py`, similar to our client tests: https://github.com/kubeflow/sdk/pull/17

Previously, we agreed to follow that folder structure.

python/kubeflow/trainer/utils/utils.py
```diff
@@ -121,20 +190,19 @@ def get_runtime_trainer(
...
         _, trainer.accelerator_count = devices
 
     # Torch and MPI plugins override accelerator count.
-    if ml_policy.torch and ml_policy.torch.num_proc_per_node:
+    if ml_policy.torch and ml_policy.torch.num_proc_per_node is not None:
         num_proc = ml_policy.torch.num_proc_per_node.actual_instance
         if isinstance(num_proc, int):
             trainer.accelerator_count = num_proc
-    elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node:
+    elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:
```
andreyvelich: Why do we need to add `is not None` here ?

python/kubeflow/trainer/utils/utils.py
```diff
@@ -121,20 +190,19 @@ def get_runtime_trainer(
     if not (trainer_container and trainer_container.image):
         raise Exception(f"Runtime doesn't have trainer container {replicated_jobs}")
 
-    # Extract image name from the container image to get appropriate Trainer.
-    image_name = trainer_container.image.split(":")[0]
-    trainer = types.ALL_TRAINERS.get(image_name, types.DEFAULT_TRAINER)
+    # Use the new detection logic with fallback
+    trainer = _detect_trainer(trainer_container)
```
andreyvelich: Do we really need to use `_` prefix for the util functions ?
E.g. we have many other internal util functions where we don't have such prefix.

python/kubeflow/trainer/utils/utils.py
```diff
@@ -107,6 +108,74 @@ def get_runtime_trainer_container(
...
+        trainer_container: The trainer container object
+
+    Returns:
+        Trainer object
+    """
+    image_name = trainer_container.image.split(":")[0]
+
+    # 1. Check existing ALL_TRAINERS mapping (backward compatibility)
+    if image_name in types.ALL_TRAINERS:
+        return copy.deepcopy(types.ALL_TRAINERS[image_name])
```
andreyvelich: Do we really need to maintain ALL_TRAINERS list if we are going to detect ML Framework based on regex ?

python/kubeflow/trainer/utils/utils.py
```diff
@@ -107,6 +108,74 @@ def get_runtime_trainer_container(
...
+    if re.search(r"torchtune", image_name, re.IGNORECASE):
+        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.TORCHTUNE])
+
+    # PyTorch patterns (more specific to avoid matching torchtune)
+    if re.search(r"pytorch", image_name, re.IGNORECASE):
+        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.TORCH])
+
+    # Generic torch patterns (but not torchtune)
+    if re.search(r"^torch(?!tune)", image_name, re.IGNORECASE):
+        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.TORCH])
```
andreyvelich: I think, we can agree that PyTorch based runtime must contain `pytorch` in the image name.
In that case, we can distinguish `torchtune` and `pytorch` images.
```suggestion
```

python/kubeflow/trainer/utils/utils.py
```diff
@@ -107,6 +108,74 @@ def get_runtime_trainer_container(
     return None
 
 
+def _detect_trainer_from_image_patterns(image_name: str) -> Optional[types.Trainer]:
```
andreyvelich: @tenzen-y @Electronic-Waste @astefanutti @jskswamy @eoinfennessy @franciscojavierarceo Do we see any concerns with regex approach ?  It might be a good and simple method to start with, but I can imagine use cases where it wouldn't work. For example, users might have two DeepSpeed runtimes:

- One uses `torchrun` 
- Another uses `mpirun`.

Perhaps in the future we can support such scenarios.

andreyvelich (commented):
Thank you for this contribution @jskswamy 🎉

python/uv.lock
jskswamy: done, here is the new PR for the same https://github.com/kubeflow/sdk/pull/35

python/tests/test_utils.py
jskswamy: Moved the files to the corresponding locations

python/kubeflow/trainer/utils/utils.py
jskswamy: #### 1. **Torch Policy Check** (`if trainer_container.accelerator_count is not None`)
```python
# Essential: Prevents AttributeError when accessing None.actual_instance
# Without this check: None.actual_instance would raise AttributeError
if trainer_container.accelerator_count is not None:
    if hasattr(trainer_container.accelerator_count, 'actual_instance'):
        trainer.accelerator_count = trainer_container.accelerator_count.actual_instance
```

#### 2. **MPI Policy Check** (`if trainer_container.mpi_policy is not None`)
```python
# Essential: Prevents setting accelerator_count to None when user explicitly sets it
# Without this check: trainer.accelerator_count would be overwritten to None
if trainer_container.mpi_policy is not None:
    trainer.accelerator_count = trainer_container.mpi_policy.num_procs
```

#### 3. **Semantic Correctness**
These checks ensure that:
- User-provided values are preserved and not overwritten
- We don't attempt operations on `None` objects
- The logic follows "only apply changes if the field is actually set"

### Code Comments Added:
I've added explanatory comments to each check to make their necessity clear for future maintainers.

python/kubeflow/trainer/utils/utils.py
jskswamy: Taken care of by having a checking for `pytorch` explicitly

python/kubeflow/trainer/utils/utils.py
jskswamy: Removed the `ALL_TRAINERS` function and the ML Framework detection is now based on the regex

python/kubeflow/trainer/utils/utils.py
jskswamy: Removed the _ prefix for all the utils functions and adhering the convention followed

python/kubeflow/trainer/utils/utils.py
jskswamy: I agree, there are use cases where image name patterns alone wouldn't be sufficient.

### Current Regex Approach — Pragmatic Starting Point

The regex approach implemented serves as a **practical starting point** that:
- ✅ **Works immediately** for the majority of common use cases
- ✅ **Supports all official Kubeflow trainer images** out of the box
- ✅ **Provides sensible defaults** without requiring users to specify trainer types manually
- ✅ **Maintains backward compatibility** with existing workflows

### Future API-Based Enhancement

For advanced scenarios like your DeepSpeed example (torchrun vs mpirun variants), we can introduce **explicit API controls** that override the regex detection:

```python
# Option 1: Explicit trainer specification
trainer = Trainer(
    image="custom/deepspeed-runtime",
    trainer_type=TrainerType.DEEPSPEED_MPI,  # Override regex detection
    # ... other configs
)

# Option 2: Runtime configuration
trainer = Trainer(
    image="custom/deepspeed-runtime", 
    runtime_config=DeepSpeedConfig(launcher="mpirun"),  # vs "torchrun"
    # ... other configs
)
```

### Approach

The regex approach handles ~90% of use cases elegantly, while keeping the door open for API-based precision when needed.

**Question**: Which approach would you prefer to proceed with?

**Option A**: Keep the current regex-based detection and enhance it incrementally with API overrides when needed

**Option B**: Move to a more explicit API-first approach where users specify trainer types directly

**Option C**: Hybrid approach where regex provides defaults, but API allows explicit overrides from day one

I'm happy to implement the changes that would be most valuable for users. What are your thoughts?

python/kubeflow/trainer/utils/utils.py
andreyvelich: @jskswamy I just meant that those 2 lines are the same in Python, isn't ?
```py
elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node:
elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:
```

python/kubeflow/trainer/utils/utils.py
andreyvelich: Let's keep the regex approach initially, we can have discussion how to cover more complex use-cases.
We just need to ensure we document that in our docs: https://www.kubeflow.org/docs/components/trainer/user-guides/

python/kubeflow/trainer/types/types.py
```diff
@@ -230,19 +230,16 @@ class Initializer:
...
-# Each Trainer representation defines trainer parameters (e.g. type, framework, entrypoint).
-# TODO (andreyvelich): We should allow user to overrides the default image names.
-ALL_TRAINERS: Dict[str, Trainer] = {
-    # Custom Trainers.
-    "pytorch/pytorch": Trainer(
+# Centralized trainer configurations to eliminate duplication
+TRAINER_CONFIGS: Dict[Framework, Trainer] = {
+    Framework.TORCH: Trainer(
         trainer_type=TrainerType.CUSTOM_TRAINER,
         framework=Framework.TORCH,
```
andreyvelich, line 237: Do we really need to keep framework argument given that `TRAINER_CONFIGS` Dict has the Framework type in the Dict key.
```suggestion
```

python/kubeflow/trainer/utils/utils.py
```diff
@@ -107,6 +108,65 @@ def get_runtime_trainer_container(
...
+    """
+    image_name = trainer_container.image.split(":")[0]
+
+    # 1. Use image pattern matching
+    trainer = detect_trainer_from_image_patterns(image_name)
+    if trainer:
+        return trainer
+
+    # 2. Fall back to DEFAULT_TRAINER
+    return copy.deepcopy(types.DEFAULT_TRAINER)
```
andreyvelich: I think, this could be simplified if `detect_trainer_from_image_patterns` just return `copy.deepcopy(types.DEFAULT_TRAINER)` instead of None.

Can you just keep all of the required code to extract trainer in the `get_trainer_from_image()` function, which accepts `image_name` as input?

That will make our unit tests easier to maintain.

python/kubeflow/trainer/types/types.py
jskswamy: Regarding the `framework` field in the `Trainer` class, I'd like to share my thoughts on why this field exists and why it serves a legitimate purpose:

## The `framework` Field Has Critical Importance

After investigating the codebase, I discovered that the `Trainer` class and `framework` field were **pre-existing** before this PR. The field was intentionally designed to serve specific purposes:

### **Critical Importance for API Design**
The `framework` field is **essential** for maintaining a clean, self-contained API:

1. **Object Identity**: A `Trainer` object must "know" what framework it represents without external context
2. **API Completeness**: When users receive a `Trainer` object, they can immediately determine its framework without reverse-engineering from other fields
3. **Serialization**: The field is crucial for JSON serialization/deserialization of trainer objects
4. **Debugging & Logging**: Essential for meaningful error messages and debugging information

### **Self-Contained Data Structure**
The `framework` field makes `Trainer` objects self-contained and self-documenting:
```python
# Example: A Trainer object "knows" what framework it represents
trainer = TRAINER_CONFIGS[Framework.DEEPSPEED]

# Self-documenting: The object tells us what it is
print(f"Using {trainer.framework} trainer with {trainer.trainer_type}")
# Output: "Using Framework.DEEPSPEED trainer with TrainerType.CUSTOM_TRAINER"

# Without the field, we'd need external context to know what framework this is
# We'd have to track which dictionary key was used to create this trainer
```

### **Breaking Changes Would Be Required**
Removing the field would require:
- Modifying any code that relies on the field for framework identification
- Potentially breaking API consumers who expect this field
- Adding complex lookup logic to determine framework from other properties

### **Architectural Integrity**
The field maintains the principle of **encapsulation** — `Trainer` object should contain all information about itself, including what framework it represents.

## Why Dictionary Instead of Array?

The choice of using `TRAINER_CONFIGS: Dict[Framework, Trainer]` instead of an array of trainers was a **performance and design optimization**:

### **Performance Benefits**
```python
# Current efficient approach with dictionary
trainer = TRAINER_CONFIGS[Framework.DEEPSPEED]  # O(1) lookup
framework = trainer.framework  # Direct access

# Alternative inefficient approach with array
def find_trainer_by_framework(framework):
    for trainer in TRAINER_ARRAY:  # O(n) search
        if trainer.framework == framework:
            return trainer
```

### **Design Benefits**
1. **Fast Lookup**: O(1) constant time access instead of O(n) linear search
2. **Type Safety**: Dictionary keys ensure we only access valid frameworks
3. **Explicit Mapping**: Clear relationship between framework and trainer configuration
4. **Extensibility**: Easy to add new frameworks without changing lookup logic

## My Take
The `framework` field serves critical architectural purposes for API design and object encapsulation. The dictionary structure provides performance benefits, but the field itself is essential for maintaining clean, self-contained objects. 

Removing the field would break the original design intent, make the API less clean and efficient, and potentially introduce breaking changes. The field was intentionally designed this way for good reasons, and I believe we should keep it to maintain the integrity of the API design.

python/kubeflow/trainer/utils/utils.py
jskswamy: I think it's a subtle but important distinction in Python.

## The `is not None` Check is Necessary

The current code is correct because `0` is a **valid and meaningful value** for `num_proc_per_node`:

```python
# Current correct implementation
elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:
    trainer.accelerator_count = ml_policy.mpi.num_proc_per_node
```

## Why Truthiness Checking Would Break CPU-Only Training

If we used truthiness checking instead:

```python
# This would be problematic
elif ml_policy.mpi and ml_policy.mpi.num_proc_per_node:
    trainer.accelerator_count = ml_policy.mpi.num_proc_per_node
```

### **Example Scenarios:**

**Scenario 1: CPU-Only Training (0 accelerators)**
```python
ml_policy.mpi.num_proc_per_node = 0  # Explicitly set to CPU-only

# With truthiness check:
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node:  # 0 is falsy!
    trainer.accelerator_count = ml_policy.mpi.num_proc_per_node  # ❌ Never executes

# With is not None check:
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:  # 0 is not None!
    trainer.accelerator_count = ml_policy.mpi.num_proc_per_node  # ✅ Executes correctly
```

**Scenario 2: GPU Training (4 accelerators)**
```python
ml_policy.mpi.num_proc_per_node = 4  # Explicitly set to 4 GPUs

# Both approaches work correctly:
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node:  # 4 is truthy ✅
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:  # 4 is not None ✅
```

**Scenario 3: Not Set (defaults to UNKNOWN)**
```python
ml_policy.mpi.num_proc_per_node = None  # Not explicitly set

# Both approaches work correctly:
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node:  # None is falsy ✅
if ml_policy.mpi and ml_policy.mpi.num_proc_per_node is not None:  # None is None ✅
```

## The Key Distinction

The `is not None` check properly distinguishes between:
- **"Not set"** (`None`) → don't override accelerator count
- **"Explicitly set to 0"** (`0`) → override with 0 (CPU-only training)
- **"Explicitly set to positive number"** → override with that number

python/kubeflow/trainer/utils/utils.py
jskswamy: Made necessary changes to simplify the `detect_trainer_from_image_patterns` function

python/kubeflow/trainer/utils/utils.py
andreyvelich: But why `num_proc_per_node=0` is a valid value ?
We should not allow user to set such value or consider this as None.

python/kubeflow/trainer/types/types.py
andreyvelich: I agree that we should have dict to represent all Trainers where key is the Framework name and value is the Trainer object.
The question is should we also keep `framework` argument in the Trainer object. This is mostly used to just show users what framework this Trainer is using.

I am fine to keep it for now.

WDYT @szaher @astefanutti @Electronic-Waste ?

python/kubeflow/trainer/utils/utils.py
andreyvelich: @jskswamy Sorry for the late reply, I meant can you just use this code snippet ?
```py

def get_runtime_trainer(....):
....
image_name = trainer_container.image.split(":")[0]
trainer = get_trainer_from_image(image_name)


def get_trainer_from_image(image_name: str) -> types.Trainer:
    """
    Detect trainer type based on image name patterns using regex.
    This method uses pattern matching on the image name to determine
    the likely trainer type.
    Args:
        image_name: The container image name.
    Returns:
        Trainer: Trainer object if detected, otherwise the DEFAULT_TRAINER is returned.
    """
    # DeepSpeed patterns
    if re.search(r"deepspeed", image_name, re.IGNORECASE):
        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.DEEPSPEED])

    # MLX patterns
    if re.search(r"mlx", image_name, re.IGNORECASE):
        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.MLX])

    # TorchTune patterns (check before PyTorch to avoid conflicts)
    if re.search(r"torchtune", image_name, re.IGNORECASE):
        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.TORCHTUNE])

    # PyTorch patterns - require explicit "pytorch" in image name for clarity
    if re.search(r"pytorch", image_name, re.IGNORECASE):
        return copy.deepcopy(types.TRAINER_CONFIGS[types.Framework.TORCH])

    return copy.deepcopy(types.DEFAULT_TRAINER)

```

python/kubeflow/trainer/utils/utils.py
andreyvelich: @jskswamy Did you get a chance to check this comment ?

jskswamy: Sorry for late reply! I've addressed this, kindly check the changes now

python/kubeflow/trainer/utils/utils.py
jskswamy: I've simplified the function as per your suggestion, kindly check

astefanutti (commented):
I may not have the full context, so please pardon my naive question, why those metadata, mostly the framework type, could not come from the training runtime itself, in the form of an annotation or a label?

For custom images, as a platform admin / user, I would understand I need to provide some hints.

andreyvelich:
> o please pardon my naive question, why those metadata, mostly the framework type, could not come from the training runtime itself, in the form of an annotation or a label?

Previously we talked with @tenzen-y and @Electronic-Waste about introducing labels to the runtime that define framework type: https://cloud-native.slack.com/archives/C0742LDFZ4K/p1741266604716149?thread_ts=1741263570.091899&cid=C0742LDFZ4K

But we decided to not add more labels since this information can be retrieved from image and APIs.

astefanutti:
> But we decided to not add more labels since this information can be retrieved from image and APIs.

The implicit contract based on the image name might prove fragile for users and downstream projects.

I understand the regex-based heuristics could provide a nice last-resort as it stands, but making that contract explicit seems simpler and more robust. I still fail to see why it couldn't be enforced in the training runtime API, or at least the SDK would only fallback to the regex-based heuristics if that API contract is made optional.

andreyvelich:
>I understand the regex-based heuristics could provide a nice last-resort as it stands, but making that contract explicit seems simpler and more robust. I still fail to see why it couldn't be enforced in the training runtime API, or at least the SDK would only fallback to the regex-based heuristics if that API contract is made optional.

@astefanutti I think, we need to figure out why we even expose Runtime's trainer to the user.
Looking at the code it is only used to give user information about the Runtime:
https://github.com/kubeflow/sdk/blob/fa5778b8eea84d7e4d7506f8c0c5adc59cb28d5b/python/kubeflow/trainer/types/types.py#L166

Information such as:
- whether this runtime can be used with `CustomTrainer` or with `BuiltinTrainer`.
- What ML framework users should use with the runtime (I think, this might be removed if we add `get_runtime_packages()` API like I showed in the KubeCon previously: https://youtu.be/Fnb1a5Kaxgo?t=556
- Entrypoint that is used while getting the TrainJob steps: https://github.com/kubeflow/sdk/blob/fa5778b8eea84d7e4d7506f8c0c5adc59cb28d5b/python/kubeflow/trainer/utils/utils.py#L260 and while creating the TrainJob using CustomTrainer from function: https://github.com/kubeflow/sdk/blob/fa5778b8eea84d7e4d7506f8c0c5adc59cb28d5b/python/kubeflow/trainer/utils/utils.py#L340-L341

Do we think that we can refactor some of this and remove the `TRAINER_CONFIGS` list ?
Thoughts @astefanutti @szaher @kramaranya @eoinfennessy ?

astefanutti:
> * whether this runtime can be used with `CustomTrainer` or with `BuiltinTrainer`. **This is important data since users can incorrectly use `CustomTrainer` with builtin Runtimes** cc @Electronic-Waste

Right, that equally applies to the `BuiltinTrainer.config`. For now there is only one `TorchTuneConfig`, but nothing guarantees it's compatible with the training runtime when there'll be more. 

> Do we think that we can refactor some of this and remove the `TRAINER_CONFIGS` list ?

For built-in trainers, it seems there is a tight coupling between the trainer and the runtime, so maybe folding things into runtime as the "source-of-truth" would be better.

andreyvelich:
@astefanutti Do you think that framework information is useful for SDK users if they can always run `get_runtime_packages()` API ?

astefanutti:
> @astefanutti Do you think that framework information is still useful for SDK users if they can always run `get_runtime_packages()` API ?

No, though it'd be needed for checking the typed configuration passed by users for built-in trainers is compatible with the training runtime?

andreyvelich:
> No, though it'd be needed for checking the typed configuration passed by users for built-in trainers is compatible with the training runtime?

This is correct, additionally we can't run the `get_runtime_packages()` API for `BuiltinTrainer` runtimes since by default it contains script for fine-tuning: https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/torchtune/llama3_2/llama3_2_3B.yaml#L71.

Also, I don't think that users needs to know about installed packages in such runtimes, since they can only modify the config (e.g. fine-tuning parameters), but not the runtime packages.

Maybe for BuiltinTrainer runtime we should have two labels:

- `trainer.kubeflow.org/trainer-type: builtin`
- `trainer.kubeflow.org/builtin-config: torchtune`

If we don't want to introduce 2nd label, we can just tell users to rely on runtime name.


Thoughts @tenzen-y @astefanutti @Electronic-Waste @rudeigerc @szaher @kramaranya  ?

andreyvelich:
I think, we should refactor our Runtime class: https://github.com/kubeflow/sdk/blob/main/python/kubeflow/trainer/types/types.py#L176-L179

eoinfennessy:
Agreed that it would be better to add APIs to TrainingRuntime and ClusterTrainingRuntime to specify the framework instead of relying on image names and regex checks. Adding trainer type would also be useful.

But why use labels instead of adding `framework` and other fields to the runtime spec? This would allow us to use schema-based validation to ensure a valid framework is provided.

One idea for this that would use cross-field validation to ensure one and only one of `customTrainerConfig` or `builtinTrainerConfig` is provided. Fields `framework` and `type` could use enum-based validation:

```
spec:
  customTrainerConfig:
    framework: "torch"
  # OR
  builtinTrainerConfig:
    type: "torchtune"
```

Probably best to consider the exact APIs alongside work on kubeflow/trainer#2752.

andreyvelich:
>There is already the trainer.kubeflow.org/accelerator label.

I am not sure if we should continue to maintain this label. IIRC, @tenzen-y has concerns introducing this label in the runtimes.

> Labels are flexible and enable listing runtimes by label selectors,

We can also do field selector, if we introduce a new API in the runtime.

What are the pros and cons to add this property under labels or APIs ?

andreyvelich:
>I'm not sure custom fields from CRDs are indexed. It might be only few fields from core APIs.

You are right @astefanutti, here is the list of supported fields: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#list-of-supported-fields

>You're right, it may not be a good example "semantically".

Let me remove this label for now, unless we design better way to explain users the accelerator types in the Runtime.

@Electronic-Waste @astefanutti @tenzen-y Any concerns to introduce these three labels to our runtimes for now ?
```yaml
trainer.kubeflow.org/trainer-type: custom
or
trainer.kubeflow.org/trainer-type: builtin
trainer.kubeflow.org/builtin-config: torchtune
```

astefanutti:
> @Electronic-Waste @astefanutti @tenzen-y Any concerns to introduce these three labels to our runtimes for now ?
> 
> ```yaml
> trainer.kubeflow.org/trainer-type: custom
> or
> trainer.kubeflow.org/trainer-type: builtin
> trainer.kubeflow.org/builtin-config: torchtune
> ```

I think that's a good start. I only wonder whether those should be within the `sdk.kubeflow.org` prefix since it's metadata meant for the SDK.

> Alternatively, we can introduce `framework` label to the runtimes, but not sure if that is really needed, since users only need to know builtin config type to use while creating TrainJob using `BuiltinTrainer`.
> 
> ```
> trainer.kubeflow.org/trainer-type: custom
> trainer.kubeflow.org/framework: torch
> ---
> trainer.kubeflow.org/trainer-type: custom
> trainer.kubeflow.org/framework: deepspeed
> ---
> trainer.kubeflow.org/trainer-type: builtin
> trainer.kubeflow.org/framework: torchtune
> ```

Actually `framework` might be a different kind of metadata than those meant to be hints for the train API in the SDK.
That one would might be a simpler way to display the runtime framework rather than relying on the runtime name.

PR #35, opened 2025-06-30
Add optional dependencies for testing

jskswamy:
## Purpose
Add support for `uv` dependency management to the Kubeflow SDK.

## Background
This addresses the following feedback from PR #31:
> "Please can we add support for uv in the separate PR ?
> We have an open issue for it: #20"

## Changes
- Add test dependencies: `pytest>=7.0.0`, `pytest-mock>=3.10.0` in `pyproject.toml`
- Configure pytest with appropriate settings (import mode, verbose output, test discovery)
- Set `testpaths = ["kubeflow"]` to support co-located tests
- Add `uv.lock` file with locked dependency versions

## Related
- Addresses feedback from PR #31

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -35,6 +35,12 @@ dependencies = [
   "kubeflow_trainer_api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api"
 ]
 
+[project.optional-dependencies]
+test = [
+  "pytest>=7.0.0",
+  "pytest-mock>=3.10.0",
+]
+
 [project.urls]
 Homepage = "https://github.com/kubeflow/trainer"
 Documentation = "https://www.kubeflow.org/docs/components/trainer/"
@@ -48,3 +54,14 @@ path = "kubeflow/trainer/__init__.py"
 
 [tool.hatch.metadata]
 allow-direct-references = true
+
+[tool.pytest.ini_options]
+addopts = [
+    "--import-mode=importlib",
+    "-v",
+    "--tb=short",
+]
+testpaths = ["kubeflow"]
+python_files = ["test_*.py", "*_test.py"]
+python_classes = ["Test*"]
+python_functions = ["test_*"]
--- a/python/uv.lock
+++ b/python/uv.lock
@@ -1,0 +1,438 @@
+version = 1
+revision = 2
+requires-python = ">=3.12"
+
+[[package]]
+name = "annotated-types"
+version = "0.7.0"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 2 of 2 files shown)


python/pyproject.toml
```diff
@@ -48,3 +54,14 @@ path = "kubeflow/trainer/__init__.py"
 
 [tool.hatch.metadata]
 allow-direct-references = true
+
+[tool.pytest.ini_options]
+addopts = [
+    "--import-mode=importlib",
```
andreyvelich, line 60: Why do we need it ?

python/pyproject.toml
```diff
@@ -48,3 +54,14 @@ path = "kubeflow/trainer/__init__.py"
...
+    "-v",
+    "--tb=short",
+]
+testpaths = ["kubeflow"]
```
szaher, line 64: should we create a dedicated directory for tests? i.e. `kubeflow/tests` ?

python/pyproject.toml
```diff
@@ -35,6 +35,12 @@ dependencies = [
   "kubeflow_trainer_api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api"
 ]
 
+[project.optional-dependencies]
+test = [
+  "pytest>=7.0.0",
```
szaher, line 40: since we are going to use `uv` let's use `ruff` as well with uv?

python/uv.lock
```diff
@@ -0,0 +1,438 @@
+version = 1
+revision = 2
+requires-python = ">=3.12"
```
szaher, line 3: are we dropping old python versions? I think we still need to support 3.10 and 3.11

3.9 version will be dropped in October this year https://devguide.python.org/versions/

python/pyproject.toml
andreyvelich: I think, for unit tests, we agreed previously to store test files alongside the actual files.
That will make it easier to find unit tests, and make it consistent with Go best practices.
KFP also does it: https://github.com/kubeflow/pipelines/tree/master/sdk/python/kfp/client

szaher:
Let's split this PR into two as there are two things happening here. pytest and uv dependency management and we need to check if we are creating workspaces or not and some other directives for uv we need to decide on if this is Ok

cc: @andreyvelich

python/pyproject.toml
franciscojavierarceo: +1

szaher:
@jskswamy  Can we update this PR to only focus on `pytest` and unit testing and rebase it on top of this PR which adds uv and ruff support + CI job for linting? https://github.com/kubeflow/sdk/pull/38
cc: @andreyvelich

andreyvelich:
This has been resolved in: https://github.com/kubeflow/sdk/pull/38

PR #38, opened 2025-07-04
Add support for UV & Ruff

szaher:

* Use UV as dependency manager 
* Use ruff for linting 
* update `pyproject.toml` to include directives for uv support
* Add lint CI job

Fixes #

**Checklist:**


python/pyproject.toml
```diff
@@ -46,5 +42,27 @@ packages = ["kubeflow"]
...
+
+[tool.uv]
+package = true
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets]
+packages = ["kubeflow"]
```
andreyvelich, line 62: Given that we only have one target, do we need it ?

python/pyproject.toml
```diff
@@ -46,5 +42,27 @@ packages = ["kubeflow"]
 [tool.hatch.version]
 path = "kubeflow/trainer/__init__.py"
 
+[tool.ruff]
+line-length = 100
```
andreyvelich, line 43: Does ruff also replace flake8 ?
https://github.com/kubeflow/sdk/blob/4933dae8cd9cad70f1cb5b95903cfe5d09db0e05/.flake8

python/pyproject.toml
```diff
@@ -46,5 +42,27 @@ packages = ["kubeflow"]
...
+select = ["E", "F", "W"]
+ignore = ["E203"]
+
+[dependency-groups]
+dev = [
+    "ruff>=0.12.2",
+]
+
+[tool.uv]
+package = true
```
andreyvelich, line 55: Is this required ?

.pre-commit-config.yaml
```diff
@@ -16,10 +16,11 @@ repos:
     rev: 25.1.0
     hooks:
       - id: black
```
andreyvelich: Does ruff also replace black and isort ?

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,30 @@
+name: Lint
```
andreyvelich: I would move this GitHub action under `test-python.yaml`,  with the similar name as here: https://github.com/kubeflow/trainer/blob/master/.github/workflows/test-python.yaml#L1

In that test, we can just run `make verify` which verifies Python code, and run `make test-unit` which runs unit tests.

Removing number of parallel GitHub actions will help our CI to execute faster, given that we have limitations for concurrent runs.

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Install dependencies
+        run: |
+          cd python
+          uv sync
+
+      - name: Run ruff
+        run: |
+          cd python
+          uv run ruff check
```
andreyvelich: We can move all of this to the Makefile for `make verify` instruction.

andreyvelich (commented):
Thanks for this @szaher

.pre-commit-config.yaml
szaher: yes, it does replace `black` and `isort`

python/pyproject.toml
szaher: yes
image

python/pyproject.toml
szaher: since we have `[build-system]` we can drop this one

python/pyproject.toml
szaher: I think it was in the file already 
https://github.com/kubeflow/sdk/blob/main/python/pyproject.toml#L43-L44

we can remove it if not required

szaher (commented):
Thanks @andreyvelich for your review

.pre-commit-config.yaml
andreyvelich: @szaher In that case, let's remove them from the pre-commit and other configs.

python/pyproject.toml
andreyvelich: Yes, we can remove this one.

syntaxsdev: agreed.

syntaxsdev (commented):
looks good, similar concerns as Andrey

andreyvelich:
@szaher Did you get a chance to address the remaining comments, so we can merge this PR ?

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install bash
+        run: sudo apt-get update && sudo apt-get install -y bash
```
andreyvelich: bash should be installed on this runner.
We didn't install it here: https://github.com/kubeflow/trainer/blob/master/.github/workflows/test-python.yaml
However, we still use it in the Makefile: https://github.com/kubeflow/trainer/blob/master/Makefile#L11

Makefile
```diff
@@ -0,0 +1,90 @@
...
+.PHONY: verify-lock
+verify-lock: ## Check UV lockfile
+	@cd $(PY_DIR) && uv lock --check
+
+.PHONY: verify-ruff
+verify-ruff: ## Run ruff check
+	@cd $(PY_DIR) && uvx ruff check --show-fixes
+
+.PHONY: verify
+verify: verify-lock verify-ruff  ## Verify everything is Ok for python sdk
```
andreyvelich: Can you just run all instructions under `verify` to install tools if that is required, similar to this: https://github.com/kubeflow/trainer/blob/master/Makefile#L158
E.g.:
```makefile
UV := $(shell which uv)
uv:
ifeq ($(UV),)
  curl -LsSf https://astral.sh/uv/$(UV_VERSION)/install.sh | sh
  $(info uv has been installed)
endif

verify: uv ruff # install all required tools
  @cd $(PY_DIR) && uv lock --check
  @cd $(PY_DIR) && uvx ruff check --show-fixes
```

Makefile
```diff
@@ -0,0 +1,90 @@
...
+
+PY_DIR := $(PROJECT_DIR)/python
+
+REPO := github.com/kubeflow/sdk
+# Location to install tool binaries
+LOCALBIN ?= $(PROJECT_DIR)/bin
+
+# Tool versions
+UV_VERSION=0.7.2
+RUFF_VERSION=0.12.3
```
andreyvelich: Is that important which version of uv and ruff we run ?
We can consider to place them under [`bin/` if specific version is required](https://github.com/kubeflow/trainer/blob/master/Makefile#L18).

szaher: not really required. I will install the latest

Makefile
szaher: Thanks Andrey!
done

.github/workflows/lint-test.yaml
szaher: I was testing since I was getting errors from `.ONESHELL`

.github/workflows/lint-test.yaml
```diff
@@ -0,0 +1,26 @@
+name: Lint
```
andreyvelich: @szaher Shall we just name this Workflow and rename file to `test-python.yaml`:
```suggestion
name: Unit Test - Python
```

Similar to Trainer test: https://github.com/kubeflow/trainer/blob/master/.github/workflows/test-python.yaml#L14, we can just run 2 jobs: pre-commit + test

Makefile
```diff
@@ -0,0 +1,65 @@
...
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
+
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+
+PY_DIR := $(PROJECT_DIR)/python
+
+REPO := github.com/kubeflow/sdk
```
andreyvelich: This is unused
```suggestion
```

Makefile
```diff
@@ -0,0 +1,65 @@
...
+# http://linuxcommand.org/lc3_adv_awk.php
+
+help: ## Display this help.
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+
+UV := $(shell which uv)
+
+.PHONY: deps
+deps:
```
andreyvelich: Can you rename it to `uv`, it is better to keep instructions with the same name as binaries we install.
```suggestion
.PHONY: uv
uv:
```

szaher:
@andreyvelich one more review please :)

Makefile
```diff
@@ -0,0 +1,62 @@
...
+	$(info uv has been installed)
+endif
+
+.PHONY: ruff
+ruff: ## Install Ruff
+	@uvx ruff --help &> /dev/null || uv tool install ruff
+
+
+.PHONY: verify
+verify: ruff  ## install all required tools
```
andreyvelich:
```suggestion
verify: uv ruff  ## install all required tools
```

.github/workflows/test-python.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+
+      - name: Install ruff
+        run: make ruff
```
andreyvelich: That is not needed
```suggestion
```

andreyvelich (commented):
Thank you for this great contribution @szaher!
/lgtm
/approve

PR #40, opened 2025-07-17
Update pyproject.toml project links

szaher:
Fix Links in pyproject.toml to point to this repo

Fixes #

**Checklist:**

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -32,9 +32,9 @@ dependencies = [
 ]
 
 [project.urls]
-Homepage = "https://github.com/kubeflow/trainer"
-Documentation = "https://www.kubeflow.org/docs/components/trainer/"
-Source = "https://github.com/kubeflow/trainer"
+Homepage = "https://github.com/kubeflow/sdk"
+Documentation = "https://www.kubeflow.org/docs/components"
+Source = "https://github.com/kubeflow/sdk"
 
 [tool.hatch.version]
 path = "kubeflow/trainer/__init__.py"


python/pyproject.toml
```diff
@@ -32,9 +32,9 @@ dependencies = [
 ]
 
 [project.urls]
-Homepage = "https://github.com/kubeflow/trainer"
-Documentation = "https://www.kubeflow.org/docs/components/trainer/"
-Source = "https://github.com/kubeflow/trainer"
+Homepage = "https://github.com/kubeflow/sdk"
+Documentation = "https://www.kubeflow.org/docs/components"
```
astefanutti, line 36: Should we plan to add a dedicated section for the SDK in the doc?

@andreyvelich @Electronic-Waste WDYT?

Electronic-Waste: Currently, we planned to add SDK related section in each component (like "how to use BuiltinTrainer" in https://github.com/kubeflow/trainer/issues/2401#issuecomment-3069859005). But I think it should be a great idea to add a dedicated section for SDK in the doc. Since users do not care about the component they use, they only care about how to use it.

andreyvelich: Actually, I feel that we should use the project docs for the SDK for now.
Given that SDK is a primary interface for users to interact with Kubeflow projects APIs, why do we want to separate them in the website ? For example, KFP doesn't separate SDK docs from [its section.](https://www.kubeflow.org/docs/components/pipelines/user-guides/)

SDK implements various clients to talk to Kubeflow projects (e.g. TrainerClient, PipelinesClient, OptimizerClient), so users understand with which project they interact with, and know which docs to check.

Additionally, since we say that Python SDK is the primary interface for Kubeflow Trainer, all of the users guides will be using the Kubeflow SDK: https://www.kubeflow.org/docs/components/trainer/user-guides/

We should also talk about it during our Training WG call.

kramaranya: I agree that we should have a dedicated SDK section once we're ready for the first SDK release. We'll need a central place to provide an overview of the Kubeflow SDK, what components are included, installation steps and links to each component's user guides. I'm not saying we duplicate content or separate the SDK docs from their components, but we should create one main place where users can start.

We want to promote adoption and if someone asks "how do I get started with the Kubeflow SDK" we don't have a clear page to point them to. Do you want to point them to README from SDK repo instead?

Dedicated SDK section will be a starting point to help users understand the scope and capabilities of the SDK before they dive into specific component documentation.

+1 to discuss this during Training WG call

Issue #48, opened 2024-12-10
[SDK] Snapshot users' workspace into distributed TrainJob workload


andreyvelich:
### What you would like to be added?

As we discussed earlier, we want to design an approach to snapshot users' workspace into TrainJob (e.g. distributed ML workload): https://github.com/kubeflow/training-operator/pull/2324#discussion_r1862719941.
To achieve this, we plan to generate a unique TrainJob ID before submitting it to the Kubernetes control plane.

During the KubeCon 2024 demo, we demonstrated how workspace snapshotting might work: https://youtu.be/Lgy4ir1AhYw?t=458.
In this demo, we pushed Python code files into S3 and then loaded them into TrainJob using initContainers.

However, we can consider various approaches, for instance:
- Using distributed cache.
- Using `kubectl cp`.

### Why is this needed?

This should streamline Data Scientists user experience while working with Kubeflow Training Python SDK.
 

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

andreyvelich:
Yes, we can also use `git` as a solution for users' workspace.
I think, @shravan-achar has a few prototypes on how we can make archive from users' workspace, push it to blob storage, and download it before training starts.

astefanutti:
@shravan-achar thanks for the extra details.

I was having in mind something similar to what `ray job submit` does with the `working_dir` parameter: https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#runtime-environments-api-ref.

I can see there would be more powerful approaches, though my understanding is that any "data plane" API solution would have to come with proper tenancy and security, while the option to rely on existing Kubernetes "control-plane" mechanism (like `cp` / `exec`) have those already built-in.

PR #49, opened 2025-07-24
Allow CustomTrainer to run a Python script directly via python_file

### LINKED ISSUE #47: Support running a Python script directly in CustomTrainer (e.g., python myscript.py)

I would like the `CustomTrainer` API in the Kubeflow Trainer SDK to support running a Python script directly by specifying a file path (e.g., `python myscript.py`), instead of requiring users to provide a Python function.

**Proposed API:**
```python
CustomTrainer(python_file="run_kubernetes.py", ...)
```
- If `python_file` is provided, the SDK should set the container entrypoint to `["python", "run_kubernetes.py"]` (or the specified file).
- No function serialization, no wrapper scripts, no subprocesses, and no use of `runpy` inside another Python process.
- This should be mutually exclusive with the existing `func` argument.

### Why is this needed?

- **Simplicity & Familiarity:** Most users have existing training scripts and expect to run them as `python myscript.py`, just like they do locally or in YAML-based jobs.
- **Avoids Indirection:** The current approach requires wrapping code in a function, which is then serialized, deserialized, and run via a generated entrypoint script. This is convoluted, harder to debug, and not how most ML workflows are structured.
- **Better UX:** Direct script execution is more transparent, easier to reason about, and matches user expectations from other ML platforms and Kubernetes YAML jobs.
- **Migration Path:** This makes it much easier for users to migrate from script-based workflows (e.g., Slurm, bash, or direct Kubernetes Jobs) to Kubeflow Trainer.
- **Cleaner Container Lifecycle:** Running the script as the main process ensures correct signal handling, exit codes, and resource cleanup.

### Love this feature?

jskswamy:

This PR adds support for running a Python script directly in `CustomTrainer` by specifying a `python_file` argument. If `python_file` is set, the job will run the specified script as the main process (`python myscript.py`) instead of requiring a function. This is mutually exclusive with `func`; an error is raised if both or neither are set. The original function-based usage is unchanged and fully backward compatible.

- Simplifies migration from script-based workflows (e.g., Slurm, bash, or direct Kubernetes Jobs) to Kubeflow Trainer.
- Matches user expectations for direct script execution (`python myscript.py`).
- Avoids the indirection and complexity of function serialization and wrapper scripts.
- Ensures the script runs as the main process, improving signal handling and debugging.

Fixes #47

**Summary of changes:**
- `CustomTrainer` now accepts an optional `python_file` argument.
- If `python_file` is set, the SDK sets the container entrypoint to `["python", python_file]`.
- Mutually exclusive with `func`; validation is added.
- No changes to existing function-based usage.

---

**Additional context:**
See [kubeflow/sdk#47](https://github.com/kubeflow/sdk/issues/47) for the original feature request and motivation.

### DIFF
--- a/kubeflow/trainer/utils/utils.py
+++ b/kubeflow/trainer/utils/utils.py
@@ -422,20 +422,16 @@ def get_trainer_crd_from_command_trainer(
     if trainer.resources_per_node:
         trainer_crd.resources_per_node = get_resources_per_node(trainer.resources_per_node)
 
-    # If packages need to be installed OR no explicit command is provided, use bash-wrapped path
-    # to preserve install features and runtime launcher behavior. Otherwise, pass-through.
-    if trainer.packages_to_install or not (trainer.command and len(trainer.command) > 0):
-        trainer_crd.command = get_command_using_user_command(
-            runtime=runtime,
-            command=list(trainer.command or []),
-            command_args=trainer.args,
-            pip_index_urls=trainer.pip_index_urls,
-            packages_to_install=trainer.packages_to_install,
-            pip_extra_args=trainer.pip_extra_args,
-        )
-    else:
-        trainer_crd.command = list(trainer.command)
-        trainer_crd.args = list(trainer.args or [])
+    # Always produce a bash-wrapped command to ensure shell interpolation (e.g. ${VAR})
+    # and to preserve runtime launcher behavior consistently.
+    trainer_crd.command = get_command_using_user_command(
+        runtime=runtime,
+        command=list(trainer.command or []),
+        command_args=trainer.args,
+        pip_index_urls=trainer.pip_index_urls,
+        packages_to_install=trainer.packages_to_install,
+        pip_extra_args=trainer.pip_extra_args,
+    )
 
     if trainer.env:
         trainer_crd.env = [
--- a/kubeflow/trainer/utils/utils_test.py
+++ b/kubeflow/trainer/utils/utils_test.py
@@ -396,15 +396,15 @@ class TestGetTrainerCRDFromCommandTrainer:
         )
         assert crd.command == expected_command
 
-    def test_pass_through_when_command_without_installs(self):
+    def test_always_bash_wrapped_even_without_installs(self):
         runtime = _build_plain_runtime()
         trainer = types.CommandTrainer(
             command=["python"],
             args=["main.py"],
         )
         crd = utils.get_trainer_crd_from_command_trainer(runtime, trainer)
-        assert crd.command == ["python"]
-        assert crd.args == ["main.py"]
+        # Should wrap into bash -c preserving python main.py
+        assert crd.command == ["bash", "-c", "python main.py"]
 
     def test_preserves_prefix_plain(self):
         runtime = _build_plain_runtime()


eoinfennessy:
I'm not sure about this approach as it assumes the training script to run is available within the training runtime image. This would require users to rebuild training images each time a change is made to a training script.

Having said that, I think it would be a big UX improvement to allow users to provide a training script to run instead of a train function, but this script should come from the user's workspace instead of being baked into the training runtime image. This is related to issue #48, which describes making a snapshot of a user's workspaces available in training pods so that [multiple] Python files can be used in training jobs.

andreyvelich (commented):
Thank you for this @jskswamy!
I am wondering what are the benefits of such API compare to just this:


```python
from custom_script import run_pytorch

TrainerClient().train(
  trainer=CustomTrainer(
    func=run_pytorch,
    num_nodes=5
   )
)
```

If the TrainJob contains users' workspace, they should be able to execute the above script.
cc @shravan-achar

andreyvelich:
@jskswamy A few questions:
1. How you can ensure that Python file is inside the TrainJob workload ?
2. As you can see, entrypoint depends on [runtime Trainer](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/constants/constants.py#L164). It could be `torchrun`, `mpirun`, or Python in case of PlainML.

jskswamy:
@andreyvelich  Thanks for pointing out the lack of framework-aware command support. I have added the following support now.

The `python_file` feature now uses the runtime's framework-specific commands instead of hardcoded `["python"]`:

- **PyTorch runtime**: Uses `torchrun train.py --args` 
- **MPI runtime**: Uses `mpirun python train.py --args`
- **PlainML runtime**: Uses `python train.py --args`

On the Python file availability, we are making an assumption that the Python file will be made available by the consumer through:

1. **Embedded in container image** - Scripts baked into the Docker image
2. **ConfigMaps** - Scripts mounted as ConfigMaps  
3. **Volume mounts** - Scripts mounted via PVC or other volumes
4. **Init containers** - Scripts downloaded by init containers

This design assumption provides flexibility for different deployment patterns while keeping the `python_file` feature simple and focused on direct script execution.

python/kubeflow/trainer/utils/utils.py
```diff
@@ -372,6 +372,20 @@ def get_trainer_crd_from_custom_trainer(
             trainer.resources_per_node
         )
 
+    if trainer.python_file:
+        if trainer.func:
+            raise ValueError("Specify only one of func or python_file in CustomTrainer.")
+        trainer_crd.command = ["python"]
+           # Combine python_file with python_args
+        args = [trainer.python_file]
```
kramaranya: This still hardcodes `command = ["python"]` for all runtimes, which will not work for distributed training with MPI and PyTorch, right?

python/kubeflow/trainer/utils/utils.py
```diff
@@ -372,6 +372,20 @@ def get_trainer_crd_from_custom_trainer(
...
+        if trainer.python_args:
+            args.extend(trainer.python_args)
+        trainer_crd.args = args
+        return trainer_crd
```
kramaranya: This early return skips several important features that users reasonably expect to work, such as 
- package installation
https://github.com/jskswamy/kubeflow-sdk/blob/feature/customtrainer-python-file/python/kubeflow/trainer/utils/utils.py#L391-L397
- environment variable handling
https://github.com/jskswamy/kubeflow-sdk/blob/feature/customtrainer-python-file/python/kubeflow/trainer/utils/utils.py#L400-L404

kramaranya (commented):
Thank you @jskswamy!

Have you considered how this affects the plugin architecture when someone writes a custom runtime?

kramaranya:
Hey @jskswamy! Did you have a chance to check the latest comments?

jskswamy:
I updated the original approach proposed in PR #49 to address the review concerns and make the feature more extensible and robust.

**What changed**
  - I replaced `python_file/python_args` with a new, explicit `CommandTrainer` that runs an arbitrary command with arguments under the runtime’s launcher.
  - The function path (`CustomTrainer`) remains intact, without special-casing or hardcoding.

**Why this fits the plugin architecture**
  - I reuse the runtime’s launcher template (torchrun/mpirun/python) and only replace the payload, so custom runtimes continue to control entrypoints and arguments.
  - No `command=["python"]` hardcoding; the launcher remains framework-aware. See discussion in [PR #49](https://github.com/kubeflow/sdk/pull/49).

**Distributed runtimes**
  - Torch uses `torchrun`, MPI uses `mpirun` (with `--user` pip installs), PlainML uses `python`.
  - This works for multi-node and device-aware training consistently.

**Shared features are preserved (no early return)**
  - Environment variables, resources, and pip installs are applied in the same builder path as functions.
  - The payload includes optional installs followed by the user command.

**File availability assumption**
  - The script/binaries must be present in the container (image/ConfigMap/volume/init). The workspace snapshot UX remains future work (issue #48 referenced in the PR thread).

**Validation strategy**
  - Consistent with `CustomTrainer`, I validate in the builder and error if `command` is missing/invalid.
  - I can add `__post_init__` later for earlier failure if needed, but builder-level checks match current patterns.

**Tests and docs**
  - Helper tests (plain/MPI) assert exact launcher shape and multiline payloads.
  - CRD builder tests cover env/resources/num_nodes and MPI `--user` installs.
  - Backend routing test verifies `CommandTrainer` dispatch.
  - README includes a minimal `CommandTrainer` usage example.

**Outcome**
  - Clear, framework-aware command execution without bypassing shared features or limiting plugin extensibility, and a clean separation from the function-based path.
  
@kramaranya , @andreyvelich kindly review the changes and let me know what you think about the new approach

README.md
```diff
@@ -75,6 +75,37 @@ TrainerClient().wait_for_job_status(job_id)
...
+CommandTrainer runs an arbitrary command inside the runtime’s launcher (torchrun/mpirun/python) while preserving package installation, env vars, and resources.
+
+```python
+from kubeflow.trainer import TrainerClient
+from kubeflow.trainer.types import types
+
+client = TrainerClient()
+rt = client.get_runtime("torch")  # or "mpi", "plainml"
+
+trainer = types.CommandTrainer(
```
andreyvelich, line 85: This is an interesting idea and somewhat aligned with what we discussed today at the Kubeflow SDK call with `KubernetesTrainer` proposed by @szaher: https://youtu.be/mv8GoWdefck?t=832

Since we distinguish the runtime trainers between [CustomTrainer and BuiltinTrainer](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/utils/utils.py#L119-L126), I am wondering if we want to introduce `CustomTrainerContainer()` type which give users control to configure `image`, `container`, `args` instead of passing the training function.

Would that be helpful for integration between KFP and Trainer ?

Thoughts @kubeflow/kubeflow-sdk-team @mprahl @franciscojavierarceo @ederign @rudeigerc?

jskswamy:
### Update on CommandTrainer behavior (pass-through + launcher-aware fallback)

- When `command` is provided and no `packages_to_install`: CRD uses `command`/`args` directly (no shell wrapper).
- When `packages_to_install` is set OR `command` is not provided: runtime launcher is reused via `get_command_using_user_command` (bash-wrapped), preserving installs and launcher selection (`torchrun`/`mpirun`/`python`).
- Mirrors BuiltinTrainer default launcher selection for the fallback path; explicit commands remain unchanged.
- Tests updated to cover both pass-through and launcher-wrapped paths.

Examples:
```python
# Direct (no wrapper)
trainer = types.CommandTrainer(command=["torchrun"], args=["--standalone", "train.py"])

# Install-aware (bash-wrapped)
trainer = types.CommandTrainer(
  command=["torchrun"],
  args=["--standalone", "train.py"],
  packages_to_install=["nemo", "deepspeed"],
)
```

andreyvelich:
This is resolved by: https://github.com/kubeflow/sdk/pull/127
/close

PR #51, opened 2025-07-27
chore(trainer): Remove accelerator label from the runtimes

andreyvelich:
As we discussed in this thread: https://github.com/kubeflow/sdk/pull/31#issuecomment-3124772141, I will remove the accelerator label from the runtime for now.

In the future we can find better way to detect accelerator of runtime if that is required.

/assign @astefanutti @Electronic-Waste @szaher @kramaranya @eoinfennessy


andreyvelich (commented):
/approve

PR #52, opened 2025-07-28
feat(trainer): Add `wait_for_job_status()` API

andreyvelich:
I added the `wait_for_job_status()` API to the TrainerClient.

This API is similar to the [v1 SDK](https://github.com/kubeflow/trainer/blob/release-1.9/sdk/python/kubeflow/training/api/training_client.py).

### Note on TrainJob Status
I updated the TrainJob Running status. I consider that when all training nodes (e.g. Pods) are in Running phase, we can transition TrainJob to the Running status. I believe that is good assumption for the majority of Trainer users.

Looking forward for your feedback!

/assign @kubeflow/kubeflow-trainer-team @astefanutti @szaher @kramaranya @eoinfennessy @briangallagher 

TODO:
- [x] Update unit tests

### DIFF
--- a/python/kubeflow/trainer/api/trainer_client.py
+++ b/python/kubeflow/trainer/api/trainer_client.py
@@ -18,7 +18,8 @@ import queue
 import random
 import string
 import uuid
-from typing import Dict, List, Optional, Union
+import time
+from typing import Dict, List, Optional, Union, Set
 
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
@@ -433,6 +434,61 @@ class TrainerClient:
 
         return logs_dict
 
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: Set = {constants.TRAINJOB_COMPLETE},
+        timeout: int = 600,
+        polling_interval=5,
+    ):
+        """Wait for TrainJob to reach the desired status
+
+        Args:
+            name: Name of the TrainJob.
+            status: Set of expected statuses. It must be subset of Running, Complete, and Failed
+                statuses.
+            timeout: How many seconds to wait until TrainJob reaches one of the expected conditions.
+            polling_interval: The polling interval in seconds to check TrainJob status.
+
+        Returns:
+            TrainJob: The training job that reaches desired status.
+
+        Raises:
+
+            ValueError: The input values are incorrect.
+            RuntimeError: Failed to get TrainJob or TrainJob reaches unexpected Failed status.
+            TimeoutError: Timeout to wait for TrainJob status.
+        """
+
+        job_statuses = {
+            constants.TRAINJOB_RUNNING,
+            constants.TRAINJOB_COMPLETE,
+            constants.TRAINJOB_FAILED,
+        }
+        if not status.issubset(job_statuses):
+            raise ValueError(
+                f"Expected status {status} must be a subset of {job_statuses}"
+            )
+        for _ in range(round(timeout / polling_interval)):
+            trainjob = self.get_job(name)
+
+            # Raise an error if TrainJob is Failed and it is not the expected status.
+            if (
+                constants.TRAINJOB_FAILED not in status
+                and trainjob.status == constants.TRAINJOB_FAILED
+            ):
+                raise RuntimeError(f"TrainJob {name} is Failed")
+
+            # Return the TrainJob if it reaches the expected status.
+            if trainjob.status in status:
+                return trainjob
+
+            time.sleep(polling_interval)
+
+        raise TimeoutError(
+            f"Timeout waiting for TrainJob {name} to reach {status} Status"
+        )
+
     def delete_job(self, name: str):
         """Delete the TrainJob.
 
@@ -485,7 +541,6 @@ class TrainerClient:
             trainer=utils.get_runtime_trainer(
                 runtime_crd.spec.template.spec.replicated_jobs,
                 runtime_crd.spec.ml_policy,
-                runtime_crd.metadata,
             ),
         )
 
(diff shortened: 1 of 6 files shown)


python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,61 @@ def get_job_logs(
...
+            ValueError: The input values are incorrect.
+            RuntimeError: Failed to get TrainJob or TrainJob reaches unexpected Failed status.
+            TimeoutError: Timeout to wait for TrainJob status.
+        """
+
+        job_statuses = {
+            constants.TRAINJOB_RUNNING,
+            constants.TRAINJOB_COMPLETE,
+            constants.TRAINJOB_FAILED,
+        }
```
kramaranya: `TRAINJOB_CREATED` should be here, right?

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,61 @@ def get_job_logs(
 
         return logs_dict
 
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: Set = {constants.TRAINJOB_COMPLETE},
+        timeout: int = 600,
+        polling_interval=5,
+    ):
```
kramaranya:
```suggestion
    def wait_for_job_status(
        self,
        name: str,
        status: Set[str] = {constants.TRAINJOB_COMPLETE},
        timeout: int = 600,
        polling_interval: int = 5,
    ):
```

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,61 @@ def get_job_logs(
 
         return logs_dict
 
+    def wait_for_job_status(
```
kramaranya, line 436: Shall we also accept `namespace` parameter?

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,61 @@ def get_job_logs(
...
+        polling_interval=5,
+    ):
+        """Wait for TrainJob to reach the desired status
+
+        Args:
+            name: Name of the TrainJob.
+            status: Set of expected statuses. It must be subset of Running, Complete, and Failed
+                statuses.
+            timeout: How many seconds to wait until TrainJob reaches one of the expected conditions.
+            polling_interval: The polling interval in seconds to check TrainJob status.
```
kramaranya: What do you think about adding a verbose boolean parameter, which will print status updates? I think we also did smth similar in v1 -- https://github.com/kubeflow/trainer/blob/release-1.9/sdk/python/kubeflow/training/api/training_client.py#L1053-L1058.

That might look like:
`client.wait_for_job_status("my-job", verbose=True)`

kramaranya (commented):
Thank you @andreyvelich!!

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: Actually, there is no need to wait for Created status, since we automatically apply it after TrainJob is created in Kubernetes cluster:
https://github.com/kubeflow/sdk/blob/476b1b2551f42c8b8589f1ace67ac4b17572ccbf/python/kubeflow/trainer/api/trainer_client.py#L654

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: No, since the namespace is controlled by TrainerClient: https://github.com/kubeflow/sdk/blob/main/python/kubeflow/trainer/api/trainer_client.py#L38
We don't allow APIs (e.g. `get_job()`) to override it.

Ideally, we should abstract the namespace context from the SDK user.

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: We can, alternatively we can use `logger.debug()` to print TrainJob status updates.
Like in other APIs: https://github.com/kubeflow/sdk/blob/main/python/kubeflow/trainer/api/trainer_client.py#L250-L252

WDYT @astefanutti @kramaranya @Electronic-Waste @szaher @briangallagher ?

python/kubeflow/trainer/api/trainer_client.py
kramaranya: Yeah, but what if the user does `client.wait_for_job_status("my-job", status={"Created"})`?

python/kubeflow/trainer/api/trainer_client.py
kramaranya: Oh, I see, makes sense now!

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: I also noticed that I incorrectly assign status to the TrainJob, fixed it here: https://github.com/kubeflow/sdk/pull/52/commits/675f101807e6e3856aea89e7817628871cbbf48c

I think, you are right @kramaranya, it might make sense to allow user wait for Created status.

Although, we don't execute creation of TrainJob async, I can imagine that users want to run something like this:
https://github.com/kubeflow/sdk/blob/main/python/kubeflow/trainer/api/trainer_client.py#L234-L240

```py
job_id = train(...)
job = wait_for_job_status(job_id, status={"Created"})
print(job.creation_timestamp)
```

python/kubeflow/trainer/api/trainer_client.py
astefanutti, line 436: Would that make sense to be a bit more generic and make it something like `wait_for_job_condition`, similar to what the `kubectl wait` command provides?

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,61 @@ def get_job_logs(
...
+            constants.TRAINJOB_CREATED,
+            constants.TRAINJOB_RUNNING,
+            constants.TRAINJOB_COMPLETE,
+            constants.TRAINJOB_FAILED,
+        }
+        if not status.issubset(job_statuses):
+            raise ValueError(
+                f"Expected status {status} must be a subset of {job_statuses}"
+            )
+        for _ in range(round(timeout / polling_interval)):
```
astefanutti: Do you think a watch request be used instead of polling?

andreyvelich: Sure, let me try that!

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: I was thinking about it, however I am not sure if we want to expose complexity of CR condition to the SDK user.
For example, `condition` has `type`, `status`, and `reason`, etc. API that we don't really need to expose to the user (at least for now).

Thus, I suggest that we just expose single TrainJob status to make it easier to read.

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -433,6 +434,75 @@ def get_job_logs(
...
+        if not status.issubset(job_statuses):
+            raise ValueError(
+                f"Expected status {status} must be a subset of {job_statuses}"
+            )
+
+        # Use Kubernetes watch API to monitor the TrainJob's Pods.
+        w = watch.Watch()
+        try:
+            for event in w.stream(
+                self.core_api.list_namespaced_pod,
```
andreyvelich, line 474: @astefanutti I had to watch for Pod events since we don't push all events to TrainJob, but I think that would be fine for now.

astefanutti: @andreyvelich I'm not sure I understand why it's needed to watch for pods since the logic then gets the TrainJob with `trainjob = self.get_job(name)`? The event should have what's needed?

python/kubeflow/trainer/api/trainer_client.py
astefanutti: Sounds good, that makes sense. We can always add a "lower" level method later if needed.

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: @astefanutti The trick is that we don't expose events of running pods to the TrainJob.
We only have Completed and Failed condition for now.
And TrainJob Running condition is set when all training node Pods are running: https://github.com/kubeflow/sdk/blob/81e43fdd655c39ba6b16fc6e6a385ffbeb2753ac/python/kubeflow/trainer/api/trainer_client.py#L661-L668

astefanutti: @andreyvelich Ah I see. I missed this. That makes sense then.

astefanutti:
/lgtm

Thanks!

PR #54, opened 2025-07-30
feat(trainer): Add environment variables argument to CustomTrainer

astefanutti:

This enables users to pass environment variables to the custom trainer.

/assign @kubeflow/kubeflow-trainer-team @szaher @kramaranya @eoinfennessy @briangallagher

**Checklist:**

### DIFF
--- a/python/kubeflow/trainer/api/trainer_client_test.py
+++ b/python/kubeflow/trainer/api/trainer_client_test.py
@@ -69,6 +69,7 @@ BASIC_TRAIN_JOB_NAME = "basic-job"
 TRAIN_JOBS = "trainjobs"
 TRAIN_JOB_WITH_BUILT_IN_TRAINER = "train-job-with-built-in-trainer"
 TRAIN_JOB_WITH_CUSTOM_TRAINER = "train-job-with-custom-trainer"
+TRAIN_JOB_WITH_CUSTOM_TRAINER_ENV = "train-job-with-custom-trainer-env"
 
 
 # --------------------------
@@ -221,12 +222,12 @@ def get_resource_requirements() -> models.IoK8sApiCoreV1ResourceRequirements:
     )
 
 
-def get_custom_trainer() -> models.TrainerV1alpha1Trainer:
+def get_custom_trainer(include_env: bool = False) -> models.TrainerV1alpha1Trainer:
     """
     Get the custom trainer for the TrainJob.
     """
 
-    return models.TrainerV1alpha1Trainer(
+    trainer = models.TrainerV1alpha1Trainer(
         command=["bash", "-c"],
         args=[
             '\nif ! [ -x "$(command -v pip)" ]; then\n    python -m ensurepip '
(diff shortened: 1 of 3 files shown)


eoinfennessy (approved):
LGTM!

python/kubeflow/trainer/types/types.py
```diff
@@ -35,6 +35,7 @@ class CustomTrainer:
         pip_index_url (`Optional[str]`): The PyPI URL from which to install Python packages.
         num_nodes (`Optional[int]`): The number of nodes to use for training.
         resources_per_node (`Optional[Dict]`): The computing resources to allocate per node.
+        env (`Optional[Dict[str, str]]`): Environment variables to set in the training containers.
```
andreyvelich: Let's be more explicit, since SDK users should not be aware of containers ?
```suggestion
        env (`Optional[Dict[str, str]]`): The environment variables to set in the training nodes.
```

python/kubeflow/trainer/utils/utils.py
```diff
@@ -410,6 +410,18 @@ def get_trainer_crd_from_custom_trainer(
...
+    if trainer.env:
+        env_vars = []
+        for key, value in trainer.env.items():
+            env_vars.append(
+                models.IoK8sApiCoreV1EnvVar(
+                    name=key,
+                    value=value
+                )
+            )
+        trainer_crd.env = env_vars
```
andreyvelich: Maybe something like this:
```suggestion
    if trainer.env:
        trainer_crd.env = [
            models.IoK8sApiCoreV1EnvVar(name=key, value=value)
            for key, value in trainer.env.items()
        ]
```

python/kubeflow/trainer/api/trainer_client_test.py
```diff
@@ -277,7 +286,7 @@ def get_train_job(
     if add_built_in_trainer:
         train_job = add_built_in_trainer_to_job(train_job)
     if add_custom_trainer:
```
kramaranya:
```suggestion
    if add_built_in_trainer:
        train_job = add_built_in_trainer_to_job(train_job)
    elif add_custom_trainer:
```

python/kubeflow/trainer/utils/utils.py
kramaranya: +1

kramaranya (commented):
Looks great, thank you @astefanutti!

python/kubeflow/trainer/api/trainer_client_test.py
```diff
@@ -695,6 +703,27 @@ def test_list_runtimes(training_client, test_case):
...
+                    num_nodes=2,
+                    env={
+                        "TEST_ENV": "test_value",
+                        "ANOTHER_ENV": "another_value",
+                    },
+                )
+            },
+            expected_output=get_train_job(
+                train_job_name=TRAIN_JOB_WITH_CUSTOM_TRAINER_ENV,
+                train_job_trainer=get_custom_trainer(include_env=True),
```
andreyvelich: Instead of passing boolean to `the get_custom_trainer()`, can we directly pass the env list, it looks clearer, I think.
```suggestion
                train_job_trainer=get_custom_trainer([
                    models.IoK8sApiCoreV1EnvVar(name="TEST_ENV", value="test_value"),
                    models.IoK8sApiCoreV1EnvVar(name="ANOTHER_ENV", value="another_value"),
                ]
                ),
```

astefanutti: Yes I agree, I kept the old approach for consistency but I prefer this. Let me change it.

astefanutti: Updated, thanks.

andreyvelich (commented):
Thanks @astefanutti!
/lgtm
/approve

PR #55, opened 2025-08-01
chore(ci): Add dev tests with master dependencies

kramaranya:
I added a test-python-dev target and changed the unit test CI jobs with a dev/prod matrix

Dev leg installs the dev dependency group, which now pulls `kubeflow_trainer_api` from the master branch. Prod leg will run with the packages published to PyPI (once API models are available there)

I also updated the CI job to run unit tests for two Python vesions -- `3.9` and `3.11`

I removed flake8 since ruff replaced it

/assign @andreyvelich @astefanutti @Electronic-Waste @tenzen-y @briangallagher

**Checklist:**

### DIFF
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -8,6 +8,15 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        dependency-group:
+          - {name: dev,  make-target: test-python-dev}
+          - {name: prod, make-target: test-python}
+
+    name: Test ${{ matrix.dependency-group.name }}
+
     steps:
       - uses: actions/checkout@v4
 
@@ -21,9 +30,10 @@ jobs:
 
       - name: Run Tests
         run: |
-          make test-python report=xml
+          make ${{ matrix.dependency-group.make-target }} report=xml
 
       - name: Upload coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }} # Pass the GITHUB_TOKEN
+          flag-name: ${{ matrix.dependency-group.name }}
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -16,11 +16,8 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.9', '3.11']
-        dependency-group:
-          - {name: dev,  make-target: test-python-dev}
-          - {name: prod, make-target: test-python}
 
-    name: Test ${{ matrix.dependency-group.name }} (Python ${{ matrix.python-version }})
+    name: Test (Python ${{ matrix.python-version }})
 
     steps:
       - uses: actions/checkout@v4
@@ -35,13 +32,13 @@ jobs:
 
       - name: Run Tests
         run: |
-          make ${{ matrix.dependency-group.make-target }} report=xml
+          make test-python report=xml
 
       - name: Upload coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          flag-name: ${{ matrix.dependency-group.name }}-py${{ matrix.python-version }}
+          flag-name: py${{ matrix.python-version }}
           parallel: true
 
   coverage:
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -5,25 +5,30 @@ on:
   push:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
+        python-version: ['3.9', '3.11']
         dependency-group:
           - {name: dev,  make-target: test-python-dev}
           - {name: prod, make-target: test-python}
 
-    name: Test ${{ matrix.dependency-group.name }}
+    name: Test ${{ matrix.dependency-group.name }} (Python ${{ matrix.python-version }})
 
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Python
+      - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: ${{ matrix.python-version }}
 
       - name: Verify
         run: make verify
@@ -35,5 +40,17 @@ jobs:
       - name: Upload coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }} # Pass the GITHUB_TOKEN
-          flag-name: ${{ matrix.dependency-group.name }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: ${{ matrix.dependency-group.name }}-py${{ matrix.python-version }}
+          parallel: true
+
+  coverage:
+    if: ${{ always() }}
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Coveralls Finished
+        uses: coverallsapp/github-action@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel-finished: true
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -38,16 +38,3 @@ jobs:
         uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
-          flag-name: py${{ matrix.python-version }}
-          parallel: true
-
-  coverage:
-    if: ${{ always() }}
-    needs: test
-    runs-on: ubuntu-latest
-    steps:
-      - name: Coveralls Finished
-        uses: coverallsapp/github-action@v2
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
-          parallel-finished: true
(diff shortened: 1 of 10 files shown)


Electronic-Waste (commented):
@kramaranya Thanks for this!
/lgtm

kramaranya:
@astefanutti @Electronic-Waste thanks for your review!
I’ve updated the CI to run unit tests on Python 3.9 and 3.11, and removed flake8 because ruff replaced it in #38 
Could you please take a look again?

python/pyproject.toml
```diff
@@ -55,6 +55,7 @@ ignore = ["E203"]
 [dependency-groups]
 dev = [
     "ruff>=0.12.2",
+    "kubeflow_trainer_api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api",
```
andreyvelich: Is there any use-case when users would like to install Kubeflow SDK with the latest module package ?
IIUC, the dependency-groups is never published to PyPI by hatch, so users can't install.
Thoughts @astefanutti @szaher @briangallagher @kramaranya @eoinfennessy ?

Makefile
```diff
@@ -67,9 +67,20 @@ uv-venv:
...
+.PHONY: test-python-dev
+test-python-dev: uv-venv
+	@uv pip install --project $(PY_DIR) --group dev "$(PY_DIR)[test]"
+	@uv run coverage run --source=kubeflow.trainer.api.trainer_client,kubeflow.trainer.utils.utils -m pytest ./python/kubeflow/trainer/api/trainer_client_test.py
+	@uv run coverage report -m kubeflow/trainer/api/trainer_client.py kubeflow/trainer/utils/utils.py
+ifeq ($(report),xml)
+	@uv run coverage xml
+else
+	@uv run coverage html
+endif
```
andreyvelich: Can we run tests over the master modules for now ?
Since we can't introduce any breaking changes to the Kubernetes API, I don't see a need to run tests against released modules.

.github/workflows/test-python.yaml
```diff
@@ -5,25 +5,52 @@ on:
   push:
     branches: [main]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
andreyvelich, line 10: IIRC, we discussed before that this part is not needed given that we don't have limitations on concurrent runs due to GitHub enterprise.

.github/workflows/test-python.yaml
```diff
@@ -5,25 +5,52 @@ on:
...
 
       - name: Upload coverage to Coveralls
         uses: coverallsapp/github-action@v2
         with:
-          github-token: ${{ secrets.GITHUB_TOKEN }} # Pass the GITHUB_TOKEN
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: ${{ matrix.dependency-group.name }}-py${{ matrix.python-version }}
+          parallel: true
+
+  coverage:
```
andreyvelich: Don't we run coverage as part of the previous job in Upload coverage to Coveralls step ?

andreyvelich: As I can see, it shows the report on the PRs: https://github.com/kubeflow/sdk/pull/56#issuecomment-3144804385

andreyvelich (commented):
Thank you for this @kramaranya!

.github/workflows/test-python.yaml
astefanutti: A use case is re-pushing the branch for the PR while the workflow is running. Even with no limitations, is that useful to keep the existing run going?

python/pyproject.toml
kramaranya: Yeah, I've been thinking about adding this to `project.optional-dependencies` as well, but since we don't update API models that often, it makes sense to me to keep models from master available only for testing. However, I'm open to other suggestions if you see it differently

Makefile
kramaranya: > Since we can't introduce any breaking changes to the Kubernetes API

Hmm, why can't we?

python/pyproject.toml
andreyvelich: From my point of view, since we allow to install control plane with the latest changes, users might want to experiment with it during the development: https://www.kubeflow.org/docs/components/trainer/operator-guides/installation/#installing-the-kubeflow-trainer-controller-manager

This gives opportunities for them to quickly adopt newer API changes into their systems.

Makefile
andreyvelich: According to Kubernetes API best practices, if we introduce any breaking changes to the existing API: `v1alpha1`, we have to create the newer API version, for instance: `v1alpha2`: https://kubernetes.io/docs/reference/using-api/deprecation-policy/

cc @tenzen-y @astefanutti 

In the future, we should discuss how to design our SDK to make it compatible with supported API versions, but we are not there yet.

.github/workflows/test-python.yaml
andreyvelich: Might be useful to compare results, but I am fine to cancel them for now.
@kramaranya Please can you apply the change to all GitHub action tests in that case ?

python/pyproject.toml
andreyvelich, line 1: Let's also migrate to the official package as part of this PR: https://pypi.org/project/kubeflow-trainer-api/

python/pyproject.toml
kramaranya: Makes sense, let me update it

kramaranya: Updated in https://github.com/kubeflow/sdk/pull/55/commits/4d8967f78c96b88b84d82eb76064aefcd6eda186

Makefile
kramaranya: Thanks a lot for explanation! That makes sense now to test it only against master branch
Updated in https://github.com/kubeflow/sdk/pull/55/commits/4d8967f78c96b88b84d82eb76064aefcd6eda186

.github/workflows/test-python.yaml
kramaranya: Yeah, I added the parallel + “coverage-finished” step there when the matrix had four legs (dev/prod × two Python versions), so Coveralls could wait for every flag upload before closing the build and merging reports -- https://docs.coveralls.io/parallel-builds

But I've removed it in https://github.com/kubeflow/sdk/pull/55/commits/189139de3ca5334e39cca55c22ba3a3bd3ded433 since we're now testing against the master modules only

.github/workflows/test-python.yaml
kramaranya: Sure, updated in https://github.com/kubeflow/sdk/pull/55/commits/4853da44bf4ebe0c80cfb9a976b5de213df5f0f5

kramaranya:
@andreyvelich @astefanutti thank you for the review!
I should have addressed all your comments and imported `kubeflow-trainer-api` from PyPI https://github.com/kubeflow/sdk/blob/e84a54fcd70bc4aeda3008e76fdee9d104d888f7/python/pyproject.toml#L31.
Please let me know if there's anything else I should update!

astefanutti:
/lgtm

Thanks @kramaranya!

CONTRIBUTING.md
```diff
@@ -28,6 +28,12 @@ cd python
...
 
+#### Development Build (Optional)
+To install the latest API modules directly from the master branch:
+```sh
+pip install -e ".[dev]"
+```
+
 Install development tools:
 ```sh
 pip install pytest black isort flake8 coverage pre-commit
```
andreyvelich: I guess, we can remove it since we have everything in dev, right ?

Makefile
```diff
@@ -66,10 +66,10 @@ uv-venv:
 		echo "uv virtual environment already exists in $(VENV_DIR)."; \
 	fi
 
- # make test-unit will produce html coverage by default. Run with `make test-unit report=xml` to produce xml report.
+ # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv
-	@uv pip install "./python[test]"
+	@uv pip install --project $(PY_DIR) --group dev "$(PY_DIR)[test,dev]"
```
andreyvelich: Please can we [move `ruff`](https://github.com/kubeflow/sdk/blob/e84a54fcd70bc4aeda3008e76fdee9d104d888f7/python/pyproject.toml#L60) and [`test` ](https://github.com/kubeflow/sdk/blob/e84a54fcd70bc4aeda3008e76fdee9d104d888f7/python/pyproject.toml#L34) packages to `dev` too ?

In that case we have all tools required for development in one place

kramaranya: Sure, shall we also add `pre-commit` package to `dev`?

andreyvelich: Yes, sure!

kramaranya: great, updated in https://github.com/kubeflow/sdk/pull/55/commits/cb1c4dfd862f92cfaa6bafee308f30dcc18821a3!

CONTRIBUTING.md
kramaranya: yep, updated in https://github.com/kubeflow/sdk/pull/55/commits/cb1c4dfd862f92cfaa6bafee308f30dcc18821a3

kramaranya:
@andreyvelich thank you for the review! Let me please know if there's anything else I should update!

Makefile
```diff
@@ -66,10 +66,10 @@ uv-venv:
 		echo "uv virtual environment already exists in $(VENV_DIR)."; \
 	fi
 
- # make test-unit will produce html coverage by default. Run with `make test-unit report=xml` to produce xml report.
+ # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv
-	@uv pip install "./python[test]"
+	@uv pip install --project $(PY_DIR) -e "$(PY_DIR)[dev]"
```
andreyvelich: Do we need this --project flag ?

kramaranya: I added it for project isolation, but since we have only one project right now, I can remove it

kramaranya: done in https://github.com/kubeflow/sdk/pull/55/commits/542ddcd946eab05ff4c4cf831d6f30cccb82ff3c

andreyvelich (commented):
Thanks for this @kramaranya!
/lgtm
/approve

PR #56, opened 2025-08-01
feat(trainer): Support Framework Labels in Runtimes

andreyvelich:
Fixes: https://github.com/kubeflow/sdk/pull/31, https://github.com/kubeflow/sdk/issues/29

I refactored SDK to support runtime labels and give users ability to override the default runtime images.

/assign @Electronic-Waste @astefanutti @kramaranya @jskswamy @ned2 @eoinfennessy @briangallagher


python/kubeflow/trainer/types/types.py
```diff
@@ -149,26 +149,32 @@ class BuiltinTrainer:
...
-    entrypoint: Optional[List[str]] = None
     accelerator_count: Union[str, float, int] = constants.UNKNOWN
+    __command: tuple[str, ...] = field(init=False, repr=False)
+
+    @property
+    def command(self) -> tuple[str, ...]:
+        return self.__command
+
+    def set_command(self, command: tuple[str, ...]):
+        self.__command = command
```
andreyvelich, line 179: I make this __command private, so it should not be accessed or modified by users.

python/kubeflow/trainer/constants/constants.py
```diff
@@ -120,27 +124,49 @@
...
+# The exec script to embed training function into container command.
+# __ENTRYPOINT__ depends on the MLPolicy, func_code and func_file is substituted in the `train` API.
+EXEC_FUNC_SCRIPT = textwrap.dedent(
+    """
+        read -r -d '' SCRIPT << EOM\n
+        {func_code}
+        EOM
+        printf "%s" \"$SCRIPT\" > \"{func_file}\"
+        __ENTRYPOINT__ \"{func_file}\""""
+)
```
andreyvelich, line 136: For runtimes with CustomTrainer, I insert the complete `command` when Runtime object is created: 

https://github.com/kubeflow/sdk/blob/f3e90202b9af447afd30e03fcace9f4b842c16cb/python/kubeflow/trainer/utils/utils.py#L155-L164

I think, this is much clear approach since we always have the final  container `command` in `runtime.trainer.__command`.

python/kubeflow/trainer/types/types.py
```diff
@@ -229,57 +235,3 @@ class Initializer:
 
     dataset: Optional[HuggingFaceDatasetInitializer] = None
     model: Optional[HuggingFaceModelInitializer] = None
-
-
-# The dict where key is the container image and value its representation.
-# Each Trainer representation defines trainer parameters (e.g. type, framework, entrypoint).
-# TODO (andreyvelich): We should allow user to overrides the default image names.
-ALL_TRAINERS: Dict[str, RuntimeTrainer] = {
```
astefanutti, line 239: Would keeping the existing mapping, but with the framework as key instead of the image, make things tidier / more structured rather than replacing that with individual strings?

python/kubeflow/trainer/constants/constants.py
```diff
@@ -53,6 +54,9 @@
 # For example, what PodTemplate must be overridden by TrainJob's .spec.trainer APIs.
 TRAINJOB_ANCESTOR_LABEL = "trainer.kubeflow.org/trainjob-ancestor-step"
 
+# The label key to identify ML framework that runtime uses (e.g. torch, deepspeed, torchtune, etc.)
+RUNTIME_FRAMEWORK_LABEL = "trainer.kubeflow.org/framework"
+
```
Electronic-Waste, line 59: Shall we also define `TRAINER_TYPE_LABEL`?

python/kubeflow/trainer/types/types.py
```diff
@@ -39,7 +39,7 @@ class CustomTrainer:
 
     func: Callable
     func_args: Optional[Dict] = None
-    packages_to_install: Optional[List[str]] = None
+    packages_to_install: Optional[list[str]] = None
```
Electronic-Waste, line 43: Why should we change from `List` to `list`?

python/kubeflow/trainer/types/types.py
andreyvelich: But we don't need it, do we ? Theoretically, users should be able to use any arbitrary framework they define in the runtime labels with CustomTrainer, even that we don't support upstream (e.g. scikit-learn, tf, etc.).
The important mapping is only for BuiltinTrainers, since we need to map the `framework` name and BuiltinTrainer `configs`: https://github.com/kubeflow/sdk/blob/f3e90202b9af447afd30e03fcace9f4b842c16cb/python/kubeflow/trainer/types/types.py#L154.

If you want, we can create mapping for BuiltinTrainer configs, but for now we have only 1 config (e.g. TorchTune).

python/kubeflow/trainer/types/types.py
```diff
@@ -149,26 +149,32 @@ class BuiltinTrainer:
     config: TorchTuneConfig
 
 
+# Change it to list: BUILTIN_CONFIGS, once we support more Builtin Trainer configs.
+TORCH_TUNE = (
```
Electronic-Waste, line 155: Do you prefer `TORCHTUNE` or `TORCH_TUNE`?

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -190,13 +194,13 @@ def train(
             # If users choose to use a custom training function.
             if isinstance(trainer, types.CustomTrainer):
                 trainer_crd = utils.get_trainer_crd_from_custom_trainer(
-                    trainer, runtime
+                    runtime, trainer
```
Electronic-Waste: Can you add validation here to ensure that users reference the right runtime, as we discussed here: https://cloud-native.slack.com/archives/C0742LDFZ4K/p1753710956860929

python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -190,13 +194,13 @@ def train(
...
                 )
 
             # If users choose to use a builtin trainer for post-training.
             elif isinstance(trainer, types.BuiltinTrainer):
                 trainer_crd = utils.get_trainer_crd_from_builtin_trainer(
-                    trainer, initializer
+                    runtime, trainer, initializer
```
Electronic-Waste: Same as above

python/kubeflow/trainer/types/types.py
astefanutti: OK that makes sense. It's more an internal implementation detail anyway.

python/kubeflow/trainer/types/types.py
andreyvelich: Python 3.9 introduce builtin support for those types: https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

python/kubeflow/trainer/types/types.py
Electronic-Waste: We should verify that:

1. CustomTrainer does not reference runtimes corresponding to BuiltinTrainer
2. BuiltTrainer only reference corresponding runtimes

python/kubeflow/trainer/constants/constants.py
andreyvelich: No, we don't really need it as we discussed here: https://github.com/kubeflow/trainer/pull/2761#discussion_r2244636856

python/kubeflow/trainer/types/types.py
andreyvelich: Maybe TORCH_TUNE is more accurate since we use camel case here: TorchTuneConfig.

python/kubeflow/trainer/api/trainer_client.py
andreyvelich: Sure, thanks for reminder!

andreyvelich (commented):
/approve

PR #57, opened 2025-08-05
feat(trainer): Add `get_runtime_packages()` API

andreyvelich:
This API is similar to what I've showed at KubeCon London talk: https://youtu.be/Fnb1a5Kaxgo?t=555

It prints list of pre-installed Python packages and GPU devices (if `nvidia-smi` is available).

/assign @astefanutti @kramaranya @Electronic-Waste 

/hold for review

### DIFF
--- a/python/kubeflow/trainer/api/trainer_client.py
+++ b/python/kubeflow/trainer/api/trainer_client.py
@@ -177,6 +177,12 @@ class TrainerClient:
         if isinstance(runtime.trainer, types.BuiltinTrainer):
             raise ValueError("Cannot get Runtime packages for BuiltinTrainer")
 
+        # Run mpirun only within the single process.
+        if runtime.trainer.command[0] == "mpirun":
+            mpi_command = list(constants.MPI_COMMAND)
+            mpi_command[1:3] = ["-np", "1"]
+            runtime.trainer.set_command(tuple(mpi_command))
+
         def print_packages():
             import subprocess
             import shutil
--- a/python/kubeflow/trainer/api/trainer_client.py
+++ b/python/kubeflow/trainer/api/trainer_client.py
@@ -186,7 +186,12 @@ class TrainerClient:
         def print_packages():
             import subprocess
             import shutil
+            import sys
 
+            # Print Python version.
+            print(f"Python: {sys.version}")
+
+            # Print Python packages.
             if shutil.which("pip"):
                 pip_list = subprocess.run(
                     ["pip", "list"], capture_output=True, text=True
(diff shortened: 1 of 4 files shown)


python/kubeflow/trainer/api/trainer_client.py
```diff
@@ -489,36 +557,28 @@ def wait_for_job_status(
...
-            for event in w.stream(
-                self.core_api.list_namespaced_pod,
-                self.namespace,
-                label_selector=constants.POD_LABEL_SELECTOR.format(trainjob_name=name),
-                timeout_seconds=timeout,
-            ):
-                # Check the status after event is generated for the TrainJob's Pods.
-                trainjob = self.get_job(name)
-                logger.debug(f"TrainJob {name}, status {trainjob.status}")
+        if polling_interval > timeout:
```
andreyvelich, line 560: @astefanutti I have to refactor the `wait_for_job_status()` API to perform polling as before.
The problem that I saw is when Pods are succeeded too fast and TrainJob controller doesn't [add the Complete condition to the `.status.conditions`.](https://github.com/kubeflow/sdk/blob/b435e07d860348965a9e79023e4e9705e6cd5992/python/kubeflow/trainer/api/trainer_client.py#L747-L752) 

Since we only watch for Pod events, we can't catch this event, and TrainJob is stuck in Running condition.

Alternatively, we can watch both for TrainJob + Pods with two Python threads, but I am not sure if it worths it.

What do you think @astefanutti ?

python/kubeflow/trainer/types/types.py
```diff
@@ -168,7 +168,8 @@ class RuntimeTrainer:
     trainer_type: TrainerType
     framework: str
     num_nodes: int = 1  # The default value is set in the APIs.
-    accelerator_count: Union[str, float, int] = constants.UNKNOWN
+    device: str = constants.UNKNOWN
+    device_count: str = constants.UNKNOWN
```
andreyvelich, line 172: I make it consistent with Step device and device_count.
I think, it looks better.

python/kubeflow/trainer/api/trainer_client.py
astefanutti: @andreyvelich I agree with you. This problem should be addressed when we'll have comprehensive TrainJob conditions. During the interim, better keep things simple in the SDK and refactor it once we'll have the new TrainJob conditions.

python/kubeflow/trainer/types/types.py
kramaranya: Sounds great. We should also update notebooks in trainer examples

andreyvelich (commented):
Thanks!
/approve

Issue #59, opened 2025-08-06
Add Model Registry Client to Kubeflow SDK


syntaxsdev:
### What you would like to be added?

Add [Model Registry Client](https://github.com/kubeflow/model-registry/tree/main/clients/python) (https://github.com/kubeflow/model-registry/tree/main/clients/python)

to Kubeflow SDK (unified SDK)

There are two ways to achieve this that follow the guides in the Kubeflow SDK proposal

### Approach 1 - Namespace package (PEP 420)
Suggested by @briangallagher 
Convert the SDK to a namespace package.
This would mean removing the `__init__.py` file from `kubeflow/python` folder.

Any future installed packages that also meet the criteria of the `kubeflow` namespace are importable, instead of being a regular package (last-in, first-out) in terms of importance.

### Approach 2 - Wrapped package
Create a `model_registry` package within the kubeflow SDK namespace, and then import Model Registry.

I **_prefer_** this approach because then in the future, additional KF SDK specific things can be done and can be wrapped around in this fashion. It also will not require Model Registry to change its import statements which could confuse users and possibly break other workflows

Open to other ideas.

### Why is this needed?

Start migration process of unified (hybrid) SDK and to meet expectations set in original SDK proposal.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

briangallagher:
Thanks Sidney. Another possible approach is to add support for the kubeflow namespace to model_registry while maintaining backward compatibility. So, for example, support both: 

```
from model_registry import ModelRegistry
from mr_openapi import OrderByField, SortOrder
```
and 
```
from kubeflow.model_registry import ModelRegistry
from kubeflow.mr_openapi import OrderByField, SortOrder
```

That would allow model registry client to be imported without the need to re-export and support Approach 1.

Hacky implementation of that is [here](https://github.com/briangallagher/model-registry/tree/namespaces)
test with: `pip install git+https://github.com/briangallagher/model-registry.git@namespaces#subdirectory=clients/python`

kahowell:
@briangallagher isn't it cleaner to instead keep all re-exporting contained in the kubeflow sdk? E.g.:

```python
# kubeflow/model_registry/__init__.py
from model_registry import ModelRegistry  # or even: from model_registry import *


__all__ = [
    "ModelRegistry",
]
```

as @syntaxsdev suggests?

What are the benefits of introducing the kubeflow name into the model-registry package structure?

kramaranya:
[...]

PR #60, opened 2025-08-06
[DRAFT] Add Model Registry to Unified SDK

syntaxsdev:

Adds Model Registry to KF SDK (as a baseline) by importable extra install.

This is inline with issue https://github.com/kubeflow/sdk/issues/59

Use case:
```py
from kubeflow.model_registry import ModelRegistryClient
```

We also have original `ModelRegistry` aliased
This immediately allows users to use surfaced ModelRegistryClient
```py
from kubeflow.model_registry import ModelRegistry
```

We do not have the `mr_openapi` added yet because it is not surfaced via MR client, so a separate fix will come later once surfaced into MR, but it is not super urgent since mr_openapi was never surfaced in the first place in the model registry client.

Fixes https://github.com/kubeflow/sdk/issues/59

**Checklist:**

### DIFF
--- a/python/kubeflow/model_registry/tests/test_import.py
+++ b/python/kubeflow/model_registry/tests/test_import.py
@@ -1,0 +1,102 @@
+import pytest
+import sys
+
+
+@pytest.mark.model_registry
+def test_import_model_registry_available():
+    """Test that ModelRegistry can be imported when model-registry package is available."""
+    try:
+        from kubeflow.model_registry import ModelRegistry, ModelRegistryClient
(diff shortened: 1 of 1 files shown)


kramaranya:
Thank you @syntaxsdev! Before we proceed with the implementation, I think we should first discuss which approach would be most efficient and maintainable long-term in the original issue #59

tarilabs:
I don't think this PR should go into Stale per description

kramaranya:
@syntaxsdev @jonburdo we can close this pr since we have #186, right?

syntaxsdev:
@kramaranya Closing

PR #61, opened 2025-08-06
chore: move pyproject.toml to root

kramaranya:
I've moved `pyproject.toml` to root and updated license path to align with https://peps.python.org/pep-0639/#add-string-value-to-license-key

/assign @andreyvelich @astefanutti @Electronic-Waste @tenzen-y


pyproject.toml
```diff
@@ -55,17 +56,15 @@ line-length = 100
...
 package = true
 
 [build-system]
 requires = ["hatchling"]
 build-backend = "hatchling.build"
 
-[tool.hatch.build.targets]
-packages = ["kubeflow"]
+[tool.hatch.build.targets.wheel]
+sources = ["python"]
```
kramaranya: `packages = ["kubeflow"]` worked when `pyproject.toml` was in the python/ directory because it could find kubeflow/ directly, but after moving `pyproject.toml` to the root, it would look for kubeflow/ at the root level which doesn't exist.

It works well with `sources = ["python"]` and python/kubeflow/ gets installed as kubeflow/ in site-packages with correct imports like `from kubeflow.trainer import ...` instead of `from python.kubeflow.trainer import ...` that would happen with `packages = ["python/kubeflow"]`.

Alternatively, we could move the source code from python/kubeflow/ to just kubeflow/ at the root and keep using `packages = ["kubeflow"]`.

andreyvelich: Let's remove the `python` directory since it is not required, and keep `packages=["kubeflow"]`

pyproject.toml
```diff
@@ -55,8 +56,6 @@ line-length = 100
 select = ["E", "F", "W"]
```
andreyvelich, line 56: Shall we also move version file to: `kubeflow/__init__.py` ?
https://github.com/kubeflow/sdk/blob/1c115c6a493b5926b7318c505377a9c045337791/pyproject.toml#L49-L50

kramaranya: Yep, that makes sense!

kramaranya:
I'll also update Trainer and Website repos to use `pip install git+https://github.com/kubeflow/sdk.git@main` once we merge this PR

Issue #24, opened 2025-01-09
[SDK] Use dictionary unpacking to pass trainer function arguments


astefanutti:
### What you would like to be added?

A more straightforward way to pass arguments to the training function, e.g.: 

```python
def train_func(batch_size=64,
               test_batch_size=1000,
               epochs=10,
               lr=1.0,
               gamma=0.7,
               dry_run=False,
               seed=1,
               log_interval=10,
               save_model=False):
...

client.train(
    trainer=Trainer(
        func=train_func,
        func_args={
            "batch_size": 64,
            "test_batch_size": 1000,
            "epochs": 10,
            "lr": 1.0,
            "gamma": 0.7,
            "dry_run": False,
            "seed": 1,
            "log_interval": 10,
            "save_model": False,
        },
    ),
)
```

### Why is this needed?

It feels closer to what would be typically written if the training script would run standalone.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

ned2:
very much +1 to this. I was going to post this suggestion myself, but found it here already.

I've been using Kubeflow Trainer actively for the last month (for the first time) and was a little surprised that the dictionary passed to `func_args` parameter wasn't unpacked when calling `func` (ie `func(**func_args)`) and is instead just passed as a single argument, which I find more awkward to work with. 

@andreyvelich, over in kubeflow/trainer#2384 you asked: _What advantages do you see to pass func args via kwargs to the function compare to existing execution?_ Here's my take on this.

Currently, if I want to pass parameters to a `custom_train_func`, I'm likely to do something like this:

```pthon
def custom_train_func(parameters):
    param1 = parameters.get("param1", param1_default_val)
    param2 = parameters.get("param2", param2_default_val)
    param3 = parameters.get("param3", param3_default_val)
    ...
```

Whereas when the dictionary is unpacked into keyword arguments, we have the following: 

```python
def custom_func(
    pram1: int = param1_default_val,
    param2: int = param2_default_val,
    param3: str = param3_default_val,
):
    ...
```

The proposed approach has the following advantages:
- **Readability**: the function signature documents the interface to your training function at a glance, no need to read the entire function to determine the available parameters and their defaults -- these could have been defined all the way throughout a function. Also, adding type hints to the function signature further improves the self-documenting nature of the code. 
[...]

PR #62, opened 2025-08-11
feat(trainer): Add support for param unpacking in the training function call

briangallagher:
Add support for param unpacking in the training function call.
Maintain backward compatibility and add a deprecation warning.

Fixes # [24](https://github.com/kubeflow/sdk/issues/24)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ uv-venv:
 .PHONY: test-python
 test-python: uv-venv
 	@uv sync
-	@uv run coverage run --source=kubeflow.trainer.backends.kubernetes.backend,kubeflow.trainer.utils.utils -m pytest ./kubeflow/trainer/backends/kubernetes/backend_test.py
+	@uv run coverage run --source=kubeflow.trainer.backends.kubernetes.backend,kubeflow.trainer.utils.utils -m pytest ./kubeflow/trainer/backends/kubernetes/backend_test.py ./kubeflow/trainer/utils/utils_test.py
 	@uv run coverage report -m kubeflow/trainer/backends/kubernetes/backend.py kubeflow/trainer/utils/utils.py
 ifeq ($(report),xml)
 	@uv run coverage xml
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -24,8 +24,8 @@ import multiprocessing
 import random
 import string
 import uuid
-from dataclasses import asdict, dataclass, field
-from typing import Any, Optional, Type
+from dataclasses import asdict
+from typing import Optional
 from unittest.mock import Mock, patch
 
 import pytest
@@ -34,28 +34,17 @@ from kubeflow_trainer_api import models
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
 from kubeflow.trainer.utils import utils
-from kubeflow.trainer.backends.kubernetes.types import KubernetesBackendConfig
 from kubeflow.trainer.backends.kubernetes.backend import KubernetesBackend
+from kubeflow.trainer.backends.kubernetes.types import KubernetesBackendConfig
+from kubeflow.trainer.test.common import TestCase
+from kubeflow.trainer.test.common import (
+    SUCCESS,
+    FAILED,
+    DEFAULT_NAMESPACE,
+    TIMEOUT,
+    RUNTIME,
+)
 
-
-@dataclass
-class TestCase:
-    name: str
-    expected_status: str
-    config: dict[str, Any] = field(default_factory=dict)
-    expected_output: Optional[Any] = None
-    expected_error: Optional[Type[Exception]] = None
-    __test__ = False
-
-
-# --------------------------
-# Constants for test scenarios
-# --------------------------
-TIMEOUT = "timeout"
-RUNTIME = "runtime"
-SUCCESS = "success"
-FAILED = "Failed"
-DEFAULT_NAMESPACE = "default"
 # In all tests runtime name is equal to the framework name.
 TORCH_RUNTIME = "torch"
 TORCH_TUNE_RUNTIME = "torchtune"
@@ -238,9 +227,9 @@ def get_custom_trainer(
             '\nif ! [ -x "$(command -v pip)" ]; then\n    python -m ensurepip '
             "|| python -m ensurepip --user || apt-get install python-pip"
             "\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet"
-            f"         --no-warn-script-location {pip_command} {packages_command}"
+            f" --no-warn-script-location {pip_command} {packages_command}"
             "\n\nread -r -d '' SCRIPT << EOM\n\nfunc=lambda: "
-            'print("Hello World"),\n\n<lambda>('
+            'print("Hello World"),\n\n<lambda>(**'
             "{'learning_rate': 0.001, 'batch_size': 32})\n\nEOM\nprintf \"%s\" "
             '"$SCRIPT" > "backend_test.py"\ntorchrun "backend_test.py"',
         ],
--- a/kubeflow/trainer/test/common.py
+++ b/kubeflow/trainer/test/common.py
@@ -1,0 +1,23 @@
+# Shared test utilities and types for Kubeflow Trainer tests.
+
+from dataclasses import dataclass, field
+from typing import Any, Optional, Type
+
+
+# Common status constants
+SUCCESS = "success"
+FAILED = "Failed"
+DEFAULT_NAMESPACE = "default"
+TIMEOUT = "timeout"
+RUNTIME = "runtime"
+
+
+@dataclass
+class TestCase:
+    name: str
+    expected_status: str = SUCCESS
+    config: dict[str, Any] = field(default_factory=dict)
+    expected_output: Optional[Any] = None
+    expected_error: Optional[Type[Exception]] = None
+    # Prevent pytest from collecting this dataclass as a test
+    __test__ = False
--- a/kubeflow/trainer/utils/utils.py
+++ b/kubeflow/trainer/utils/utils.py
@@ -15,9 +15,9 @@
 import inspect
 import os
 import textwrap
-from typing import Any, Callable, Optional
-from urllib.parse import urlparse
 
+from typing import Callable, Optional, Any
+from urllib.parse import urlparse
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
 from kubeflow_trainer_api import models
@@ -268,15 +268,19 @@ def get_script_for_python_packages(
     if is_mpi:
         options.append("--user")
 
-    script_for_python_packages = textwrap.dedent(
+    header_script = textwrap.dedent(
         """
         if ! [ -x "$(command -v pip)" ]; then
             python -m ensurepip || python -m ensurepip --user || apt-get install python-pip
         fi
 
-        PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \
-        --no-warn-script-location {} {}
-        """.format(
+        """
+    )
+
+    script_for_python_packages = (
+        header_script
+        + "PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet "
+        + "--no-warn-script-location {} {}\n".format(
             " ".join(options),
             packages_str,
         )
@@ -318,12 +322,16 @@ def get_command_using_train_func(
     # Wrap function code to execute it from the file. For example:
     # TODO (andreyvelich): Find a better way to run users' scripts.
     # def train(parameters):
-    #     print('Start Training...')
+    #   print('Start Training...')
     # train({'lr': 0.01})
     if train_func_parameters is None:
-        func_code = f"{func_code}\n{train_func.__name__}()\n"
+        func_call = f"{train_func.__name__}()"
     else:
-        func_code = f"{func_code}\n{train_func.__name__}({train_func_parameters})\n"
+        # Always unpack kwargs for training function calls.
+        func_call = f"{train_func.__name__}(**{train_func_parameters})"
+
+    # Combine everything into the final code string.
+    func_code = f"{func_code}\n{func_call}\n"
 
     is_mpi = runtime.trainer.command[0] == "mpirun"
     # The default file location for OpenMPI is: /home/mpiuser/<FILE_NAME>.py
--- a/kubeflow/trainer/utils/utils_test.py
+++ b/kubeflow/trainer/utils/utils_test.py
@@ -12,22 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass
-from typing import Any, Dict
-
 import pytest
 
 from kubeflow.trainer.utils import utils
 from kubeflow.trainer.constants import constants
+from kubeflow.trainer.types import types
+from kubeflow.trainer.test.common import TestCase, SUCCESS, FAILED
 
-
-@dataclass
-class TestCase:
-    name: str
-    config: Dict[str, Any]
-    expected_output: str
-    __test__ = False
-
+def _build_runtime() -> types.Runtime:
+    runtime_trainer = types.RuntimeTrainer(
+        trainer_type=types.TrainerType.CUSTOM_TRAINER,
+        framework="torch",
+        device="cpu",
+        device_count="1",
+    )
+    runtime_trainer.set_command(constants.DEFAULT_COMMAND)
+    return types.Runtime(name="test-runtime", trainer=runtime_trainer)
 
 @pytest.mark.parametrize(
     "test_case",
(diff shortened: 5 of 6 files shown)


astefanutti:
/lgtm

Very nice!

kubeflow/trainer/utils/utils.py
```diff
@@ -318,12 +320,44 @@ def get_command_using_train_func(
...
+        try:
+            from typing import get_type_hints
+            type_hints = get_type_hints(train_func)
+        except Exception:
+            type_hints = {}
+
+        call_with_single_dict = (
+            len(params) == 1
+            and (
+                type_hints.get(params[0].name, dict) is dict
```
kramaranya: Hmm, when there's no type hint, this returns `dict` as default value, which makes `dict is dict` always true, right?

kubeflow/trainer/utils/utils.py
```diff
@@ -318,12 +320,44 @@ def get_command_using_train_func(
...
+                or params[0].annotation is inspect.Parameter.empty
```
kramaranya: This treats all functions without type annotations as deprecated style, what about smth like `def train(learning_rate=0.1)`?

kramaranya (commented):
Thank you @briangallagher!

kubeflow/trainer/utils/utils.py
briangallagher: Yes, that's right but it is in combination with len(params) == 1.
So, it favours the original dictionary param over the unpacked params, in the scenario where we can not be definitive. I'm open to a more definitive approach to distinguish the user intention?

kubeflow/trainer/utils/utils.py
briangallagher: Yes, that is an edge case but I think we should catch most cases. See previous comment, open to a more definitive approach if possible.

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,156 @@
...
+        trainer_type=types.TrainerType.CUSTOM_TRAINER,
+        framework="torch",
+        device="cpu",
+        device_count="1",
+    )
+    runtime_trainer.set_command(constants.DEFAULT_COMMAND)
+    return types.Runtime(name="test-runtime", trainer=runtime_trainer)
+
+
+def test_get_command_using_train_func_with_kwargs_unpacking():
```
andreyvelich: @briangallagher Please can you refactor these test cases similar to trainer_client_test: https://github.com/kubeflow/sdk/blob/7b748fb3544762468d2734cae616958b3d15ea17/kubeflow/trainer/api/trainer_client_test.py#L577-L600

With TestCases for each function: `test_get_command_using_train_func()`

kubeflow/trainer/utils/utils.py
```diff
@@ -318,12 +320,44 @@ def get_command_using_train_func(
     # Wrap function code to execute it from the file. For example:
     # TODO (andreyvelich): Find a better way to run users' scripts.
     # def train(parameters):
```
andreyvelich, line 324: As I asked here, please can you ensure that this new approach will work well with `optimize()` API: https://github.com/kubeflow/sdk/issues/24#issuecomment-3168514509

kubeflow/trainer/utils/utils.py
```diff
@@ -318,12 +320,44 @@ def get_command_using_train_func(
...
+        if call_with_single_dict:
+            message = (
+                f"Calling '{train_func.__name__}' with a single dict argument is deprecated."
+                " Please update your function to accept individual keyword arguments."
+            )
+            warnings.warn(
+                message,
+                DeprecationWarning,
+                stacklevel=2,
+            )
```
andreyvelich: We can put the msg directly to the warning.

```suggestion
            warnings.warn(
                f"Calling '{train_func.__name__}' with a single dict argument is deprecated."
                " Please update your function to accept individual keyword arguments."
                DeprecationWarning,
                stacklevel=2,
            )
```

kubeflow/trainer/utils/utils.py
andreyvelich: @briangallagher Are we making this only for deprecation policies ?
Since we haven't released the SDK yet, we can just update the examples with the newer approach.

kramaranya:
I think it would be great to include this in release 0.1, any thoughts? @briangallagher @andreyvelich @astefanutti

kubeflow/trainer/utils/utils.py
briangallagher: I created an [issue ](https://github.com/kubeflow/trainer/issues/2801) for trainer examples to be updated. 
@andreyvelich Do you want me to remove the warning completely? Add a simple print warning maybe?

kubeflow/trainer/utils/utils.py
briangallagher: Hi @andreyvelich I can see what Katib is current doing regarding substituting parameters in the controller but can I clarify what you're asking for here in the context of Trainer? Is it to replicate this functionality for Trainer i.e. write the train func to a file, pass command line arguments and allow the controller to override? 
This would require changes to the SDK and Operator? Could it be done in a dedicated issue?

kubeflow/trainer/utils/utils.py
briangallagher: @andreyvelich Your comment above seems to suggest removing this entirely?

kubeflow/trainer/utils/utils.py
andreyvelich: @briangallagher I would remove everything if that won't break E2Es in this PR.
You can update the website + examples after this PR gets merged .

kubeflow/trainer/utils/utils.py
andreyvelich: I think, that Trainer class will also be part of `optimize()` API in the future cc @astefanutti @kramaranya 
So my question is, will the following API work for us ?

```py
client.optimize(
    trainer=Trainer(
        func=train_func,
        func_args={
            "epochs": optimizer.search.int("10", "15"),
            "batch_size":  optimizer.search.int("128", "160"),
        },
    ),
)
```

briangallagher: @andreyvelich I don't see any issue with that proposal and this PR but I think we can dig deeper once the proposal for the new APIs are in place?

kubeflow/trainer/utils/utils.py
briangallagher: I remove this now and yes, I can update the website and examples

kubeflow/trainer/utils/utils_test.py
briangallagher: test is updated now to be similar style to trainer_client_test.py 
It doesn't cover the full utils.py module but the test can evolve over time to increase coverage.

kramaranya:
@briangallagher could you rebase the PR please?

kubeflow/trainer/utils/utils.py
andreyvelich: Sounds good @briangallagher!

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,139 @@
...
+class TestCase:
+    name: str
+    expected_status: str
+    func: Any
+    func_args: Optional[Dict[str, Any]]
+    expect_contains: List[str] = field(default_factory=list)
+    expect_not_contains: List[str] = field(default_factory=list)
+    runtime_override: Optional[types.Runtime] = None
+    expected_error: Optional[Type[Exception]] = None
+    __test__ = False
```
andreyvelich: Similar to this comment: https://github.com/kubeflow/sdk/pull/79#discussion_r2314607545
Shall we use unified structure of TestCase like here: https://github.com/kubeflow/sdk/blob/47fed69c9b46663e9f70dad078affb24518b8199/kubeflow/trainer/backends/kubernetes/backend_test.py#L41C1-L49C1?
```py
@dataclass
class TestCase:
    name: str
    expected_status: str
    config: dict[str, Any] = field(default_factory=dict)
    expected_output: Optional[Any] = None
    expected_error: Optional[Type[Exception]] = None
    __test__ = False
```

Since we can re-use this TestCase class for other tests in `utils_test.py`.

kubeflow/trainer/utils/utils.py
```diff
@@ -316,12 +316,35 @@ def get_command_using_train_func(
...
+        # Decide whether to pass a single dict or unpack as kwargs using the same hinting logic.
+        sig = inspect.signature(train_func)
+        params = list(sig.parameters.values())
+        try:
+            from typing import get_type_hints
+            type_hints = get_type_hints(train_func)
+        except Exception:
+            type_hints = {}
+
+        call_with_single_dict = (
```
andreyvelich: Why do you need call_with_single_dict check ?
As we discussed before, users should set the parameters via kwargs always, isn't ?
```py
trainer=CustomTrainer(
    func=train_pytorch,
    func_args={
        "batch_size": 123 
    },
)
```

briangallagher: @andreyvelich  I misunderstood your previous comment as removed the warning entirely i.e. Keep the logic but just removed the warning. 
It's updated now, always unpack.

kubeflow/trainer/utils/utils_test.py
briangallagher: @andreyvelich Added a common test module, moved TestCase and some common constants there.

Makefile
```diff
@@ -67,8 +67,8 @@ uv-venv:
  # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv
-	@uv sync
-	@uv run coverage run --source=kubeflow.trainer.backends.kubernetes.backend,kubeflow.trainer.utils.utils -m pytest ./kubeflow/trainer/backends/kubernetes/backend_test.py
+	@uv sync --all-groups
```
andreyvelich: Why do you need it ?

kubeflow/trainer/tests/common.py
```diff
@@ -0,0 +1,23 @@
+# Shared test utilities and types for Kubeflow Trainer tests.
```
andreyvelich: Maybe we can name this directory `test` similar to Trainer: https://github.com/kubeflow/trainer/tree/master/test.

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,143 @@
...
+    return "ok"
+
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="with args dict always unpacks kwargs",
+            expected_status=SUCCESS,
+            config={
+                "func": train_func_single_arg,
```
andreyvelich: Is there a ways for us to define this function as lambda, like here: https://github.com/kubeflow/sdk/blob/5304c5d0abedb25358af5c48586730647fe8d438/kubeflow/trainer/backends/kubernetes/backend_test.py#L714 ?

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,143 @@
...
+                "func_args": {"batch_size": 128, "learning_rate": 0.001, "epochs": 20},
+            },
+            expected_output={
+                "contains": [
+                    "train_func_single_arg(**",
+                    str({"batch_size": 128, "learning_rate": 0.001, "epochs": 20}),
+                ],
+                "not_contains": [],
```
andreyvelich: Why do you need to have this `non_contains` check ?

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,143 @@
...
+                    str({"a": 3, "b": "hi", "c": 0.2}),
+                ],
+                "not_contains": [],
+            },
+        ),
+    ],
+)
+def test_get_command_using_train_func(test_case: TestCase):
+    print("Executing test:", test_case.name)
+    runtime = test_case.config.get("runtime") or _build_runtime()
```
andreyvelich: Can you just define `runtime` in every test case, so you don't need this line ?

andreyvelich (commented):
We might need to manually merge this PR, since tests are failing.
@briangallagher Did you verify that your SDK changes are working ?

Makefile
briangallagher: to install the dev dependencies. 
I've updated it to `uv sync --group dev`

kubeflow/trainer/tests/common.py
briangallagher: done

kubeflow/trainer/utils/utils_test.py
briangallagher: done

kubeflow/trainer/utils/utils_test.py
briangallagher: not needed, removed

kubeflow/trainer/utils/utils_test.py
briangallagher: done

Makefile
```diff
@@ -67,8 +67,8 @@ uv-venv:
  # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv
-	@uv sync
-	@uv run coverage run --source=kubeflow.trainer.backends.kubernetes.backend,kubeflow.trainer.utils.utils -m pytest ./kubeflow/trainer/backends/kubernetes/backend_test.py
+	@uv sync --group dev
```
kramaranya: `uv sync` includes dev group by default https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups
```suggestion
	@uv sync
```

kubeflow/trainer/test/common.py
```diff
@@ -0,0 +1,23 @@
+# Shared test utilities and types for Kubeflow Trainer tests.
+
+from dataclasses import dataclass, field
+from typing import Any, Dict, Optional, Type
```
kramaranya: Can we use built in dict?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -34,28 +34,19 @@
...
-
-# --------------------------
-# Constants for test scenarios
-# --------------------------
-TIMEOUT = "timeout"
-RUNTIME = "runtime"
-SUCCESS = "success"
-FAILED = "Failed"
-DEFAULT_NAMESPACE = "default"
+"""Constants imported from kubeflow.trainer.tests.common"""
```
kramaranya:
```suggestion
"""Constants imported from kubeflow.trainer.test.common"""
```

kramaranya: That should be just a comment, right?

kramaranya (commented):
Thank you @briangallagher!
I just left a few nits

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,135 @@
...
+        )
+
+        assert test_case.expected_status == SUCCESS
+        assert len(command) == 3 and command[0] == "bash" and command[1] == "-c"
+        script_content = command[2]
+
+        for token in (test_case.expected_output or {}).get("contains", []):
+            assert token in script_content
+        for token in (test_case.expected_output or {}).get("not_contains", []):
+            assert token not in script_content
```
andreyvelich: you can remove it, right ?
```suggestion
```

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,135 @@
...
+            name="without args calls function with no params",
+            expected_status=SUCCESS,
+            config={
+                "func": (lambda: print("Hello World")),
+                "func_args": None,
+                "runtime": _build_runtime(),
+            },
+            expected_output={
+                "contains": ["<lambda>()"],
+                "not_contains": ["(**"],
```
andreyvelich:
```suggestion
```

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,135 @@
...
+                "func": (lambda: print("Hello World")),
+                "func_args": {"batch_size": 128, "learning_rate": 0.001, "epochs": 20},
+                "runtime": _build_runtime(),
+            },
+            expected_output={
+                "contains": [
+                    "<lambda>(**",
+                    str({"batch_size": 128, "learning_rate": 0.001, "epochs": 20}),
+                ],
+            },
```
andreyvelich: Please can you follow the same output as here: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/utils/utils_test.py#L46C13-L56 ?
Just define the complete output in the `expected_output` argument and assert it with returns from get_command_using_train_func().
```suggestion
            expected_output=[
                    "bash", "-c" "torchrun", "<lambda>(**....")", ....
            ]
```

briangallagher:
> We might need to manually merge this PR, since tests are failing. @briangallagher Did you verify that your SDK changes are working ?

Yes, tested and working

Makefile
briangallagher: removed now

kubeflow/trainer/test/common.py
briangallagher: yep, done

kubeflow/trainer/backends/kubernetes/backend_test.py
briangallagher: comment removed

kubeflow/trainer/utils/utils_test.py
briangallagher: done

kubeflow/trainer/test/common.py
```diff
@@ -0,0 +1,23 @@
...
+TIMEOUT = "timeout"
+RUNTIME = "runtime"
+
+
+@dataclass
+class TestCase:
+    name: str
+    expected_status: str
+    config: dict[str, Any] = field(default_factory=dict)
+    expected_output: Any | None = None
```
kramaranya: I think this requires Python 3.10 https://peps.python.org/pep-0604/

andreyvelich: Yes, let's use Optional[Any] = None for now, since we still support Python 3.9

andreyvelich:
@briangallagher You might need to rebase this PR once more time

kubeflow/trainer/test/common.py
briangallagher: Sorry, changes pushed now.

kubeflow/trainer/utils/utils_test.py
```diff
@@ -0,0 +1,123 @@
...
+            train_func=test_case.config.get("func"),
+            train_func_parameters=test_case.config.get("func_args"),
+            pip_index_url=constants.DEFAULT_PIP_INDEX_URL,
+        )
+
+        assert test_case.expected_status == SUCCESS
+        assert len(command) == 3 and command[0] == "bash" and command[1] == "-c"
+        script_content = command[2]
+        for token in (test_case.expected_output or []):
+            assert token in script_content
```
andreyvelich: As I asked in this comment, please can you move the entire command in the `expected_output`: https://github.com/kubeflow/sdk/pull/62#discussion_r2324629315
```suggestion
        assert test_case.expected_status == SUCCESS
        assert command == test_case.expected_output
```

briangallagher: done

andreyvelich (commented):
Thank you for this great contribution @briangallagher!
/lgtm
/assign @kramaranya @astefanutti @Electronic-Waste

Electronic-Waste (commented):
@briangallagher Thanks for this!

/lgtm

kubeflow/trainer/utils/utils_test.py
```diff
@@ -48,7 +48,7 @@ class TestCase:
                 '    python -m ensurepip || python -m ensurepip --user || '
                 'apt-get install python-pip\n'
                 'fi\n\n'
-                'PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet '
+                'PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet         '
```
kramaranya: Why do we need to do that?

kramaranya (commented):
Overall looks great! I just left one question

kubeflow/trainer/utils/utils_test.py
briangallagher: That's require because of how the pip output is formatted [here](https://github.com/kubeflow/sdk/blob/427b35d8c9ab502e5ba1c2b41cae41554f7777e1/kubeflow/trainer/utils/utils.py#L277)
The reason I have it in this PR is because these tests never actually ran on the previous PR i.e. utils_tests.py was never added to the test-python Make command so I don't think it ever worked.  

I've updated this PR now to change the formatting of the "PIP..." output so these spaces are no longer required.

andreyvelich (commented):
Thanks @briangallagher!
/lgtm
/approve

kubeflow/trainer/utils/utils_test.py
kramaranya: I see, thank you!

kramaranya:
Thank you @briangallagher!
/lgtm

PR #64, opened 2025-08-18
fix(trainer): Keep the original runtime command in get_runtime_packages() API

andreyvelich:
We should not modify the runtime's trainer command when user calls `get_runtime_packages()` API.

/assign @kubeflow/wg-training-leads @astefanutti @kramaranya @szaher


Electronic-Waste (commented):
@andreyvelich Thanks for this!
/approve
/lgtm

PR #65, opened 2025-08-21
feat(ci): Add automated release CI job

kramaranya:
I added a CI job for release process automation:

The `Prepare Release` GitHub Action automatically:

- Detects the version change in `kubeflow/__init__.py`
- Creates or updates the `release-X.Y` branch
- Cherry-picks the version bump commit from main

The `Release` GitHub Action automatically:

- Runs tests and builds the package
- Creates and pushes the release tag
- Publishes to PyPI (requires manual approval)
- Creates GitHub Release (requires manual approval)

I also added the script to generate CHANGELOG and update the version.
And added Release README

I've tested this workflow on my branch - https://github.com/kramaranya/sdk/actions/runs/17457296258
PyPI project - https://pypi.org/project/kubeflow-test/
GitHub release - https://github.com/kramaranya/sdk/releases/tag/v0.8.1


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,67 @@
...
+          BRANCH=release-$MAJOR_MINOR
+
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "branch=$BRANCH" >> $GITHUB_OUTPUT
+
+      - name: Check if tag already exists
+        shell: bash
+        run: |
+          if git ls-remote --tags origin "${{ steps.vars.outputs.version }}" | grep -q "refs/tags/${{ steps.vars.outputs.version }}"; then
+            echo "Tag ${{ steps.vars.outputs.version }} already exists. Skipping CI."; exit 0; fi
```
briangallagher: Is the intention here to continue to the cherry pick even if the tag exists?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,67 @@
...
+          VERSION="${{ steps.vars.outputs.version }}"
+          BRANCH="${{ steps.vars.outputs.branch }}"
+
+          MAIN_SHA="${{ github.sha }}"
+
+          if git ls-remote --heads origin "$BRANCH" | grep -q "$BRANCH"; then
+            echo "Using existing branch: $BRANCH"
+            git fetch origin "$BRANCH":"$BRANCH"
+            git checkout "$BRANCH"
+            if ! git cherry-pick -x "$MAIN_SHA"; then
```
briangallagher: Small suggestion: Add a check for the SHA before cherry-pick:
something like: 

```
if git merge-base --is-ancestor "$MAIN_SHA" "$BRANCH"; then
  echo "Commit already present in $BRANCH"
else
  git cherry-pick -x "$MAIN_SHA"
fi
```

Makefile
```diff
@@ -64,6 +64,13 @@ uv-venv:
 		echo "uv virtual environment already exists in $(VENV_DIR)."; \
 	fi
 
+.PHONY: release
+release: uv-venv
+	@if [ -z "$(VERSION)" ]; then echo "Usage: make release VERSION=v0.1.0"; exit 1; fi
+	@VERSION_NUM=$$(echo "$(VERSION)" | sed 's/^v//'); \
+	sed -i '' "s/__version__ = \".*\"/__version__ = \"$$VERSION_NUM\"/" kubeflow/__init__.py; \
```
briangallagher: Will this work on non Mac?

.github/workflows/prepare-release.yml
kramaranya: Yes initially, but now I think it makes sense to skip the cherry pick as well if the tag already exists
Let me update it!

.github/workflows/prepare-release.yml
kramaranya: Good point, thank you!

Makefile
kramaranya: Good catch! I created a small python script instead

kramaranya:
@briangallagher I've addressed your comments, thank you for review!

/assign @briangallagher @andreyvelich @Electronic-Waste @astefanutti

Makefile
```diff
@@ -64,6 +64,12 @@ uv-venv:
 		echo "uv virtual environment already exists in $(VENV_DIR)."; \
 	fi
 
+.PHONY: release
+release: uv-venv
```
andreyvelich: Shall we just use `install-dev` here ?
```suggestion
release: install-dev
```

scripts/gen-changelog.py
```diff
@@ -0,0 +1,144 @@
+import argparse
+import os
+import sys
+
```
andreyvelich: @kramaranya Please can you explore whether we can leverage the Kubernetes release-notes CLI to generate changelog: https://github.com/kubernetes/release/tree/master/cmd/release-notes ?
I know that Kueue uses it:
```
release-notes --org kubernetes-sigs --repo kueue --branch release-0.3 \
--start-sha 4a0ebe7a3c5f2775cdf5fc7d60c23225660f8702 --end-sha a51cf138afe65677f5f5c97f8f8b1bc4887f73d2 \
--dependencies=false --required-author=""
```

I’m wondering if there’s an existing tool or workflow we could re-use to reduce the overhead of maintaining our own scripts for changelog generation.


cc @tenzen-y @kannon92 @mimowo

scripts/set-version.py
```diff
@@ -0,0 +1,21 @@
+import argparse
```
andreyvelich: Maybe you directly define this in the `make release` instruction. It looks like very simple change to update `kubeflow/__init__.py` file?
It is Kubernetes best practice to use shell scripts where it is possible, since it doesn't require to install Python environment.

.github/workflows/test-python.yaml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
   push:
     branches: [main]
+  workflow_call:
```
andreyvelich: Why do you need this ?

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,76 @@
+name: Prepare Release
```
andreyvelich: @kramaranya Can we combine this workflow into single action? I don't think that we need this guard check.
For example:
```yaml
name: Release

on:
  push:
    branches: [ main ]
    paths:
      - 'kubeflow/__init__.py'
  workflow_dispatch: {}

permissions:
  contents: write
  id-token: write

jobs:
  prepare:
    runs-on: ubuntu-latest
    outputs:
      version: ${{ steps.prepare.outputs.version }}
      branch: ${{ steps.prepare.outputs.branch }}
      is-prerelease: ${{ steps.prepare.outputs.is-prerelease }}
    steps:
      - uses: actions/checkout@v4
        with: { fetch-depth: 0 }

      - id: prepare
        run: |
          set -euo pipefail

          # --- Extract version and branch ---
          VERSION=v$(sed -n 's/^__version__ = "\(.*\)"/\1/p' kubeflow/__init__.py)
          MAJOR_MINOR=$(echo "$VERSION" | sed 's/^v//' | cut -d. -f1,2)
          BRANCH=release-$MAJOR_MINOR

          echo "version=$VERSION" >> $GITHUB_OUTPUT
          echo "branch=$BRANCH" >> $GITHUB_OUTPUT

          if [[ "$VERSION" =~ -rc($|\.) ]]; then
            echo "is-prerelease=true" >> $GITHUB_OUTPUT
          else
            echo "is-prerelease=false" >> $GITHUB_OUTPUT
          fi

          # --- Ensure release branch exists ---
          MAIN_SHA="${{ github.sha }}"
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"

          if git ls-remote --heads origin "$BRANCH" | grep -q "$BRANCH"; then
            echo "Branch $BRANCH exists. Updating..."
            git fetch origin "$BRANCH":"$BRANCH"
            git checkout "$BRANCH"
            git merge --no-edit "$MAIN_SHA" || true
          else
            echo "Creating new branch $BRANCH at $MAIN_SHA"
            git checkout -B "$BRANCH" "$MAIN_SHA"
          fi

          git push origin "$BRANCH"

  create-tag:
    runs-on: ubuntu-latest
    needs: prepare
    environment:
      name: release
      url: https://pypi.org/project/kubeflow/
    steps:
      - uses: actions/checkout@v4
        with: { fetch-depth: 0 }
      - run: |
          VERSION="${{ needs.prepare.outputs.version }}"
          BRANCH="${{ needs.prepare.outputs.branch }}"
          if ! git ls-remote --tags origin "$VERSION" | grep -q "refs/tags/$VERSION"; then
            git fetch origin "$BRANCH":"$BRANCH"
            git checkout "$BRANCH"
            git tag "$VERSION"
            git push origin "$VERSION"
          else
            echo "Tag $VERSION already exists. Skipping."
          fi

  build:
    runs-on: ubuntu-latest
    needs: create-tag
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with: { python-version: '3.11' }
      - run: |
          set -euo pipefail
          make verify
          uv build
          uvx twine check dist/*
      - uses: actions/upload-artifact@v4
        with:
          name: dist-${{ needs.prepare.outputs.version }}
          path: dist/

  publish-pypi:
    runs-on: ubuntu-latest
    needs: build
    environment:
      name: release
      url: https://pypi.org/project/kubeflow/
    steps:
      - uses: actions/download-artifact@v4
        with:
          name: dist-${{ needs.prepare.outputs.version }}
          path: dist/
      - uses: pypa/gh-action-pypi-publish@release/v1
        with: { verbose: true }

  github-release:
    runs-on: ubuntu-latest
    needs: [build, publish-pypi]
    environment:
      name: release
      url: https://github.com/${{ github.repository }}/releases
    steps:
      - uses: actions/checkout@v4
      - id: changelog
        run: |
          VERSION="${{ needs.prepare.outputs.version }}"
          MAJOR_MINOR=$(echo "$VERSION" | sed 's/^v//' | cut -d. -f1,2)
          FILE="CHANGELOG/CHANGELOG-$MAJOR_MINOR.md"
          if [[ -f "$FILE" ]]; then
            HEADER="^# \\[${VERSION//./\\.}\\]"
            SECTION=$(sed -n "/$HEADER/,\$p" "$FILE" | tail -n +2)
            NEXT=$(echo "$SECTION" | grep -m1 "^# \\[v[0-9]" || true)
            if [[ -n "$NEXT" ]]; then
              CHANGELOG=$(echo "$SECTION" | sed -n "1,/^# \\[v[0-9]/p" | sed '1d;$d')
            else
              CHANGELOG=$(echo "$SECTION" | sed '1d')
            fi
            echo "changelog<<EOF" >> $GITHUB_OUTPUT
            echo "$CHANGELOG" >> $GITHUB_OUTPUT
            echo "EOF" >> $GITHUB_OUTPUT
          else
            echo "changelog=**Note:** $FILE not found" >> $GITHUB_OUTPUT
          fi
      - uses: softprops/action-gh-release@v1
        with:
          tag_name: ${{ needs.prepare.outputs.version }}
          name: Kubeflow SDK ${{ needs.prepare.outputs.version }}
          body: ${{ steps.changelog.outputs.changelog }}
          prerelease: ${{ needs.prepare.outputs.is-prerelease == 'true' }}
          generate_release_notes: true
```

Makefile
kramaranya: Sure, makes sense

scripts/set-version.py
kramaranya: Sure, let me update this

.github/workflows/test-python.yaml
kramaranya: I was using this to make this workflow reusable in other workflows (like release or prepare release). But I've updated this to just use `make test-python` inside `release.yaml` and removed `workflow_call:` from `test-python`

Makefile
andreyvelich: @kramaranya FYI, if we want to make `sed` command align between Mac and Linux, we can install `gnu-sed`, like in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/Makefile#L72-L81

scripts/gen-changelog.py
kramaranya: I think we can make that work, but the issue is that it relies on PR labels for categorization (like `kind/feature`, `kind/bug`, etc) and also needs a dedicated release notes line in the PR description.
We can manually add labels to previous PRs, but we can't modify their descriptions. So maybe we could use our current release script for this first release, then going forward adapt new PRs to follow the k8s release script format with proper PR templates?
Alternatively, I found this tool (https://github.com/orhun/git-cliff) that might work well for us since it uses PR titles and can categorize them
WDYT?

.github/workflows/prepare-release.yml
kramaranya: Sounds good to me, I've updated the workflow

Makefile
kramaranya: I see, thank you! I think `sed -i.bak` works as well, shall we just keep as it is right now and don't introduce any extra dependency?

scripts/gen-changelog.py
andreyvelich: > Then going forward adapt new PRs to follow the k8s release script format with proper PR templates?

Yes, that makes sense, we can adopt the k8s model for the future releases. I just want us to not maintain script if we can re-use the existing process.

>Alternatively, I found this tool (https://github.com/orhun/git-cliff) that might work well for us since it uses PR titles and can categorize them

Interesting, do you know how it defined categories ? Just features/bug fixes/documentation ?

Makefile
andreyvelich: @kramaranya  `sed -i.bak` will work for MacOS and Linux ?

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
+# Releasing the Kubeflow SDK
```
andreyvelich: Shall we move this file to `RELEASE.md` ? I noticed that many OSS projects doing this, and it is easier to find.

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+- [Write](https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-permission-levels-for-an-organization#permission-levels-for-repositories-owned-by-an-organization)
+  permission for the Kubeflow SDK repository.
+
+- Create a [GitHub Token](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) and set it as `GITHUB_TOKEN` environment variable.
+
+- Install development dependencies:
+
+  ```bash
+  uv sync
+  ```
```
andreyvelich: Is that required since we just need to update the version and generate changelog ?

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Versioning Policy
+
+Kubeflow SDK version format follows [Semantic Versioning](https://semver.org/).
+Kubeflow SDK versions are in the format of `vX.Y.Z`, where `X` is the major version, `Y` is
```
andreyvelich: Since Python versions don't support `v`, I am wondering if we should use `X.Y.Z` versions for Kubeflow SDK ?

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+the minor version, and `Z` is the patch version.
+The patch version contains only bug fixes.
+
+Additionally, Kubeflow SDK does pre-releases in this format: `vX.Y.Z-rc.N` where `N` is a number
```
andreyvelich: Should this be ?
`X.Y.ZrcN`

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+   make release VERSION=vX.Y.Z
+   ```
+
+This updates:
+- `kubeflow/__init__.py` with `__version__ = "X.Y.Z"`
+- `CHANGELOG/CHANGELOG-X.Y.md` with a new top entry `# [vX.Y.Z] (YYYY-MM-DD)`
+
+2. Open a PR:
+   - Review `kubeflow/__init__.py` and `CHANGELOG/CHANGELOG-X.Y.md`
+   - Open a PR to `main` and get it reviewed and merged
```
andreyvelich: If we create a new patch release, the PR should be created against release branch instead of `main`, right ?

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+**Verification**: Confirm the release branch was created/updated!
+
+#### 3. Automated Release Process
+
+The `Release` GitHub Action automatically:
+
+- Runs tests and builds the package
+- Creates and pushes the release tag
+- Publishes to PyPI (requires manual approval)
+- Creates GitHub Release (requires manual approval)
```
andreyvelich: Please update it.

docs/release/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Updates existing `release-X.Y` branch
+- Cherry-pick fixes from main to release branch
+- Example: `make release VERSION=v0.2.1`
+
+#### Release Candidate (vX.Y.Z-rc.N)
+
+- Creates pre-release
+- GitHub Release marked as "pre-release"
+- Example: `make release VERSION=v0.2.0-rc.1`
```
andreyvelich: Maybe we can remove this section, since it duplicates some info from here: https://github.com/kubeflow/sdk/blob/482ccb9c34bd5b214065d4738a289bc8d64dec91/docs/release/README.md#release-branches-and-tags

.github/workflows/release.yml
```diff
@@ -0,0 +1,210 @@
...
+            git fetch origin "$BRANCH":"$BRANCH"
+            git checkout "$BRANCH"
+            if git merge-base --is-ancestor "$MAIN_SHA" "$BRANCH"; then
+              echo "Commit $MAIN_SHA already present in $BRANCH. Skipping cherry-pick."
+            else
+              if ! git cherry-pick -x "$MAIN_SHA"; then
+                echo "Cherry-pick failed. Please resolve manually on $BRANCH." >&2
+                exit 1
+              fi
+            fi
```
andreyvelich, line 74: Would that work if we want to create a new patch release for previous releases ?
E.g. if the `main` branch has the latest `0.3.0` release, but we want to make `0.1.4` release at `release-0.1` branch.

I imagine a use-case when we want to make a new patch release in release branch, and we should not update the SDK version in the main branch.

Also, I can see that in Kueue they manually update the changelog in the main branch after the new patch release: https://github.com/kubernetes-sigs/kueue/pull/6810

.github/workflows/release.yml
```diff
@@ -0,0 +1,210 @@
...
+          git checkout "$BRANCH"
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Setup build environment
+        run: |
+          make verify
```
andreyvelich, line 98: Is that fine that we install SDK in the dev mode before running `uv build` and `uv twine check` ?
https://github.com/kubeflow/sdk/blob/482ccb9c34bd5b214065d4738a289bc8d64dec91/Makefile#L54

.github/workflows/release.yml
```diff
@@ -0,0 +1,210 @@
...
+
+      - name: Validate package
+        run: |
+          uvx twine check dist/*
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: dist-${{ needs.prepare.outputs.version }}
+          path: dist/
```
andreyvelich, line 123: Why do want to upload artifacts here ?
Maybe we can directly publish to pypi as part of this step ?

docs/release/README.md
kramaranya: Sure, I've thinking about this too

docs/release/README.md
kramaranya: Nope, removed it

docs/release/README.md
kramaranya: Do you mean we also use `X.Y.Z` for tags and branches?

docs/release/README.md
kramaranya: Why? I followed semver similar to Trainer https://github.com/kubeflow/trainer/releases/tag/v2.0.0-rc.1

docs/release/README.md
andreyvelich: Yeah, I feel like adding `v` to the SDK tags is an overkill.
Since the final release artifact of Kubeflow SDK is a Python package, we can just follow the Python conventions:

`X.Y.Z` to tag the official release
`X.Y.ZrcN` to tag the pre-release


In that case, you don't need to remove the `v` once you publish the release.

We can still maintain `release-X.Y` branches for the `kubeflow/sdk`.

For example, KFP also does this: https://github.com/kubeflow/pipelines/tree/sdk-2.14.3

@kubeflow/kubeflow-sdk-team Any thoughts ?

docs/release/README.md
kramaranya: Not really, with the current CI the idea is that we always update the version and changelog on main, and CI will cherry-pick it when needed. I kept CI to be triggered on version changes on main even for patch releases to minimize potential conflicts. What do you think about this approach?

docs/release/README.md
andreyvelich: The only problem that I see is that `X.Y.ZrcN` is not semver compatible release.
It must be `1.0.0-rc.1`, so I am not sure what we should do in that case.

kramaranya: I'm leaning towards following Python conventions, for RCs as well. [langgraph](https://github.com/langchain-ai/langgraph/releases/tag/1.0.0a3) and [llama-stack](https://github.com/llamastack/llama-stack/releases/tag/v0.2.21rc3) do the same

docs/release/README.md
andreyvelich: @kramaranya I guess, that might be challenging due to issue I mentioned here: https://github.com/kubeflow/sdk/pull/65#discussion_r2347019274

.github/workflows/release.yml
kramaranya: I kept it in separate steps to allow manual approval before publishing it to PyPI

andreyvelich: @kramaranya Can we just move this part to the `publish-pypi` step, so you. don't need to push and pull artifact?
```
      - name: Build and validate package
        run: |
          uv build
          uvx twine check dist/*
 
```

.github/workflows/release.yml
kramaranya: IIUC we never update the version on the main branch for patch releases, is that right? So whenever we do a patch release, we update the version and changelog on the release branch and then cherry-pick the changelog to main? If so, I'll update the CI accordingly

andreyvelich: I think the main branch should always correspond to the latest official release (including the most recent patch to the latest release).
The only exception is when we need to publish a patch release for a previous minor version (e.g., X.Y-1). In that case, we should update the `__version__` in the corresponding `release-X.Y-1` branch.

Does it make sense @kramaranya ?

.github/workflows/release.yml
kramaranya: Yeah, but we will need to set up python and checkout the release branch twice. Also if we push artifacts you can manually install them from github UI if needed - https://github.com/kramaranya/sdk/actions/runs/17457296258
Do we really want to avoid pushing and pulling artifacts?

.github/workflows/release.yml
kramaranya: Ah I see what you mean, thank you for the explanation! Let me update the CI for this case

.github/workflows/release.yml
andreyvelich: Sure, make sense, in that case maybe we can upload the artifact to the GitHub release, so users can do something like:
```
pip install https://github.com/kubeflow/sdk/releases/download/<tag>/<artifact>.tar.gz
or
pip install https://github.com/kubeflow/sdk/releases/download/<tag>/<artifact>.whl
```

kramaranya: Yep, sounds good to me, I'll update GH release notes for that

kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/8f56800434ca7d6d8738fcf07ff77abd4177238e

.github/workflows/release.yml
kramaranya: I addressed this in https://github.com/kubeflow/sdk/pull/65/commits/73021944904f3072837d7c6382908dfab88a503e

docs/release/README.md
kramaranya: I updated this to use `X.Y.Z` versions in https://github.com/kubeflow/sdk/pull/65/commits/366af93c5c8704c4b6186253fc3ce8edfa3993d9
Let me know if anyone has objections cc @kubeflow/kubeflow-sdk-team

docs/release/README.md
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/366af93c5c8704c4b6186253fc3ce8edfa3993d9

docs/release/README.md
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/8199dd193089022b67caafb0cfaac0fb1e81d226

docs/release/README.md
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/d6d5974ebd34ec18c18b1b46041b523895c73980

.github/workflows/release.yml
kramaranya: That shouldn't cause any issues since `uv build` ignores dev dependencies anyway. We're already using `test-python` with `uv sync`. If needed we could duplicate the makefile commands with `uv sync --no-dev`, but since we can't introduce breaking changes to the trainer API models, we should be fine just using `uv sync`, WDYT?

Makefile
kramaranya: I think so, see https://en.wikipedia.org/wiki/Sed#In-place_editing and https://ss64.com/mac/sed.html

.github/workflows/release.yml
andreyvelich: If `uv build` ignores dev dependencies that should be fine.

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+
+          if git ls-remote --heads origin "$BRANCH" | grep -q "$BRANCH"; then
+            echo "Using existing branch: $BRANCH"
+            git fetch origin "$BRANCH":"$BRANCH"
+            git checkout "$BRANCH"
+            if git merge-base --is-ancestor "$MAIN_SHA" "$BRANCH"; then
+              echo "Commit $MAIN_SHA already present in $BRANCH. Skipping cherry-pick."
+            else
+              if ! git cherry-pick -x "$MAIN_SHA"; then
+                echo "Cherry-pick failed. Please resolve manually on $BRANCH." >&2
```
andreyvelich, line 71: Do we have a use-case when we should manually resolve cherry-pick ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Checkout release branch
+        run: |
+          BRANCH="${{ needs.prepare.outputs.branch }}"
+          git fetch origin "$BRANCH":"$BRANCH" || true
+          git checkout "$BRANCH"
```
andreyvelich: Can we just do this ?
```suggestion
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
          ref: ${{ needs.prepare.outputs.branch }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+            echo "Version mismatch"; exit 1; fi
+          echo "Version verified: $TAG_VERSION"
+
+      - name: Build package
+        run: |
+          uv build
+
+      - name: Validate package
+        run: |
+          uvx twine check dist/*
```
andreyvelich: Maybe we can combine it ?
```suggestion
      - name: Build and Validate package
        run: |
          uv build
          uvx twine check dist/* 
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+          path: dist/
+
+  create-tag:
+    name: Create and push tag
+    needs: [prepare, build]
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
andreyvelich, line 132: Same here, can you checkout to the release branch here ?
```suggestion
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
          ref: ${{ needs.prepare.outputs.branch }}
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+      - name: Create tag
+        run: |
+          VERSION="${{ needs.prepare.outputs.version }}"
+          BRANCH="${{ needs.prepare.outputs.branch }}"
+          if git ls-remote --tags origin "$VERSION" | grep -q "refs/tags/$VERSION"; then
+            echo "Tag $VERSION already exists. Skipping"; exit 0; fi
```
andreyvelich, line 138: Did you check the existence of tag here because publish-pypi or github-release job might fail ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+          git fetch origin "$BRANCH":"$BRANCH"
+          git checkout "$BRANCH"
+          git tag "$VERSION"
```
andreyvelich, line 139: Do we want to create signed tag, I noticed that k8s does this ?
```
git tag -s $VERSION
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+          git push origin "$VERSION"
+
+  publish-pypi:
+    name: Publish to PyPI
+    needs: [prepare, build, create-tag]
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+      url: https://pypi.org/project/kubeflow/
```
andreyvelich, line 148: What about secrets ? Do we need to generate Token to push to `kubeflow` PyPI ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+    name: Create GitHub Release
+    needs: [prepare, build, create-tag, publish-pypi]
+    runs-on: ubuntu-latest
+    environment:
+      name: release
+      url: https://github.com/kubeflow/sdk/releases
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
andreyvelich, line 170: do you need to checkout to the release branch ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+                echo "changelog<<EOF"
+                echo "$CHANGELOG"
+                echo "EOF"
+              } >> $GITHUB_OUTPUT
+            else
+              echo "changelog=**Note:** No changelog section found for $VERSION in $CHANGELOG_FILE" >> $GITHUB_OUTPUT
+            fi
+          else
+            echo "changelog=**Note:** $CHANGELOG_FILE not found" >> $GITHUB_OUTPUT
+          fi
```
andreyvelich: I think, you can simplify this script.
We will ensure that changelog exists before publishing the release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,225 @@
...
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v1
+        with:
+          tag_name: ${{ needs.prepare.outputs.version }}
+          name: Kubeflow SDK ${{ needs.prepare.outputs.version }}
```
andreyvelich: Do we want to be consistent with naming of releases for [KF Trainer](https://github.com/kubeflow/trainer/releases/tag/v2.0.0) and [Llama Stack](https://github.com/llamastack/llama-stack/releases/tag/v0.2.21), and remove the project name from the release ?
```suggestion
          name: ${{ needs.prepare.outputs.version }}
```

.github/workflows/release.yml
kramaranya: Generally no, but if that happens it shouldn't be a problem cause we can always rerun the workflow

.github/workflows/release.yml
kramaranya: Good idea!

.github/workflows/release.yml
kramaranya: Sure!

.github/workflows/release.yml
kramaranya: I check if the tag exists first because sometimes we might have already published tag but the pypi publish failed. In those cases we want to be able to rerun the workflow to publish to pypi without trying to create the tag again, since pushing a duplicate tag would cause an error and block the workflow with pypi and github release steps

.github/workflows/release.yml
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/22d8cc877460b9076fb12c340d2f4bec7a7da9cc

.github/workflows/release.yml
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/22d8cc877460b9076fb12c340d2f4bec7a7da9cc

.github/workflows/release.yml
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/65/commits/22d8cc877460b9076fb12c340d2f4bec7a7da9cc

.github/workflows/release.yml
kramaranya: I guess for that we will need add GPG key to repo secrets, do we have this already?

.github/workflows/release.yml
kramaranya: Nope, we'll use PyPI trusted publishing https://docs.pypi.org/trusted-publishers/adding-a-publisher/

.github/workflows/release.yml
andreyvelich: Sounds good.

.github/workflows/release.yml
andreyvelich: No, we haven't done it yet.
We can do that in the future.

.github/workflows/release.yml
andreyvelich: I see, I guess I should update it since only I have access to the registry ?

.github/workflows/release.yml
kramaranya: Yes, for patch releases! Updated in https://github.com/kubeflow/sdk/pull/65/commits/98b1a07a5275ef50ee150d9340f21e8feb610816

.github/workflows/release.yml
kramaranya: Yep, I was planning to explain you everything we need to configure before publishing this :)
We will need to add a trusted publishing to PyPI and create a github env

.github/workflows/release.yml
kramaranya: I've tried to simplify it in https://github.com/kubeflow/sdk/pull/65/commits/04e4b9b1329ef5972328ec27fedea93b4bfb7fc9

.github/workflows/release.yml
kramaranya: Sure, updated in https://github.com/kubeflow/sdk/pull/65/commits/7a98c0cb809bae33284da67e0d1235cd779c393a

andreyvelich (commented):
Thanks for this effort @kramaranya!
/lgtm
/assign @szaher @astefanutti @Electronic-Waste

andreyvelich:
>I see! Did you need to manually remove the links for every release?

Yes, we have this `perl` command to replace it: https://github.com/kubeflow/trainer/tree/c6997fcc36a7a771071a3f127fe192225a2f48b9/docs/release#create-github-release

andreyvelich (commented):
Thanks @kramaranya!
I think, we can move forward with this.
/lgtm
/assign @szaher @astefanutti @Electronic-Waste

andreyvelich:
/approve
/hold
@kramaranya Feel free to un-hold this.

PR #66, opened 2025-08-22
chore: Add KEP to the supported PR titles

szaher:

SDK PRs needs to support a new type for proposals

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/check-pr-title.yaml
+++ b/.github/workflows/check-pr-title.yaml
@@ -28,7 +28,7 @@ jobs:
             fix
             feat
             revert
-            proposal
+            KEP
 
           scopes: |
             ci


.github/workflows/check-pr-title.yaml
```diff
@@ -28,11 +28,13 @@ jobs:
             fix
             feat
             revert
+            KEP
```
kramaranya, line 31: Do we really need it? I think we should use `feat: KEP-NN`

.github/workflows/check-pr-title.yaml
```diff
@@ -28,11 +28,13 @@ jobs:
...
 
           scopes: |
             ci
             docs
             examples
+            proposals
```
kramaranya, line 37: We can just use `feat(docs): KEP-NN` for that

szaher: proposal are not really docs... wDYT?

.github/workflows/check-pr-title.yaml
szaher: I would say yes, KEP is different from features. features are new features that we introduce to the sdk but KEP is a proposal or enhancement proposal.

.github/workflows/check-pr-title.yaml
andreyvelich: I kind of agree with @kramaranya, we can use `(docs)` scope for KEPs. Otherwise, we can just to `feat: KEP-nn`, since scope is not required.
I would like us to keep `feat`, `fix` in the title name since it is easier to group them in CHANGELOG later.

szaher: if we use `feat` it will be confusing to users or whoever reads the changelog or release notes. Users would assume this feature is already implemented as the release notes mentions it as one of the PRs.

If we add `KEP` it clearly signals this is a TODO or planned feature... 

feat: Local Execution >>> contains KEP
feat: Local Execution  >>> Actual implementation

cc: @andreyvelich @kramaranya

andreyvelich: Yeah, but we can group them in the dedicated subsection in the CHANGELOG, like we did here:
https://github.com/kubeflow/trainer/blob/master/CHANGELOG.md#llm-trainer-v2

andreyvelich: FYI, KFP also uses `docs` prefix for KEPs: https://github.com/kubeflow/pipelines/pull/11551

szaher: If we add the `KEP` type I think it will be consistent with the following. 
image
which I believe is cleaner.
> With KEP type
KEP 2: Local execution 
feat: Local execution

> Without KEP type (Current Behaviour) 
docs: KEP-2 Local exection
feat: Local execution 

Which one do you think is cleaner?

I believe we can group both or leave both in the release notes in the respective sections

andreyvelich: I suggest that we use `feat:` and `fix:` prefix, and we add the KEP number after.
E.g.:
```
feat: KEP-2 - Add local backend for KF Trainer
fix: KEP-2 - Update the runtime return type
chore: KEP-2 Add unit tests for utils
```

Since that will be simpler to group fixes, chores, and features dedicated to the particular KEP.

kramaranya: I agree with @andreyvelich, that would be much cleaner. Also, if we add a KEP prefix, it could confuse contributors, and they might end up using both options: 
```
feat: KEP-NN - ...
KEP-NN: ...
```
So I think it’s better to stick with one clear format to keep it consistent

PR #34, opened 2025-06-21
KEP-2: Local Execution Mode Proposal

szaher:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

PR #33, opened 2025-06-21
[WIP] Implement TrainerClient Backends & Local Process

szaher:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

PR #68, opened 2025-08-22
feat: Implement Kubernetes Backend

szaher:

Fixes #

**Checklist:**

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -12,518 +12,122 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 import logging
-import multiprocessing
-import queue
-import random
-import string
-import time
-import uuid
-from typing import Dict, List, Optional, Union, Set
+from typing import Dict, Optional, Union, Set, List
 
+from kubeflow.trainer.backends.kubernetes.k8s import KubernetesBackend
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
-from kubeflow.trainer.utils import utils
-from kubeflow_trainer_api import models
-from kubernetes import client, config, watch
+from kubeflow.trainer.backends.kubernetes.types import KubernetesBackendConfig
+
 
 logger = logging.getLogger(__name__)
 
 
 class TrainerClient:
-    def __init__(
-        self,
-        config_file: Optional[str] = None,
-        context: Optional[str] = None,
-        client_configuration: Optional[client.Configuration] = None,
-        namespace: Optional[str] = None,
+
+    def __init__( self,
+            backend_config: KubernetesBackendConfig = KubernetesBackendConfig(),
     ):
-        """TrainerClient constructor. Configure logging in your application
-            as follows to see detailed information from the TrainerClient APIs:
-            .. code-block:: python
-                import logging
-                logging.basicConfig()
-                log = logging.getLogger("kubeflow.trainer.api.trainer_client")
-                log.setLevel(logging.DEBUG)
+        """Initialize a Kubeflow Trainer client.
 
         Args:
-            config_file: Path to the kube-config file. Defaults to ~/.kube/config.
-            context: Set the active context. Defaults to current_context from the kube-config.
-            client_configuration: Client configuration for cluster authentication.
-                You have to provide valid configuration with Bearer token or
-                with username and password. You can find an example here:
-                https://github.com/kubernetes-client/python/blob/67f9c7a97081b4526470cad53576bc3b71fa6fcc/examples/remote_cluster.py#L31
-            namespace: Target Kubernetes namespace. If SDK runs outside of Kubernetes cluster it
-                takes the namespace from the kube-config context. If SDK runs inside
-                the Kubernetes cluster it takes namespace from the
-                `/var/run/secrets/kubernetes.io/serviceaccount/namespace` file. By default it
-                uses the `default` namespace.
+            backend_config: Backend configuration. Either KubernetesBackendConfig or
+                            LocalProcessBackendConfig, or None to use the backend's
+                            default config class. Defaults to KubernetesBackendConfig.
         """
+        # initialize training backend
+        if isinstance(backend_config, KubernetesBackendConfig):
+            self.__backend = KubernetesBackend(backend_config)
+        else:
+            raise ValueError("Invalid backend config '{}'".format(backend_config))
 
-        if namespace is None:
-            namespace = utils.get_default_target_namespace(context)
 
-        # If client configuration is not set, use kube-config to access Kubernetes APIs.
-        if client_configuration is None:
-            # Load kube-config or in-cluster config.
-            if config_file or not utils.is_running_in_k8s():
-                config.load_kube_config(config_file=config_file, context=context)
-            else:
-                config.load_incluster_config()
-
-        k8s_client = client.ApiClient(client_configuration)
-        self.custom_api = client.CustomObjectsApi(k8s_client)
-        self.core_api = client.CoreV1Api(k8s_client)
-
-        self.namespace = namespace
-
-    def list_runtimes(self) -> List[types.Runtime]:
+    def list_runtimes(self) -> types.Runtime:
         """List of the available Runtimes.
 
-        Returns:
-            List[Runtime]: List of available training runtimes.
-                If no runtimes exist, an empty list is returned.
+            Returns:
+                List[Runtime]: List of available training runtimes.
+                    If no runtimes exist, an empty list is returned.
 
-        Raises:
-            TimeoutError: Timeout to list Runtimes.
-            RuntimeError: Failed to list Runtimes.
+            Raises:
+                TimeoutError: Timeout to list Runtimes.
+                RuntimeError: Failed to list Runtimes.
         """
-
-        result = []
-        try:
-            thread = self.custom_api.list_cluster_custom_object(
-                constants.GROUP,
-                constants.VERSION,
-                constants.CLUSTER_TRAINING_RUNTIME_PLURAL,
-                async_req=True,
-            )
-
-            runtime_list = models.TrainerV1alpha1ClusterTrainingRuntimeList.from_dict(
-                thread.get(constants.DEFAULT_TIMEOUT)
-            )
-
-            if not runtime_list:
-                return result
-
-            for runtime in runtime_list.items:
-                if not (
-                    runtime.metadata
-                    and runtime.metadata.labels
-                    and constants.RUNTIME_FRAMEWORK_LABEL in runtime.metadata.labels
-                ):
-                    logger.warning(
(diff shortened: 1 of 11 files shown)


kubeflow/trainer/api/trainer_client.py
```diff
@@ -12,518 +12,122 @@
...
-    def list_runtimes(self) -> List[types.Runtime]:
+    def list_runtimes(self) -> types.Runtime:
         """List of the available Runtimes.
 
-        Returns:
-            List[Runtime]: List of available training runtimes.
-                If no runtimes exist, an empty list is returned.
+            Returns:
+                List[Runtime]: List of available training runtimes.
+                    If no runtimes exist, an empty list is returned.
```
andreyvelich: @szaher Can you go over my comments in this PR which are applicable for Kubernetes backend, and address them here: https://github.com/kubeflow/sdk/pull/33#pullrequestreview-3145167873 ?

andreyvelich: For example: https://github.com/kubeflow/sdk/pull/33#discussion_r2294221116

szaher: I think it's all address apart from spaces. I'll create another PR later to add ruff configuration

kubeflow/trainer/backends/base.py
```diff
@@ -0,0 +1,65 @@
...
+# limitations under the License.
+
+import abc
+
+from typing import Dict, List, Optional, Set, Union
+from kubeflow.trainer.constants import constants
+from kubeflow.trainer.types import types
+
+
+class TrainingBackend(abc.ABC):
```
andreyvelich: As I mentioned here, shall we name it RunnerBackend ?
https://github.com/kubeflow/sdk/pull/34#discussion_r2294406231

kubeflow/trainer/backends/kubernetes/k8s.py
```diff
@@ -0,0 +1,703 @@
+# Copyright 2024 The Kubeflow Authors.
```
andreyvelich: name this file backend.py.

kubeflow/trainer/backends/kubernetes/types.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class KubernetesBackendConfig(types.BackendConfig):
+    namespace: Optional[str] = None
+    config_file: Optional[str] = None
+    context: Optional[str] = None
+    client_configuration: Optional[client.Configuration] = None
+
+    class Config:
+        arbitrary_types_allowed = True
```
andreyvelich, line 27: Ref: https://github.com/kubeflow/sdk/pull/33#discussion_r2294226284

kubeflow/trainer/types/types.py
```diff
@@ -238,3 +237,8 @@ class Initializer:
 
     dataset: Optional[HuggingFaceDatasetInitializer] = None
     model: Optional[HuggingFaceModelInitializer] = None
+
+
+# Training Backends Types
+class BackendConfig(BaseModel):
+    pass
```
andreyvelich: Ref: https://github.com/kubeflow/sdk/pull/33#discussion_r2294227246

kubeflow/trainer/api/trainer_client.py
```diff
@@ -12,73 +12,37 @@
 # See the License for the specific language governing permissions and
```
andreyvelich, line 12: Please keep the order of APIs like in Kube backend.

kubeflow/trainer/backends/kubernetes/k8s.py
```diff
@@ -0,0 +1,703 @@
...
+                Configuration for Custom Training Task or Config-driven Task with Builtin Trainer.
+
+        Returns:
+            str: The unique name of the TrainJob that has been generated.
+
+        Raises:
+            ValueError: Input arguments are invalid.
+            TimeoutError: Timeout to create TrainJobs.
+            RuntimeError: Failed to create TrainJobs.
+        """
```
andreyvelich: We don't need to duplicate the docstring for backend.py, since we can easily make it out of date.
Let's keep it only for `trainer_client.py` since it is user-facing.

kubeflow/trainer/backends/kubernetes/types.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+from kubernetes import client
+
+from kubeflow.trainer.types import types
+
+
+class KubernetesBackendConfig(types.BackendConfig):
```
andreyvelich: Ref: https://github.com/kubeflow/sdk/pull/33#discussion_r2294230059

andreyvelich (commented):
Thanks @szaher!

kubeflow/trainer/api/trainer_client.py
szaher: I have added a pr for ruff config that should fix this but I prefer to merge it after this one #69

kubeflow/trainer/backends/kubernetes/types.py
szaher: This will allow easier serialization when we added trainer UI will remove it for now np

kubeflow/trainer/backends/base.py
szaher: I believe `TrainingBackend` is more suitable but sure np

kubeflow/trainer/backends/kubernetes/k8s.py
szaher: make sense but I would love to avoid stutter but Ok np

kubeflow/trainer/api/trainer_client.py
szaher: This should be fixed automatically when we merge the ruff PR #69

kubeflow/trainer/backends/kubernetes/types.py
szaher: pydantic models are much richer and will help us in the future if we decided to add UI. also it provides validation, de/serialization ...etc.

If it's a must be a dataclass then I propose to use the follow dataclass 

```
from pydantic.dataclasses import dataclass
```
more info https://docs.pydantic.dev/latest/concepts/dataclasses/

kubeflow/trainer/backends/kubernetes/k8s.py
szaher: I missed this one, will delete it.

kubeflow/trainer/types/types.py
szaher: same as https://github.com/kubeflow/sdk/pull/68#discussion_r2294926196

kubeflow/trainer/api/trainer_client.py
andreyvelich: How ruff will fix the order of APIs ? I guess, you should manually update it in the file.

kubeflow/trainer/types/types.py
andreyvelich: In this comment I meant that configs can directly inherit the BaseModel.
There is no need for BackedConfig class.

kubeflow/trainer/api/trainer_client.py
szaher: sorry I thought it was about the imports. the API order is already updated in this PR

kubeflow/trainer/types/types.py
szaher: Sure

kubeflow/trainer/api/trainer_client.py
andreyvelich: I can see that they don't have same order, isn't ?
Screenshot 2025-08-25 at 11 29 06
Screenshot 2025-08-25 at 11 29 12

kubeflow/trainer/backends/kubernetes/types.py
andreyvelich: @szaher Did you push commit to delete it ?

kubeflow/trainer/backends/base.py
```diff
@@ -0,0 +1,65 @@
...
+# limitations under the License.
+
+import abc
+
+from typing import Dict, List, Optional, Set, Union
+from kubeflow.trainer.constants import constants
+from kubeflow.trainer.types import types
+
+
+class RunnerBackend(abc.ABC):
```
andreyvelich: @szaher What are thoughts between RunnerBackend and ExecutionBackend as we discussed here: https://github.com/kubeflow/sdk/pull/34#discussion_r2294934486 ?

kubeflow/trainer/api/trainer_client.py
```diff
@@ -89,441 +53,82 @@ def list_runtimes(self) -> List[types.Runtime]:
...
             the entire model training process, e.g. `CustomTrainer`.
         - Config-driven Task with Existing Trainer: Training with a trainer that already includes
             the post-training logic, requiring only parameter adjustments, e.g. `BuiltinTrainer`.
-
         Args:
-            runtime: Reference to one of existing Runtimes. By default the
-                torch-distributed Runtime is used.
-            initializer:
+            runtime (`types.Runtime`): Reference to one of existing Runtimes.
+            initializer (`Optional[types.Initializer]`):
```
andreyvelich: I noticed that we don't need to set type of Args in the docstring, since we already add them to the function args.
Like here, we don't say that `name` must be string: https://github.com/kubeflow/sdk/blob/2dc5daddd4e03d89db4cb4fd82f9e806ef3e3d4f/kubeflow/trainer/api/trainer_client.py#L58-L61
```suggestion
            runtime: Reference to one of existing Runtimes.
            initializer:
```

kubeflow/trainer/backends/base.py
szaher: anyways, I moved it already to `RunnerBackend` for now

kubeflow/trainer/backends/kubernetes/types.py
szaher: This is required since `kubernetes.client.Configuration` isn't a core/native Pydantic type so we need this to allow pydantic to use it.

kubeflow/trainer/api/trainer_client.py
szaher: this happened as `get_runtime_packages` was added after I started working on this PR with sometime so as a result of the merge with main it ended up at the end of the file. sorting it now ...

kubeflow/trainer/backends/base.py
szaher: I believe `ExecutionBackend` could be more suitable here.

szaher: renaming ...

kubeflow/trainer/api/trainer_client.py
szaher: I think `ruff format` might do that for us automatically (will check in #69 )

andreyvelich (commented):
Thank you for this great contribution @szaher!
/lgtm
/approve

PR #41, opened 2025-07-17
Update CONTRIBUTING.md to use uv

szaher:
**What this PR does / why we need it**:

* Update CONTRIBUTING.md to use uv
* Add pre-commit as dev dependency 
* add pytest as dev dependency

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

Issue #36, opened 2025-07-03
Build and Release Kubeflow to pypi


szaher:
### What you would like to be added?

Add automated gate jobs into the CI pipeline to validate releases of the Kubeflow SDK

- [ ] All gate jobs complete successfully on main branch and tag.

Release Type | Installation Command
-- | --
Latest Stable | pip install kubeflow-sdk
Specific Version | pip install kubeflow-sdk==MAJOR.MINOR.PATCH
Pre-release (RC) | pip install --pre kubeflow-sdk
Main Branch | pip install  kubeflow-sdk:main or kubeflow-sdk:commit

release would require all CI jobs to be successful (lint, test, ...etc)

### Why is this needed?

To Enhance UX of kubeflow SDK and encourage users to install it and use with ease, we need to release the SDK to pip to increase user adoption.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

franciscojavierarceo:
yeah we do a similar thing in Feast and it works well, although sometimes it can be chaotic if actions change.

PR #71, opened 2025-08-25
chore: Move dev extras to dependency-groups

### LINKED ISSUE #70: PyPI Publishing Blocked by Git Dependencies
I've been implementing an automated release process for Kubeflow SDK and when testing it I found one blocking issue -- PyPI publishing fails with this error:

```
Can't have direct dependency: kubeflow-trainer-api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api
```

Our `pyproject.toml` contains a git dependency in the `dev` optional dependencies, that is not allowed when publishing to PyPi:
```
[project.optional-dependencies]
dev = [
  "pytest>=7.0",
  "pytest-mock>=3.10",
  "coverage>=7.0",
  "kubeflow_trainer_api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api",
  "ruff>=0.12.2",
  "pre-commit>=4.2.0",
]
```

**What could we do instead?**
1) Remove `[project.optional-dependencies].dev` entirely and move all dev dependencies to dependency groups:

```toml
[dependency-groups]
dev = [
  "pytest>=7.0",
  "pytest-mock>=3.10", 
  "coverage>=7.0",
  "kubeflow-trainer-api @ git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api",
  "ruff>=0.12.2",
  "pre-commit>=4.2.0",
]
```

This means we don't publish dev extra to PyPi

2) Keep the dev extra but use only published PyPI versions:

```toml
[project.optional-dependencies]
dev = [
  "pytest>=7.0",
  "pytest-mock>=3.10",
  "coverage>=7.0", 
  "kubeflow-trainer-api>=2.0.0",  # Published PyPI version only
  "ruff>=0.12.2",
  "pre-commit>=4.2.0",
]
```

This means our dev extra only tests against stable released versions of `kubeflow-trainer-api`, not the latest master changes and users can't get Kubeflow SDK with latest changes of trainer API models

3) We can keep dev extra with published versions for PyPI compatibility and add dependency groups for latest changes in `kubeflow-trainer-api`, that will look smth like that:

```toml
[project.optional-dependencies]
dev = [
  "pytest>=7.0",
  "pytest-mock>=3.10",
  "coverage>=7.0",
  "kubeflow-trainer-api>=2.0.0",
  "ruff>=0.12.2", 
  "pre-commit>=4.2.0",
]

[dependency-groups]
dev-latest = [
  {include-group = "dev"},
  "kubeflow-trainer-api @ git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api",
]
```

This means we have PyPI-compatible dev dependencies and access to latest trainer API changes through dependency groups. However users still can't get latest changes from PyPI

I'd go with option 1, to just use `[dependency-groups]`, since we can test against master trainer API chnages and users who need the latest trainer API changes can install directly from git:
```
pip install git+https://github.com/kubeflow/sdk.git@main --dependency-groups=dev
```

Are there any other approaches we can consider? WDYT @andreyvelich @Electronic-Waste @astefanutti

kramaranya:
PyPi doesn't allow to use git dependency in extras when publishing to PyPI. I've moved dev extras to dependency-groups

Fixes #70 

**Checklist:**

/assign @andreyvelich @astefanutti @Electronic-Waste

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,13 @@ source .venv/bin/activate
 
 Install dependencies in editable mode:
 ```sh
-uv sync
+pip install -e .
+```
+
+#### Development Build (Optional)
+To install development tools and the latest API modules directly from the master branch:
+```sh
+uv pip install -e . --group dev
 ```
 
 ## Development Workflow


szaher:
Thanks Anya!

I think if we follow the proposal here https://github.com/kubeflow/sdk/issues/36 to publish `kubeflow-trainer-api`, we won't have problems publishing kubeflow sdk to pypi.org as we can easily replace the git+https with a python wheel from pypi.org with `master` tag instead of specific version which will always give us the latest code.

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ pip install -e .
 #### Development Build (Optional)
 To install development tools and the latest API modules directly from the master branch:
 ```sh
-pip install -e ".[dev]"
+uv pip install -e . --group dev
```
szaher, line 33: if `dev` is a dependency group we probably better use the following 

```suggestion
uv sync --group dev
```

Makefile
```diff
@@ -67,7 +67,7 @@ uv-venv:
  # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv
-	@uv pip install -e ".[dev]"
+	@uv pip install -e . --group dev
```
szaher: same here 

```suggestion
	@uv sync --group dev
```

CONTRIBUTING.md
kramaranya: I left as it is, since it'll be updated as part of https://github.com/kubeflow/sdk/pull/41
But I'm happy to do update it!

kramaranya: In that case it should be just `uv sync` since `dev` dependency group is included by default in uv sync -- https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups

kramaranya: Updated

Makefile
kramaranya: Updated

kramaranya:
> I think if we follow the proposal here https://github.com/kubeflow/sdk/issues/36 to publish kubeflow-trainer-api, we won't have problems publishing kubeflow sdk to pypi.org as we can easily replace the git+https with a python wheel from pypi.org with master tag instead of specific version which will always give us the latest code.

We don't currently publish `kubeflow-trainer-api` with a master tag to PyPI. If we decide to do something like that, it would be part of https://github.com/kubeflow/trainer/issues/2155. For now, we can move this to dependency-groups for dev deps

That said, have you seen other projects publishing packages with a master tag? I don't think PyPI supports mutable master or main tags.

Instead, what we could do for the Kubeflow SDK is publish a `kubeflow-nightly` version.
Any thoughts on that @szaher @andreyvelich?

CONTRIBUTING.md
```diff
@@ -24,13 +24,7 @@ source .venv/bin/activate
...
 Install dependencies in editable mode:
 ```sh
-pip install -e .
-```
-
-#### Development Build (Optional)
-To install development tools and the latest API modules directly from the master branch:
-```sh
-pip install -e ".[dev]"
+uv sync
```
szaher: Let's avoid partial updates since this will be addressed in #41 with the rest of uv commands

szaher (changes requested):
you're right, pypi.org doesn't support mutable versions like main/master/stable.

so I think for now our best option is to use dependency-groups let uv install it by default.

CONTRIBUTING.md
kramaranya: Updated :)

astefanutti:
We may have to adjust the approach when there'll be multiple release branches maintained but we are not there yet.

/lgtm

Thanks @kramaranya!

PR #74, opened 2025-08-28
feat: Support multiple pip index URLs in CustomTrainer

### LINKED ISSUE #72: Support Multiple Pip Index URLs

Trainer SDK only supports a single `pip_index_url` parameter in `CustomTrainer` - https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/utils/utils.py#L256-L260

We should extend `CustomTrainer` to accept `pip_index_urls` as a list of URLs, where the first URL becomes the primary --index-url and subsequent URLs become --extra-index-url parameters in the generated pip install command. It should still default to PyPI when no URLs are specified. 

### Why is this needed?

We would like to let users install packages from multiple sources simultaneously.
KFP already does this: https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/dsl/component_factory.py#L102-L107

Is related to https://github.com/kubeflow/katib/issues/2565

### Love this feature?

wassimbensalem:
## What this PR does

Implements support for multiple pip index URLs in CustomTrainer as requested in #72.

### Changes Made

- **Types**: Changed `pip_index_url` to `pip_index_urls: list[str]` in CustomTrainer
- **Constants**: Added `DEFAULT_PIP_INDEX_URLS` with environment variable support
- **Utils**: Implemented logic for `--index-url` and `--extra-index-url` in pip install commands
- **Tests**: Updated test cases to use new field names
- **Backward Compatibility**: Maintained `DEFAULT_PIP_INDEX_URL` for existing code

### Usage Examples

```python
# Single URL (backward compatible)
trainer = CustomTrainer(
    func=training_function,
    pip_index_urls=["https://pypi.org/simple"]
)

# Multiple URLs
trainer = CustomTrainer(
    func=training_function,
    pip_index_urls=[
        "https://pypi.org/simple",
        "https://private.repo.com/simple"
    ]
)
```

### Environment Variable Support

```bash
export DEFAULT_PIP_INDEX_URLS="https://pypi.org/simple,https://private.repo.com/simple"
```

### Technical Details

- Uses `field(default_factory=...)` to prevent mutable default issues
- First URL becomes `--index-url`, subsequent URLs become `--extra-index-url`
- Automatically includes PyPI if not in the provided URLs
- Follows the same pattern as KFP for consistency

Closes #72

### DIFF
--- a/kubeflow/trainer/constants/constants.py
+++ b/kubeflow/trainer/constants/constants.py
@@ -128,10 +128,8 @@ POD_LABEL_SELECTOR = ("{}={{trainjob_name}},{} in ({}, {}, {}, {})").format(
 # The default PIP index URL to download Python packages.
 DEFAULT_PYPI_URL = "https://pypi.org/simple"
 # Handle environment variable for multiple URLs (comma-separated)
-DEFAULT_PIP_INDEX_URLS = (
-    os.getenv("DEFAULT_PIP_INDEX_URLS", DEFAULT_PYPI_URL).split(",") 
-    if os.getenv("DEFAULT_PIP_INDEX_URLS") 
-    else [DEFAULT_PYPI_URL]
+DEFAULT_PIP_INDEX_URLS = os.getenv("DEFAULT_PIP_INDEX_URLS", DEFAULT_PYPI_URL).split(
+    ","
 )
 # Keep backward compatibility
 DEFAULT_PIP_INDEX_URL = DEFAULT_PYPI_URL
--- a/kubeflow/trainer/types/types.py
+++ b/kubeflow/trainer/types/types.py
@@ -32,7 +32,9 @@ class CustomTrainer:
         func_args (`Optional[Dict]`): The arguments to pass to the function.
         packages_to_install (`Optional[List[str]]`):
             A list of Python packages to install before running the function.
-        pip_index_urls (`Optional[list[str]]`): The PyPI URLs from which to install Python packages. The first URL will be the index-url, and remaining ones are extra-index-urls.
+        pip_index_urls (`Optional[list[str]]`): The PyPI URLs from which to install
+            Python packages. The first URL will be the index-url, and remaining ones
+            are extra-index-urls.
         num_nodes (`Optional[int]`): The number of nodes to use for training.
         resources_per_node (`Optional[Dict]`): The computing resources to allocate per node.
         env (`Optional[Dict[str, str]]`): The environment variables to set in the training nodes.
@@ -41,7 +43,9 @@ class CustomTrainer:
     func: Callable
     func_args: Optional[Dict] = None
     packages_to_install: Optional[list[str]] = None
-    pip_index_urls: list[str] = field(default_factory=lambda: constants.DEFAULT_PIP_INDEX_URLS)
+    pip_index_urls: list[str] = field(
+        default_factory=lambda: constants.DEFAULT_PIP_INDEX_URLS
+    )
     num_nodes: Optional[int] = None
     resources_per_node: Optional[Dict] = None
     env: Optional[Dict[str, str]] = None
@@ -152,7 +156,9 @@ class BuiltinTrainer:
 
 
 # Change it to list: BUILTIN_CONFIGS, once we support more Builtin Trainer configs.
-TORCH_TUNE = BuiltinTrainer.__annotations__["config"].__name__.lower().replace("config", "")
+TORCH_TUNE = (
+    BuiltinTrainer.__annotations__["config"].__name__.lower().replace("config", "")
+)
 
 
 class TrainerType(Enum):
--- a/kubeflow/trainer/utils/utils.py
+++ b/kubeflow/trainer/utils/utils.py
@@ -20,10 +20,11 @@ import threading
 from typing import Any, Callable, Dict, Optional
 from urllib.parse import urlparse
 
-from kubeflow.trainer.constants import constants
-from kubeflow.trainer.types import types
 from kubeflow_trainer_api import models
 from kubernetes import config
+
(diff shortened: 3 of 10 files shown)


kubeflow/trainer/types/types.py
```diff
@@ -32,7 +32,7 @@ class CustomTrainer:
         func_args (`Optional[Dict]`): The arguments to pass to the function.
         packages_to_install (`Optional[List[str]]`):
             A list of Python packages to install before running the function.
-        pip_index_url (`Optional[str]`): The PyPI URL from which to install Python packages.
+        pip_index_urls (`Optional[list[str]]`): The PyPI URLs from which to install Python packages.
```
andreyvelich: Please add the comment that the first URL will be used as the default index, and the rest will be used as additional indexes.

kubeflow/trainer/types/types.py
```diff
@@ -41,7 +41,7 @@ class CustomTrainer:
     func: Callable
     func_args: Optional[Dict] = None
     packages_to_install: Optional[list[str]] = None
-    pip_index_url: str = constants.DEFAULT_PIP_INDEX_URL
+    pip_index_urls: list[str] = field(default_factory=lambda: constants.DEFAULT_PIP_INDEX_URLS)
```
andreyvelich: Why do you want to use `field` here ?

kubeflow/trainer/utils/utils.py
```diff
@@ -255,28 +255,37 @@ def get_resources_per_node(
 
 def get_script_for_python_packages(
     packages_to_install: list[str],
-    pip_index_url: str,
-    is_mpi: bool,
+    pip_index_urls: list[str] = constants.DEFAULT_PIP_INDEX_URLS,
+    is_mpi: bool = False,
+    include_pypi: bool = True
```
andreyvelich: Why do you want to introduce this flag ?
I would let user explicitly sets the PyPI index in the list if they need it:
```py
pip_index_urls=["pypi.custom.com/simple", "https://pypi.org/simple"]
```

kubeflow/trainer/utils/utils.py
```diff
@@ -255,28 +255,37 @@ def get_resources_per_node(
...
         if ! [ -x "$(command -v pip)" ]; then
             python -m ensurepip || python -m ensurepip --user || apt-get install python-pip
         fi
 
         PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \
-        --no-warn-script-location --index-url {} {} {}
+        --no-warn-script-location {} {}
         """.format(
-            pip_index_url,
+            options_str,
```
andreyvelich: You can directly set it here.
```suggestion
           " ".join(options),
```

kubeflow/trainer/types/types.py
wassimbensalem: If we use simply the constants.DEFAULT_PIP_INDEX_URLS, in any instance that we create if someone modifies constants.DEFAULT_PIP_INDEX_ULRLS it will affect all created instance. Sothe field with default factory will assure that we have a fresh copy of it each time. This is optional, I can remove it as well !

kubeflow/trainer/utils/utils.py
wassimbensalem: I Just was thinking if the user forgot to set the default pypi. For instance he chose pip_index_urls = ["repoA", "repoB"], but the package wasnt there, so we have a fallback solution which is the default pypi. This is also optional and I can remove it :)

astefanutti:
@wassimbensalem could you please sign the commits?

kubeflow/trainer/types/types.py
andreyvelich: I see, we probably need to do this to return new copy all the time object is created ?
```py
pip_index_urls: list[str] = field(default_factory=lambda: list(constants.DEFAULT_PIP_INDEX_URLS))
```

kubeflow/trainer/utils/utils.py
andreyvelich: We can have it in the future, if we get any complains from users.
I would just define single constant that can be overridden by env:
```py
DEFAULT_PIP_INDEX_URLS = os.getenv(
    "DEFAULT_PIP_INDEX_URLS", "https://pypi.org/simple"
).split(",")

```

kubeflow/trainer/constants/constants.py
```diff
@@ -126,7 +126,15 @@
...
 
 # The default PIP index URL to download Python packages.
-DEFAULT_PIP_INDEX_URL = os.getenv("DEFAULT_PIP_INDEX_URL", "https://pypi.org/simple")
+DEFAULT_PYPI_URL = "https://pypi.org/simple"
+# Handle environment variable for multiple URLs (comma-separated)
+DEFAULT_PIP_INDEX_URLS = (
+    os.getenv("DEFAULT_PIP_INDEX_URLS", DEFAULT_PYPI_URL).split(",") 
+    if os.getenv("DEFAULT_PIP_INDEX_URLS") 
+    else [DEFAULT_PYPI_URL]
+)
```
kramaranya: We could simplify this to not call `getenv` twice

kubeflow/trainer/utils/utils.py
kramaranya: Agree with @andreyvelich

kramaranya (commented):
Thank you @wassimbensalem!
Could you also sign your commits and fix the code formatting -- https://github.com/kubeflow/sdk/blob/main/CONTRIBUTING.md#development-workflow?

kubeflow/trainer/types/types.py
wassimbensalem: Exactly

PR #78, opened 2025-08-30
fix: trainer client backend public

### LINKED ISSUE #77: Expose backend in TrainerClient for direct access"

Expose the backend attribute in TrainerClient as public (instead of private) so that users can directly access backend functionality.

### Why is this needed?

Currently, backend in TrainerClient is private, which prevents users from accessing backend functionality directly. This change makes it public, allowing advanced users to interact with the backend when needed.

Related: https://github.com/kubeflow/trainer/pull/2762

### Love this feature?

jaiakash:
**What this PR does / why we need it:**
This PR makes the backend attribute in TrainerClient publicly accessible. Currently, it is private, which restricts users from programmatically interacting with the underlying backend. Exposing it allows developers to extend functionality, debug, or customize backend usage more easily.

Fixes #77

**Checklist:**

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -38,7 +38,7 @@ class TrainerClient:
         """
         # initialize training backend
         if isinstance(backend_config, KubernetesBackendConfig):
-            self.__backend = KubernetesBackend(backend_config)
+            self.backend = KubernetesBackend(backend_config)
         else:
             raise ValueError("Invalid backend config '{}'".format(backend_config))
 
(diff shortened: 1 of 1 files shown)


kramaranya (commented):
I agree we should keep it public. Thank you @jaiakash!
/lgtm

kramaranya:
We probably shall add [alpaca notebook](https://github.com/kubeflow/trainer/blob/master/examples/torchtune/llama3_2/alpaca-trainjob-yaml.ipynb) to the e2e, so we cover this functionality as well, wdyt? @andreyvelich @jaiakash

andreyvelich:
> We probably shall add [alpaca notebook](https://github.com/kubeflow/trainer/blob/master/examples/torchtune/llama3_2/alpaca-trainjob-yaml.ipynb) to the e2e, so we cover this functionality as well, wdyt? @andreyvelich @jaiakash

We are working on it as part of, which enables testing on GPUs: https://github.com/kubeflow/trainer/pull/2762

andreyvelich (commented):
/approve

jaiakash:
Thanks @andreyvelich @kramaranya 
We are working on it as part of, which enables testing on GPUs: https://github.com/kubeflow/trainer/pull/2762

PR #80, opened 2025-08-30
chore(trainer): Use explicit exception chaining

andreyvelich:
I updated exception handling to use explicit exception chaining via `raise ... from e`.
This ensures that the original exception is preserved and clearly visible in tracebacks.

```py
try:
    TrainerClient().get_job(name="test")
except Exception as e:
    print(e.__cause__)
```
Without this change it prints `None`, but it should print:
```
(404)
Reason: Not Found
...
```

Also, I updated a few docstrings

/assign @astefanutti @kramaranya @szaher @Electronic-Waste

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -25,7 +25,7 @@ import random
 import string
 import uuid
 from dataclasses import asdict, dataclass, field
-from typing import Any, Dict, Optional, Type
+from typing import Any, Optional, Type
 from unittest.mock import Mock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


kubeflow/trainer/api/trainer_client.py
```diff
@@ -129,28 +151,29 @@ def get_job_logs(
...
         timeout: int = 600,
         polling_interval: int = 2,
     ) -> types.TrainJob:
-        """Wait for TrainJob to reach the desired status
+        """Wait for a TrainJob to reach a desired status.
 
         Args:
             name: Name of the TrainJob.
-            status: Set of expected statuses. It must be subset of Created, Running, Complete, and
+            status: Expected statues. Must be a subset of Created, Running, Complete, and
```
kramaranya:
```suggestion
            status: Expected statuses. Must be a subset of Created, Running, Complete, and
```

kramaranya (commented):
Thank you @andreyvelich!
Would you mind updating [types.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/types/types.py#L19), [utils.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/utils/utils.py#L20) and [backend_test.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend_test.py#L28) to use built-in dict instead of Dict?

PR #81, opened 2025-09-01
chore: Enable parallel builds for coveralls

kramaranya:
I noticed we get this error sometimes when running unit test CI job:
```
Error: Unprocessable Entity
Response: {"message":"Can't add a job to a build that is already closed. Build 17225808879 is closed. See docs.coveralls.io/parallel-builds","error":true}
```
Example from https://github.com/kubeflow/sdk/actions/runs/17225808879/job/48870011364
So I enabled parallel mode to upload coverage data for both Python versions simultaneously and keep the Coveralls build open
https://docs.coveralls.io/parallel-builds

/assign @andreyvelich @astefanutti @Electronic-Waste

### DIFF
--- a/.github/workflows/test-python.yaml
+++ b/.github/workflows/test-python.yaml
@@ -38,3 +38,16 @@ jobs:
         uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          parallel: true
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
Thank you @kramaranya!
/lgtm
/approve

PR #82, opened 2025-09-02
chore: Add welcome new contributors CI

kramaranya:
I added a CI to write a comment to the PRs from new contributors. I'm using https://github.com/actions/first-interaction
An example of this GH action is https://github.com/google/blockly/pull/9250#pullrequestreview-3045474517

I borrowed this idea from Model Registry, thanks @tarilabs! 

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -1,0 +1,28 @@
+name: Welcome new contributors
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+jobs:
+  welcome:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/first-interaction@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          pr-message: |
+            🎉 **Welcome to the Kubeflow SDK!** 🎉
+
+            Thanks for opening your first PR! We're happy to have you as part of our community 🚀
+
+            **Here's what happens next:**
+
+            - If you haven't already, please check out our [Contributing Guide](https://github.com/kubeflow/sdk/blob/main/CONTRIBUTING.md)
+            - Our team will review your PR soon! cc @kubeflow/kubeflow-sdk-team
+            - Feel free to ask questions in the comments if you need any help or clarification
+
+            Thanks again for contributing to Kubeflow! 🙏
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -25,7 +25,7 @@ jobs:
             - Our team will review your PR soon! cc @kubeflow/kubeflow-sdk-team
 
             **Join the community:**
-            - **Slack**: Join our [#kubeflow-ml-experience](https://www.kubeflow.org/docs/about/community/#kubeflow-slack-channels) Slack channel
+            - **Slack**: Join our [#kubeflow-ml-experience](https://www.kubeflow.org/docs/about/community/#kubeflow-slack-channels) and [#kubeflow-trainer](https://www.kubeflow.org/docs/about/community/#kubeflow-slack-channels) Slack channels
             - **Meetings**: Attend the [Kubeflow SDK and ML Experience](https://bit.ly/kf-ml-experience) bi-weekly meetings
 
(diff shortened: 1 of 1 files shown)


.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/first-interaction@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          pr-message: |
+            🎉 **Welcome to the Kubeflow SDK!** 🎉
+
+            Thanks for opening your first PR! We're happy to have you as part of our community 🚀
+
+            **Here's what happens next:**
+
```
andreyvelich, line 23: Maybe it worths to also link the Kubeflow contributor guide, since it explains that commits must be singed: https://www.kubeflow.org/docs/about/contributing/

.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,28 @@
...
+            - If you haven't already, please check out our [Contributing Guide](https://github.com/kubeflow/sdk/blob/main/CONTRIBUTING.md)
```
andreyvelich: Shall we also encourage them to join Kubeflow Slack channels (e.g. `#kubeflow-trainer` and `#kubeflow-ml-experience`) and community calls ?

.github/workflows/welcome-new-contributors.yaml
```diff
@@ -0,0 +1,28 @@
...
+    types:
+      - opened
+
+jobs:
+  welcome:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    steps:
+      - uses: actions/first-interaction@v3
```
andreyvelich, line 14: Do you know how it detects the first interaction ? Does it check how many PRs user opens ?

andreyvelich (commented):
That's great idea, thank you for this @kramaranya!
We should also add this for Kubeflow Trainer.
cc @kubeflow/kubeflow-trainer-team @kubeflow/kubeflow-sdk-team @kubeflow/wg-training-leads

.github/workflows/welcome-new-contributors.yaml
kramaranya: Good point, let me add this

.github/workflows/welcome-new-contributors.yaml
kramaranya: Sure, I'll update it!
However I think we should just refer to `#kubeflow-ml-experience`, wdyt?

.github/workflows/welcome-new-contributors.yaml
kramaranya: I don't know for sure, I think it just checks whether the PR author has ever opened a PR here before https://github.com/actions/first-interaction/tree/main?tab=readme-ov-file#first-interaction
and just hope it'll work... :)

.github/workflows/welcome-new-contributors.yaml
andreyvelich: From my point of view, it would be beneficial for contributors to join both of these channels, since Kubeflow SDK will affect many Kubeflow projects.

kramaranya: Yeah, but in this case will we need to also add Slack channels for model registry, pipelines, katib etc?

andreyvelich: Let's add them once we support it in Kubeflow SDK ?

kramaranya: I meant do we actually want to explicitly point to each project's slack channel? If we list multiple channels, it might confuse new contributors about where to start or which channel to use for what.

I'll add trainer to the list, but in the future (same in README) maybe we could point to `#kubeflow-ml-experience` as the primary channel and add a note that contributors should also join other project specific channels (like trainer, pipelines) if they're working on features that touch those components?

andreyvelich: Sure, we can do that.
>If they're working on features that touch those components?

In any case, contributors will work on features that touch some Kubeflow sub-projects, so they should be aware of the development.

andreyvelich (commented):
Thank you @kramaranya!
If you can create similar action in KF Trainer, that would be great!
/lgtm

Electronic-Waste (commented):
@kramaranya Thanks for this great contribution. LGTM!

/lgtm

Issue #75, opened 2025-08-29
Refactor `get_job_logs()` API for Kubeflow Trainer


andreyvelich:
Currently, `get_job_logs()` returns [dict where key is the step name and value is the step's logs](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L132).
I don't think that it is necessary since users already have to provide[ `step` and `node_rank` in the function arg.](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L130C9-L131)

I propose that we change the API as follows:

```py
 def get_job_logs(
        self,
        name: str,
        step: str = constants.NODE + "-0",
        follow: Optional[bool] = False,
    ) -> str:
```

The downside of this approach is that we don't allow to get logs from the multiple nodes at the same time.
e.g.
```
[node 0] This is data from the pod 0
[node 1] This is data from the pod 1
```
IMO, this might not be needed since users more interested to get logs for the particular rank (e.g. GPU), like [what DeepSpeed is printing](https://github.com/kubeflow/trainer/blob/master/examples/deepspeed/text-summarization/T5-Fine-Tuning.ipynb). 

Thoughts @kramaranya @astefanutti @szaher @Electronic-Waste @briangallagher @rudeigerc ?

kramaranya:
That makes sense to me! One thing I've been thinking about is returning an Iterator instead:
```python
 def get_job_logs(
        self,
        name: str,
        step: str = constants.NODE + "-0",
        follow: Optional[bool] = False,
    ) -> Iterator[str]:
```

This way users can process and filter logs quite easily:
```python
for line in TrainerClient().get_job_logs(name=job_id, follow=True):
    if "loss" in line:
        print(line, end="")
```

That is what Ray does for `tail_job_logs` : https://github.com/ray-project/ray/blob/89fef22c869a732bd22582a9876616beb5f34a71/python/ray/dashboard/modules/job/sdk.py#L463-L482

I think this is more efficient way to stream logs, especially for `follow=True` cases, rather than returning big string and waiting for it to be built in memory. And it should be more intuitive way for data scietists to look for specific patterns and extract metrics from logs.

Since we already use `watch.Watch().stream`, refactoring this should not be a problem https://github.com/kubeflow/sdk/blob/f342acd65d3910435dde72ca4b92c56848422e7c/kubeflow/trainer/backends/kubernetes/backend.py#L351-L358

WDYT?

PR #83, opened 2025-09-03
feat(trainer): Refactor get_job_logs() API with Iterator

andreyvelich:
Fixes: https://github.com/kubeflow/sdk/issues/75

I updated the `get_job_logs()` to always return iterator from TrainJob's step logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Join Slack](https://img.shields.io/badge/Join_Slack-blue?logo=slack)](https://www.kubeflow.org/docs/about/community/#kubeflow-slack-channels)
 [![Coverage Status](https://coveralls.io/repos/github/kubeflow/sdk/badge.svg?branch=main)](https://coveralls.io/github/kubeflow/sdk?branch=main)
 [![Ask DeepWiki](https://deepwiki.com/badge.svg)](https://deepwiki.com/kubeflow/sdk)
+
 <!-- TODO(kramaranya): update when release [![Python Supported Versions](https://img.shields.io/pypi/pyversions/kubeflow.svg?color=%2334D058)](https://pypi.org/project/kubeflow/) -->
 
 ## Overview
@@ -36,6 +37,7 @@ ML applications rather than managing complex infrastrutcure.
 ```bash
 pip install git+https://github.com/kubeflow/sdk.git@main
 ```
+
 <!-- TODO(kramaranya): update before release pip install -U kubeflow -->
 
 ### Run your first PyTorch distributed job
@@ -49,7 +51,7 @@ def get_torch_dist():
     import torch.distributed as dist
 
     dist.init_process_group(backend="gloo")
-    print(f"PyTorch Distributed Environment")
+    print("PyTorch Distributed Environment")
     print(f"WORLD_SIZE: {dist.get_world_size()}")
     print(f"RANK: {dist.get_rank()}")
     print(f"LOCAL_RANK: {os.environ['LOCAL_RANK']}")
@@ -70,17 +72,17 @@ job_id = TrainerClient().train(
 TrainerClient().wait_for_job_status(job_id)
 
 # Print TrainJob logs
-print(TrainerClient().get_job_logs(name=job_id, node_rank=0)["node-0"])
+print("\n".join(TrainerClient().get_job_logs(name=job_id)))
 ```
 
 ## Supported Kubeflow Projects
 
-| Project                     | Status | Description                                                |
-|-----------------------------|--------|------------------------------------------------------------|
+| Project                     | Status           | Description                                                |
+| --------------------------- | ---------------- | ---------------------------------------------------------- |
 | **Kubeflow Trainer**        | ✅ **Available** | Train and fine-tune AI models with various frameworks      |
-| **Kubeflow Katib**          | 🚧 Planned | Hyperparameter optimization                                |
-| **Kubeflow Pipelines**      | 🚧 Planned | Build, run, and track AI workflows                         |
-| **Kubeflow Model Registry** | 🚧 Planned | Manage model artifacts, versions and ML artifacts metadata |
+| **Kubeflow Katib**          | 🚧 Planned       | Hyperparameter optimization                                |
+| **Kubeflow Pipelines**      | 🚧 Planned       | Build, run, and track AI workflows                         |
+| **Kubeflow Model Registry** | 🚧 Planned       | Manage model artifacts, versions and ML artifacts metadata |
 
 ## Community
 
@@ -98,6 +100,7 @@ Kubeflow SDK is a community project and is still under active development. We we
 ## Documentation
 
 <!-- TODO(kramaranya): add kubeflow sdk docs -->
+
 - **[Design Document](https://docs.google.com/document/d/1rX7ELAHRb_lvh0Y7BK1HBYAbA0zi9enB0F_358ZC58w/edit)**: Kubeflow SDK design proposal
 - **[Component Guides](https://www.kubeflow.org/docs/components/)**: Individual component documentation
 - **[DeepWiki](https://deepwiki.com/kubeflow/sdk)**: AI-powered repository documentation
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -163,7 +163,7 @@ class TrainerClient:
         Args:
             name: Name of the TrainJob.
             step: Step of the TrainJob to collect logs from, like dataset-initializer or node-0.
-            follow: Whether to stream logs in realtime as they produced.
+            follow: Whether to stream logs in realtime as they are produced.
 
         Returns:
             Iterator of log lines.
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import logging
-from typing import Optional, Union
+from typing import Optional, Union, Iterator
 
 from kubeflow.trainer.constants import constants
 from kubeflow.trainer.types import types
@@ -120,8 +120,7 @@ class TrainerClient:
             runtime: Reference to one of the existing runtimes.
 
         Returns:
-            List: List of created TrainJobs.
-                If no TrainJob exist, an empty list is returned.
+            List of created TrainJobs. If no TrainJob exist, an empty list is returned.
 
         Raises:
             TimeoutError: Timeout to list TrainJobs.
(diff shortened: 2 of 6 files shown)


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -328,92 +328,48 @@ def get_job_logs(
...
         name: str,
         follow: Optional[bool] = False,
-        step: str = constants.NODE,
-        node_rank: int = 0,
-    ) -> dict[str, str]:
-        """Get the logs from TrainJob"""
-
+        step: str = constants.NODE + "-0",
+    ) -> Iterator[str]:
+        """Get the TrainJob logs"""
```
kramaranya, line 333: Do we need this docstring here?

kubeflow/trainer/api/trainer_client.py
```diff
@@ -148,12 +147,33 @@ def get_job(self, name: str) -> types.TrainJob:
...
+        from kubeflow.trainer import TrainerClient
+
+        for logline in TrainerClient().get_job_logs(name="s8d44aa4fb6d", follow=True):
+            print(logline)
+        ```
+
+        Args:
+            name: Name of the TrainJob.
+            step: Step of the TrainJob to collect logs from, like dataset-initializer or node-0.
+            follow: Whether to stream logs in realtime as they produced.
```
kramaranya:
```suggestion
            follow: Whether to stream logs in realtime as they are produced.
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -173,7 +173,7 @@ def print_packages():
         )
 
         self.wait_for_job_status(job_name)
-        print(self.get_job_logs(job_name)["node-0"])
+        print(self.get_job_logs(job_name))
```
kramaranya: This will print a generator object, not logs, right?
I think we should instead do:
```python
for line in self.get_job_logs(job_name):
    print(line, end="")
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -328,92 +328,48 @@ def get_job_logs(
...
-            # Create thread and queue per stream, for non-blocking iteration.
-            log_queue_pool = utils.get_log_queue_pool(log_streams)
-
-            # Iterate over every watching pods' log queue
-            while True:
-                for index, log_queue in enumerate(log_queue_pool):
-                    if all(finished):
+                # Stream logs incrementally
+                for logline in log_stream:
+                    if logline is None:
```
kramaranya: Can this actually yield None?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -328,92 +328,48 @@ def get_job_logs(
...
+            return iter([])
+
+        try:
+            if follow:
+                log_stream = watch.Watch().stream(
                     self.core_api.read_namespaced_pod_log,
                     name=pod_name,
                     namespace=self.namespace,
-                    container=constants.NODE,
+                    container=re.sub(r"-\d+$", "", step),  # Remove the number for the node step.
```
kramaranya: nit: we call this twice, cold just count once and reuse it

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -328,92 +328,48 @@ def get_job_logs(
...
-        logs_dict = {}
-
-        # TODO (andreyvelich): Potentially, refactor this.
-        # Support logging of multiple Pods.
-        # TODO (andreyvelich): Currently, follow is supported only for node container.
-        if follow and step == constants.NODE:
-            log_streams = []
-            log_streams.append(
-                watch.Watch().stream(
+            return iter([])
```
kramaranya: Shall we raise RuntimeError or log a warning in this case?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -933,10 +931,12 @@ def test_get_job_logs(trainer_client, test_case):
     """Test TrainerClient.get_job_logs with basic success path."""
     print("Executing test:", test_case.name)
     try:
+        trainer_client.namespace = test_case.config.get("namespace", DEFAULT_NAMESPACE)
```
kramaranya: This should be from backend, right?
```suggestion
        trainer_client.backend.namespace = test_case.config.get("namespace", DEFAULT_NAMESPACE)
```

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: We also need to update other references, for example [README](https://github.com/andreyvelich/sdk/blob/947f74c50daf7e8809f734cbf8086364c056cd92/README.md?plain=1#L73)

astefanutti: Would `print(*self.get_job_logs(job_name), sep="\n")` work?

Otherwise, would there be a way to override the string representation of the returned iterator?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -328,92 +328,48 @@ def get_job_logs(
...
 
-            # Create thread and queue per stream, for non-blocking iteration.
-            log_queue_pool = utils.get_log_queue_pool(log_streams)
-
-            # Iterate over every watching pods' log queue
-            while True:
-                for index, log_queue in enumerate(log_queue_pool):
-                    if all(finished):
+                # Stream logs incrementally
+                for logline in log_stream:
```
astefanutti, line 356: Are we sure each item are entire lines?

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: Hmm it could, but I think we might run out of memory :) And there'll be no streaming cause we have to wait for iterator to finish, which works against the purpose of iterator

Why don't just use this? 
```python
for line in self.get_job_logs(job_name):
    print(line, end="")
```

astefanutti: I agree, and prefer the streaming approach, it was really to get a one liner :)

kramaranya: Got you! I think users can still do smth like `"".join(client.get_job_logs(job_name))` if they prefer just one string

kramaranya: Alternatively we could have two APIs -- one for streaming and another for returning a complete string, similar to what Ray does:
`tail_job_logs` for streaming that uses an Iterator https://github.com/ray-project/ray/blob/master/python/ray/dashboard/modules/job/sdk.py#L463-L482
`get_job_logs` that returns all logs as a single string https://github.com/ray-project/ray/blob/master/python/ray/dashboard/modules/job/sdk.py#L434-L449

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: Not necessary, but this is just reminder how this API is used for developers and AI tools 🙂

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: >Hmm it could, but I think we might run out of memory :)

Why we run out of memory ? Since we just print the `pip list` + `nvidia-smi` the log would be small.

Users can do something like this if they don't want to define loop:
```py
print("\n".join(TrainerClient().get_job_logs(name=job_id)))
```

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: I think, we can just do this:
```py
  if logline is None:
      return
```

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: Not sure if we should raise an Exception here, since at this stage TrainJob is not yet produced logs, so we just return empty logs.
@astefanutti thoughts ?

kubeflow/trainer/backends/kubernetes/backend_test.py
andreyvelich: Actually, Trainer Client here is the Kubernetes backend, not TrainerClient() https://github.com/kubeflow/sdk/blob/5ce8e7bf0836722629fec5a4d916648faf20f8d2/kubeflow/trainer/backends/kubernetes/backend_test.py#L108
Let me rename it.

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: Yes, log_stream is `<generator object Watch.stream at 0x1073c6340>` object, and we can return it line by line.

andreyvelich:
@astefanutti @kramaranya I will open PRs to update Trainer examples + website today.
Since this is breaking change to the existing examples and E2Es are failing, are we ok to merge this PR manually ?

kubeflow/trainer/backends/kubernetes/backend.py
astefanutti: I've checked and it indeed does yield items line by line: ttps://github.com/kubernetes-client/python/blob/6e7c539f52dec4e993d2c32a4408920d8522f47e/kubernetes/base/watch/watch.py#L54-L83

I wasn't sure whether we had to do it ourselves or not.

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: I see, sounds good to me!

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: > Why we run out of memory ? Since we just print the pip list + nvidia-smi the log would be small.

Yeah, I meant when logs are large

kramaranya: > Alternatively we could have two APIs -- one for streaming and another for returning a complete string, similar to what Ray does

@andreyvelich @astefanutti any thoughts on that?

andreyvelich: I would suggest that we consolidate it in the single `get_job_logs()` API, similar to how we consolidate BuiltinTrainer and CustomTrainer into `train()` API.
I don't see much value to separate them, since it is better to return Iterator[str] for both.

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: Hmm, IIUC it never yields None https://github.com/kubernetes-client/python/blob/master/kubernetes/base/watch/watch.py#L213-L216

We could just do
```python
for logline in log_stream:
    yield logline
```

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: I see, makes sense to me to return an empty iterator then

kubeflow/trainer/backends/kubernetes/backend_test.py
kramaranya: Ahh I see, thank you!

kubeflow/trainer/backends/kubernetes/backend.py
kramaranya: SGTM

kramaranya:
> @astefanutti @kramaranya I will open PRs to update Trainer examples + website today. Since this is breaking change to the existing examples and E2Es are failing, are we ok to merge this PR manually ?

Sure, makes sense

andreyvelich (commented):
/approve

andreyvelich:
Looks like E2Es are succeed after the change.
We don't need to manually merge this PR 🎉

README.md
```diff
@@ -70,17 +72,17 @@ job_id = TrainerClient().train(
 TrainerClient().wait_for_job_status(job_id)
 
 # Print TrainJob logs
-print(TrainerClient().get_job_logs(name=job_id, node_rank=0)["node-0"])
+print("\n".join(TrainerClient().get_job_logs(name=job_id)))
```
kramaranya, line 75: Does it worth showing `follow=True` here?

andreyvelich: By default `wait_for_job_status()` waits until TrainJob is complete, so showing example with follow is unnecessary here.

kramaranya: Oh yeah, missed this part

Issue #84, opened 2025-09-05
Improvements for Kubeflow Trainer Integration in Kubeflow Pipelines


mprahl:
Hello,
I'm a Kubeflow Pipelines maintainer and I'm working on making it easier to launch training jobs from a pipeline component/step as a side project.

The high-level goals of the integration:

* Allow setting resource requests (CPU, GPU, memory), environment variables, node selectors, PVCs, etc. on the component but have it apply to the TrainJob instead (done in https://github.com/kubeflow/pipelines/pull/12185)
* Leverage KFP's concepts of artifacts from outputs from the training job (share a PVC between the component and the train job)
* Pass component input parameters from the pipeline step to the training job
* Abstract creating the train job with the user's input training function, stream the logs of the train job, and poll until the train job completes

Potential enhancements to the Kubeflow SDK:

- [ ] Enhance `CustomTrainer` with more options that are available in the CRD API (e.g. `podSpecOverrides`)
- [ ] Enhance `CustomTrainer` with an optional argument of `train_job_patch: dict` that deep‑merges into the TrainJob spec before submission. This enables advanced users to tweak any CRD field safely without forking the SDK.
- [ ] Allow setting a name for the `TrainJob`
- [ ] Enhance `wait_for_job_status` with `stream_logs=True` keyword argument to stream logs of the main train job pod in the background. See `stream_main_worker_logs` in the code snippet below.
- [ ] A high-level single call similar to the `submit_train_job` in the code snippet below.
- [ ] Encode training kwargs as JSON in an env var (e.g., `TRAINING_CONFIG`) and decode inside the ephemeral script before calling the function. This is more likely to succeed than literally injecting the dictionary.
- [ ] Set a default `HOME` environment variable to a writeable location to ensure `pip installs` work when the container isn't running as root.
- [ ] Ensure pip installs in the ephemeral script always use `--user` to work when the container isn't running as root.
- [ ] Enhance `CustomTrainer` to accept the new KFP type `dsl.TaskConfig` (https://github.com/kubeflow/pipelines/pull/12185) so users can leverage the Kubeflow SDK directly in the pipeline.

This is the user experienced I got working in a proof of concept:

```python
from kfp import dsl
from typing import Optional
import kfp
import kfp.kubernetes

@dsl.component(
    # Required to create a Kubeflow Trainer job
    packages_to_install=["kubernetes"],
    # This is a new feature that hasn't been released yet.
[...]

Issue #92, opened 2025-09-12
Training Options for TrainJob customization


kramaranya:
### What you would like to be added?

We’d like to propose an “options” style API for the Kubeflow Trainer SDK that lets users customize a TrainJob without surfacing the full Kubernetes surface and writing YAML files.
The idea is to add an optional options parameter to `train()` inspired by golang style package [options](https://github.com/uber-go/guide/blob/master/style.md#functional-options) where each option is a small callable that mutates the TrainJob before submit.
This gives data scientists a simple way to do common tasks like adding labels and annotations, enabling Kueue, attaching per‑Pod template overrides:
```python

from kubeflow.trainer import TrainerClient, CustomTrainer
from kubeflow.trainer.options import (
    WithLabels,
    WithAnnotations,
    WithKueue,
    WithPodSpecOverrides,
)

def MyCustomOption(service_account: str):
    def apply(job: dict):
        spec = job.setdefault("spec", {})
        spec.setdefault("podSpecOverrides", []).append({"serviceAccountName": service_account})
    return apply

job_id = TrainerClient().train(
    runtime=runtime,
    trainer=CustomTrainer(
        func=train,
        packages_to_install=["transformers", "boto3"],
    ),
    options=[
        WithLabels({"project": "training", "owner": "platform-team"}),
        WithAnnotations({"version": "v1"}),
        WithKueue(
            queue="ml-queue",
            topology_labels={"kueue.x-k8s.io/podset-name": "node"},
            managed_by="kueue.x-k8s.io/multikueue",
        ),
        WithPodSpecOverrides([
            {
                "targetJobs": ["node"],
                "labels": {"workload": "training"},
                "containers": [
                    {"name": "node", "volumeMounts": [{"name": "data", "mountPath": "/workspace/data"}]}
                ],
                "volumes": [
                    {"name": "data", "persistentVolumeClaim": {"claimName": "datasets-pvc"}}
                ],
            }
        ]),
        MyCustomOption("trainer-sa"),
    ],
)
``` 

### Why is this needed?

This addresses the current gaps we have such as setting [labels/annotations from the SDK](https://github.com/kubeflow/sdk/issues/87), using Kueue, and mounting [volumes/volumeMounts](https://github.com/kubeflow/sdk/issues/90).
The API would stay the same by default; if a user doesn’t pass options, behavior is unchanged.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

PR #97, opened 2025-09-15
feat: add optional name parameter to training client's train() method…

abhijeet-dhumal:
- The Train backend API was intentionally designed to only support auto-generated names. 
- Custom names for trainjob improves job identification and management
- Meaningful names are useful for tracking and debugging

Partially fixes https://github.com/kubeflow/sdk/issues/84
- Added the name parameter to train method to allow specifying name for a trainjob
- Added proper Kubernetes name validation
- Supporting both custom names AND auto-generation
- Added unit tests for the same

**Checklist:**

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -93,6 +93,7 @@ class TrainerClient:
         runtime: Optional[types.Runtime] = None,
         initializer: Optional[types.Initializer] = None,
         trainer: Optional[Union[types.CustomTrainer, types.BuiltinTrainer]] = None,
+        name: Optional[str] = None,
     ) -> str:
         """Create a TrainJob. You can configure the TrainJob using one of these trainers:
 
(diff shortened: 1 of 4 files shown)


kubeflow/trainer/api/trainer_client.py
```diff
@@ -93,6 +93,7 @@ def train(
         runtime: Optional[types.Runtime] = None,
         initializer: Optional[types.Initializer] = None,
         trainer: Optional[Union[types.CustomTrainer, types.BuiltinTrainer]] = None,
+        name: Optional[str] = None,
```
kramaranya, line 96: Thank you @abhijeet-dhumal!
I'm thinking whether this should be part of options instead https://github.com/kubeflow/sdk/issues/92? Are there any benefits?
```python
job_id = TrainerClient().train(
    runtime=runtime,
    trainer=CustomTrainer(
        func=train,
    ),
    options=[
        WithName("train")
    ],
)
```
WDYT? @abhijeet-dhumal @astefanutti @briangallagher @szaher @andreyvelich

abhijeet-dhumal: @kramaranya Yeah it makes sense...works for me, but will for others to confirm !

astefanutti: +1 to have it as part of the options approach.

PR #111, opened 2025-09-29
feat: Add docker backend

### LINKED ISSUE #108: Create Docker Backend to support Local Execution

Currently there is a Kubernetes backend for full remote distributed training and SubProcess backend for Local execution. 
The proposal is to add an additional backend for Docker, also to be used in a Local context. 

### Why is this needed?

This bridges the gap between the current subprocess backend and a full kubernetes backend by facilitating training in a containerised environment and across multiple containers. 

### Love this feature?

briangallagher:
This PR adds a docker backend, extending the current Kubernetes and Local Process backends. 
Its designed for Local Execution

Fixes #108 

**Verification**
This [WIP PR](https://github.com/kubeflow/trainer/pull/2869) to add an example notebook for local docker execution can be used to verify this PR
Update the pip install to: 
`!pip install "git+https://github.com/briangallagher/sdk.git@docker-backend#egg=kubeflow[docker]"`


kubeflow/trainer/backends/docker/backend.py
```diff
@@ -0,0 +1,536 @@
...
+    list_local_runtimes,
+)
+from kubeflow.trainer.backends.docker.types import LocalDockerBackendConfig
+from kubeflow.trainer.constants import constants
+from kubeflow.trainer.types import types
+
+logger = logging.getLogger(__name__)
+
+
+DOCKER_LABEL_PREFIX = "trainer.kubeflow.ai"
```
astefanutti, line 67:
```suggestion
DOCKER_LABEL_PREFIX = "trainer.kubeflow.org"
```

kubeflow/trainer/backends/docker/backend.py
```diff
@@ -0,0 +1,536 @@
...
+import logging
+import os
+from pathlib import Path
+import random
+import shutil
+import string
+import uuid
+
+try:
+    import docker  # type: ignore
```
astefanutti, line 49: How difficult would it be to support Podman as well: https://pypi.org/project/podman/

Should it be a different backend or could this be a "container" backend that could work for both Docker and Podman depending on what's installed / configured?

briangallagher: I think it's a little bit cleaner to have a separate backend as it uses the podman client and the code is a bit different. The networking is slightly different. 
But, a nice addition might be some auto discovery after all backends are added.  
Meaning, look for Podman, then docker finally fall back to subprocess.

andreyvelich: I would agree with @astefanutti to unify all container-based functionality under single backend.
From the end-user point of view it doesn't matter whether they use Docker, Podman, or Container backend to run local containers.

briangallagher:
Closing this, the code will be merged into https://github.com/kubeflow/sdk/pull/119

PR #112, opened 2025-09-30
feat(cache): KEP-2655: Support provisioning of cache with Kubeflow SDK

akshaychitneni:
Adds cache initializer in sdk

Fixes # https://github.com/kubeflow/trainer/issues/2866

**Checklist:**


kubeflow/trainer/types/types.py
```diff
@@ -220,6 +220,20 @@ class HuggingFaceDatasetInitializer:
     access_token: Optional[str] = None
 
 
+@dataclass
+class DataCacheInitializer:
+    cluster_size: int
+    metadata_loc: str
+    schema_name: str
+    table_name: str
+    env: Optional[dict[str, str]] = None
```
andreyvelich: Do we want to be explicit which optional values user can set rather than giving them arbitrary `env` configuration in DataCacheInitializer?

kubeflow/trainer/utils/utils.py
```diff
@@ -526,34 +526,50 @@ def get_args_in_dataset_preprocess_config(
...
+        Union[types.HuggingFaceDatasetInitializer, types.DataCacheInitializer]
+    ] = None,
 ) -> Optional[models.TrainerV1alpha1DatasetInitializer]:
     """
     Get the TrainJob dataset initializer from the given config.
     """
-    if not isinstance(dataset, types.HuggingFaceDatasetInitializer):
-        return None
+    if isinstance(dataset, types.HuggingFaceDatasetInitializer):
+        # TODO (andreyvelich): Support more parameters.
```
andreyvelich: you probably can remove this TODO for now.
```suggestion
```

kubeflow/trainer/utils/utils.py
```diff
@@ -526,34 +526,50 @@ def get_args_in_dataset_preprocess_config(
...
+                        name=constants.INITIALIZER_ENV_ACCESS_TOKEN,
+                        value=dataset.access_token,
+                    ),
+                ]
+                if dataset.access_token
+                else None
+            ),
+        )
+        return dataset_initializer
+    elif isinstance(dataset, types.DataCacheInitializer):
```
andreyvelich, line 593: you probably can make this more generic, since ENV names are equal to the DataClass field names:
```py
    envs = []
    for f in fields(dataset):
        name = f.name.upper()
        value = getattr(dataset, f.name)
        envs.append(models.IoK8sApiCoreV1EnvVar(name=name, value=str(value)))
```

kubeflow/trainer/types/types.py
```diff
@@ -220,6 +220,20 @@ class HuggingFaceDatasetInitializer:
     access_token: Optional[str] = None
 
 
+@dataclass
+class DataCacheInitializer:
+    cluster_size: int
+    metadata_loc: str
+    schema_name: str
+    table_name: str
```
andreyvelich: Would it be possible to leverage `storage_uri` to set location or it has some limitations ?

e.g.
```
cache://<CATALOG_NAME>/<DATABASE_NAME>/<TABLE_NAME>
```

I can imagine that metadata location can be set separately.

akshaychitneni: As discussed, lets have uri to `cache://<DATABASE_NAME>/<TABLE_NAME>` and use metadata_loc

andreyvelich:
@akshaychitneni Since we merged this PR, did you get a chance to review the rest comments: https://github.com/kubeflow/trainer/pull/2793 ?

kubeflow/trainer/types/types.py
akshaychitneni: I think it would be hard to keep the env in sync with initializer and sdk

andreyvelich: I would prefer that we keep these values in sync to reduce users confusion, since they will be always aware of parameters they can set in DataCacheInitializer config.

kubeflow/trainer/constants/constants.py
```diff
@@ -171,3 +171,6 @@
 
 # The Instruct Datasets class in torchtune
 TORCH_TUNE_INSTRUCT_DATASET = "torchtune.datasets.instruct_dataset"
+
+# Allowed environment variables for DataCacheInitializer
+DATACACHE_ALLOWED_ENV_VARS = {"head_cpu", "head_mem", "worker_cpu", "worker_mem", "iam_role"}
```
andreyvelich: Given that you want to maintain this List, removing values from it will be still consider as a breaking change.
@akshaychitneni Could we just add explicit options to DataCacheInitializer config that users can modify ?

kubeflow/trainer/types/types.py
```diff
@@ -274,11 +304,11 @@ class Initializer:
     """Initializer defines configurations for dataset and pre-trained model initialization
 
     Args:
-        dataset (`Optional[HuggingFaceDatasetInitializer]`): The configuration for one of the
-            supported dataset initializers.
+        dataset (`Optional[Union[HuggingFaceDatasetInitializer, DataCacheInitializer]]`):
+            The configuration for one of the supported dataset initializers.
```
Electronic-Waste, line 326: Can we briefly introduce what is `HuggingFaceDatasetInitializer` and `DataCacheInitializer`? I believe, it will help users have a clear understanding of when to use `DataCacheInitializer`.

kubeflow/trainer/types/types_test.py
```diff
@@ -0,0 +1,115 @@
...
+            config={
+                "storage_uri": "cache://test_schema/test_table/extra",
+                "cluster_size": 3,
+                "metadata_loc": "gs://my-bucket/metadata",
+            },
+            expected_error=ValueError,
+        ),
+    ],
+)
+def test_datacacheinitializer(test_case: TestCase):
```
Electronic-Waste:
```suggestion
def test_datacache_initializer(test_case: TestCase):
```

Will this be better? Or do you prefer the original one or `test_data_cache_initializer`?

Electronic-Waste (commented):
@akshaychitneni Thanks for this. Just a few nits

kubeflow/trainer/types/types_test.py
andreyvelich: @akshaychitneni Let's use `test_data_cache_initializer()` to follow Python best practices of converting PascalCase to snake_case for testing.

kubeflow/trainer/types/types.py
andreyvelich: Yes, @akshaychitneni please can you add the short description to those data-classes ?
https://github.com/kubeflow/sdk/blob/1c767a22bb4e1ecb5031732de984a6007a3cc41c/kubeflow/trainer/types/types.py#L266

kubeflow/trainer/types/types.py
```diff
@@ -258,10 +258,44 @@ class TrainJob:
...
 class HuggingFaceDatasetInitializer:
+    """Configuration for downloading datasets from HuggingFace Hub."""
+
     storage_uri: str
     access_token: Optional[str] = None
 
 
+@dataclass
+class DataCacheInitializer:
+    """Configuration for distributed data caching system for training workloads."""
```
andreyvelich: @akshaychitneni Please can you explain Args for this DataClass, so users aware how to use it ? You can find example for CustomTrainer:
https://github.com/kubeflow/sdk/blob/6814b181dc612b4093369fb33e9a80cff387a8ab/kubeflow/trainer/types/types.py#L30-L40

PR #113, opened 2025-10-02
fix(ci): Skip release workflow on forks

kramaranya:
I've noticed that my fork has started a release workflow when I pulled the version update from remote. I disabled this CI on forks


.github/workflows/release.yml
```diff
@@ -80,6 +80,7 @@ jobs:
 
   build:
     name: Build package
+    if: ${{ github.repository == 'kubeflow/sdk' }}
```
andreyvelich: Since all of the jobs after `prepare` have `needs` parameter, you can keep if only for the first Job: `prepare` ?
```suggestion
```

kramaranya: Oh yeah, good point, let me update it

PR #115, opened 2025-10-02
chore(changelog) : Replace Python script with git-cliff for Changelog generation

### LINKED ISSUE #99: Adopt external changelog generation tool

We need to transition from our custom [changelog generation script](https://github.com/kramaranya/sdk/blob/auto-release/scripts/gen-changelog.py) to an existing external tool for future releases.

For the first release, we'll keep using our current script so we don't delay anything. But for future releases, we should use either the [Kubernetes release-notes CLI](https://github.com/kubernetes/release/tree/master/cmd/release-notes) (that Kueue uses) or the [git-cliff](https://github.com/orhun/git-cliff) tool, (or any better solution you might find)

This might require some process changes like updating our PR template to include release notes sections, adding PR labels like `kind/feature`, `kind/bug`, etc.

### Why is this needed?

We don't want to keep maintaining our own changelog script when there are good tools already available that other projects maintain

### Love this feature?

rohansaibuddhi:
Replaces the python script being currently used with git-cliff. 

**Which issue(s) this PR fixes** :
Will close the issue #99 when the PR gets merged. 

Fixes #99

**Checklist:**

-The change closely follows the working of the Python script. In the current state of the repository, both the methods should generate the same changelog. 
-Requires developers to follow "conventional commits" as mentioned in contributing.md.
-Works best with squash merges - https://git-cliff.org/docs/#how-should-i-manage-prs


Makefile
```diff
@@ -71,7 +71,15 @@ release: install-dev
...
-	@uv run python scripts/gen-changelog.py --token=$${GITHUB_TOKEN} --version=$(VERSION)
+	@PREV_TAG=$$(git tag -l --sort=-creatordate | head -1); \
+	CHANGELOG_FILE=CHANGELOG/CHANGELOG-$$(echo $(VERSION) | cut -d. -f1-2).md; \
+	if [ -z "$$PREV_TAG" ]; then \
+	  GITHUB_TOKEN=$${GITHUB_TOKEN} git-cliff --tag $(VERSION) -o $$CHANGELOG_FILE; \
+	elif [ -f "$$CHANGELOG_FILE" ]; then \
+	  GITHUB_TOKEN=$${GITHUB_TOKEN} git-cliff $$PREV_TAG..HEAD --tag $(VERSION) --prepend $$CHANGELOG_FILE; \
+	else \
+	  GITHUB_TOKEN=$${GITHUB_TOKEN} git-cliff $$PREV_TAG..HEAD --tag $(VERSION) -o $$CHANGELOG_FILE; \
+	fi
```
kramaranya: This will generate chnagelog for RCs, could we skip it in this case since we decided that we'll provide changelogs for stable versions only?
cc @andreyvelich

Makefile
```diff
@@ -71,7 +71,15 @@ release: install-dev
 	@V_NO_V=$(VERSION); \
 	sed -i.bak "s/^__version__ = \".*\"/__version__ = \"$$V_NO_V\"/" kubeflow/__init__.py && \
 	rm -f kubeflow/__init__.py.bak
-	@uv run python scripts/gen-changelog.py --token=$${GITHUB_TOKEN} --version=$(VERSION)
+	@PREV_TAG=$$(git tag -l --sort=-creatordate | head -1); \
```
kramaranya: What if the last tag was 0.2.1rc1, the new release 0.2.1 will only show changes since the RC, not since 0.2.0, right? Could you update it so that for stable releases we compute the previous tag from the last stable tag?

cliff.toml
```diff
@@ -0,0 +1,56 @@
...
+  commit_preprocessors = []
+  protect_breaking_commits = false
+  commit_parsers = [
+      { message = "^feat", group = "New Features" },
+      { message = "^fix", group = "Bug Fixes" },
+      { message = "^doc", group = "Documentation" },
+      { message = "^perf", group = "Performance" },
+      { message = "^refactor", group = "Refactor" },
+      { message = "^style", group = "Styling" },
+      { message = "^test", group = "Testing" },
```
kramaranya: We currently only allow feat, fix, chore and revert, see https://github.com/kubeflow/sdk/blob/main/.github/workflows/check-pr-title.yaml#L26-L30

cliff.toml
```diff
@@ -0,0 +1,56 @@
...
+  render_always = true
+
+  [git]
+  conventional_commits = false
+  filter_unconventional = false
+  require_conventional = false
+  split_commits = false
+  commit_preprocessors = []
+  protect_breaking_commits = false
+  commit_parsers = [
```
kramaranya, line 45: Can we also skip commits for Kubeflow SDK releases, like this https://github.com/kubeflow/sdk/commit/46b42d09a4e641a951aa16bef24b7d855440e291?

cliff.toml
```diff
@@ -0,0 +1,56 @@
...
+  {% for group, commits in commits | group_by(attribute="group") %}
+  ## {{ group | striptags | trim | upper_first }}
+
+  {% for commit in commits -%}
+  - {{ commit.message | split(pat="(#") | first | trim }}{% if commit.breaking %} [**breaking**]{% endif %}{% for link in commit.links %} ([{{ link.text }}]({{ link.href }})){% endfor %}{% if commit.github.username %} by @{{ commit.github.username }}{% endif %}
+  {% endfor -%}
+  {% endfor %}
+  """
+  trim = true
+  render_always = true
```
kramaranya: Why do we want to render always?

cliff.toml
```diff
@@ -0,0 +1,56 @@
...
+
+  [git]
```
kramaranya, line 38: Could you please take a look if we could also add a new section "New Contributors" like here https://github.com/orhun/git-cliff/releases/tag/v2.10.1? I think it would be great to have it!

kramaranya (commented):
This looks great, thank you for this @rohansaibuddhi!
Could you please update https://github.com/kubeflow/sdk/blob/main/RELEASE.md as well?

Makefile
rohansaibuddhi: Makes sense. This will lead to duplication of commits between the last stable tag and 0.2.1rc1, considering we do generate changelog for RCs, would that behavior be okay?

In case there is no last stable commit, it would take all commits since the first commit. Is that expected behavior? I know we do not have to deal with this edge case since we already have 0.1.0, but wanted to confirm.

cliff.toml
rohansaibuddhi: would we want this to be an exact match like '^Kubeflow SDK Official Release' or something more flexible like ".*Release [0-9]". If it is the former, we should probably add the convention to RELEASE.md as well, let me know what you think.

cliff.toml
rohansaibuddhi: should we just skip rendering the 'Changelog [version] (date)' in the case there are no changes since the last release?

kramaranya:
Hey @rohansaibuddhi! Sorry for the delayed reply. Could you please rebase your PR and sign the commits?

kramaranya:
This has been resolved with #226 
/close

PR #117, opened 2025-10-03
fix(ci): Fix first-time PR welcome workflow

kramaranya:
I'v noticed the workflow to welcome new contributors [didn't work](https://github.com/kubeflow/sdk/actions/runs/18207106549/job/51839990091?pr=115), so I've fixed the warnings and hopefully it will work :)

/asign @kubeflow/kubeflow-sdk-team

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -1,7 +1,7 @@
 name: Welcome new contributors
 
 on:
-  pull_request_target:
+  pull_request:
     types:
       - opened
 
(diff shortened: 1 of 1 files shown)


.github/workflows/welcome-new-contributors.yaml
```diff
@@ -1,7 +1,7 @@
 name: Welcome new contributors
 
 on:
-  pull_request_target:
+  pull_request:
```
andreyvelich, line 4: Do we need to add permissions: to write to the issue ?
https://github.com/actions/first-interaction?tab=readme-ov-file#usage

kramaranya: Probably yes, I'll also add a message for issues

kramaranya: I updated, please take a look. I'm hoping this will fix the CI

PR #118, opened 2025-10-06
chore: Update minimum required python version to 3.10

szaher:
python 3.9 EOL is end of October as per https://devguide.python.org/versions/
I suggest we move to python3.11 directly which provides better performance.

image

Fixes #

**Checklist:**


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "kubeflow"
 dynamic = ["version"]
-requires-python = ">=3.9"
+requires-python = ">=3.11"
```
kramaranya: Shall we still support 3.10?
cc @andreyvelich @astefanutti

kramaranya (commented):
Thank you!
Please update https://github.com/kubeflow/sdk/blob/main/pyproject.toml#L18

pyproject.toml
andreyvelich: Yes, I would suggest we jump to 3.10 first.
IMO, we should support 3.9 for SDK v0.2 release at least.
The 3.9 EOF is October 2025, and I would imagine we still have users on that version: https://devguide.python.org/versions/

andreyvelich: /hold

Issue #114, opened 2025-10-02
Create Podman Backend to support Local Execution


Fiona-Waters:
### What you would like to be added?

Following on from https://github.com/kubeflow/sdk/issues/108 we propose to also add a Podman backend for use in a Local context.

### Why is this needed?

Podman allows rootless operation, is useful in disconnected environments and having both Podman and Docker implementations will allow users to choose what matches their infra without changing code.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

PR #110, opened 2025-09-27
feat(logging): Implement structured and configurable logging support

anencore94:
This PR implements comprehensive logging support for the Kubeflow SDK as requested in Issue #85.

### Core Logging Infrastructure
- **NullHandler Pattern**: Prevents logging noise by default, users can override with their own configuration
- **Configurable Logging**: Support for console, detailed, and JSON output formats
- **Structured Logging**: JSON formatter for log aggregation systems (ELK stack, Fluentd, etc.)
- **Environment-based Configuration**: Configure logging via environment variables

### Key Components
- `kubeflow/trainer/logging/config.py`: Centralized logging configuration
- `kubeflow/trainer/logging/formatters.py`: Custom formatters including JSON structured logging
- `kubeflow/trainer/logging/logging_test.py`: Comprehensive test suite (16 tests)

### Integration Points
- **SDK Integration**: Added debug logging to `TrainerClient` for better observability
- **Package Integration**: Exposed logging utilities through `kubeflow.trainer.logging`
- **NullHandler Setup**: Configured at package level to prevent logging noise

## Issue #85 Requirements Fulfilled

✅ **Consistent use of Python's logging library instead of print statements**
- All SDK operations now use proper Python logging
- NullHandler pattern prevents unwanted output by default

✅ **Support for different logging levels (DEBUG, INFO, WARNING, ERROR)**
- Full support for all standard Python logging levels
- Configurable via `setup_logging()` function or environment variables

✅ **Ability for users to configure log formatting and destinations**
- Console, detailed, and JSON output formats
- File output support
- Custom formatter support

✅ **Clear and actionable log messages for key SDK operations**
- Debug messages in TrainerClient initialization
- Backend selection logging
- Job creation and ID logging

## Testing

- **16 comprehensive unit tests** covering all logging functionality
- **NullHandler pattern verification** with proper isolation
- **SDK integration testing** with real TrainerClient usage
- **Application integration examples** with file and console logging
- **All tests pass** with proper linting compliance

### Basic Usage
```python
from kubeflow.trainer import TrainerClient, setup_logging

# Setup logging (optional - NullHandler prevents noise by default)
setup_logging(level="DEBUG", format_type="console")

# Use SDK - debug messages will appear if logging is configured
client = TrainerClient()
```

### JSON Logging for Production
```python
setup_logging(level="INFO", format_type="json")
# Logs will be in JSON format suitable for log aggregation
```

### Environment Configuration
```bash
[...]

PR #95, opened 2025-09-15
feat(docs): add readme for localprocess

szaher:
**What this PR does / why we need it**:

add a readme to explain how to use the local-process backend.

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

PR #119, opened 2025-10-07
feat: Add Podman backend and sync Docker backend implementation

Fiona-Waters:

This PR adds a Podman backend to support local execution. In line with this and to avoid duplication of code I have updated some of the Docker backend implementation files.
cc @briangallagher 

Fixes #https://github.com/kubeflow/sdk/issues/114

**Checklist:**


astefanutti:
@Fiona-Waters that sounds to me. I agree the fallback logic may really apply to choose the default container runtime.

Other than that, I'd be inclined to drop the "Local" prefix entirely. Even Kubernetes could run local with KinD, and I doubt the SDK will ever do remote process.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,627 @@
...
+                        if master_ip:
+                            logger.info(f"Master node IP address: {master_ip}")
+                        else:
+                            logger.warning(
+                                "Could not retrieve master IP address. "
+                                "Worker nodes will fall back to DNS resolution."
+                            )
+
+            # Store job in backend
+            self._jobs[job_name] = _Job(
```
astefanutti: Would that be possible to avoid relying on that in-memory "registry" and consistently rely on the state from the container runtime itself?

Fiona-Waters: I've updated to store metadata as labels on containers and networks allowing us to query the container runtime for all job information. See this [commit](https://github.com/kubeflow/sdk/pull/119/commits/24f0bbd68b00286e34d8bb8ab84edc9d78480de7) - please let me know what you think. Thanks

kubeflow/trainer/backends/local_runtime_loader.py
```diff
@@ -0,0 +1,110 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from kubeflow.trainer.types import types as base_types
+
+LOCAL_RUNTIMES_DIR = Path(__file__).parents[1] / "config" / "local_runtimes"
```
astefanutti: Maybe:

```suggestion
LOCAL_RUNTIMES_DIR = Path(__file__).parents[1] / "config" / "container_runtimes"
```

otherwise could be confusing for local process backend?

astefanutti (commented):
@Fiona-Waters thanks for this awesome work!

That looks good to me overall.

/assign @kubeflow/kubeflow-sdk-team @briangallagher

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
 print("\n".join(TrainerClient().get_job_logs(name=job_id)))
 ```
 
+## Local Development
```
andreyvelich, line 74: Can we also include docs about Trainer local execution to the user guides ?
https://www.kubeflow.org/docs/components/trainer/user-guides/
you can also add info from the @szaher PR: https://github.com/kubeflow/sdk/pull/95

kubeflow/trainer/backends/container/README.md
```diff
@@ -0,0 +1,162 @@
+# ContainerBackend
```
andreyvelich: I would suggest that we move this docs to the user guides for now: https://www.kubeflow.org/docs/components/trainer/user-guides/ as we discussed here: https://github.com/kubeflow/sdk/pull/95#discussion_r2397186554

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,716 @@
...
+                    adapter = PodmanClientAdapter(self.cfg.container_host)
+                    adapter.ping()
+                    logger.info("Using Podman as container runtime")
+                    return adapter
+                except Exception as podman_error:
+                    logger.debug(f"Podman initialization failed: {podman_error}")
+                    raise RuntimeError(
+                        "Neither Docker nor Podman is available. "
+                        "Please install Docker or Podman, or use LocalProcessBackendConfig instead."
+                    ) from podman_error
```
andreyvelich: I think, this can be simplified as follows:
```py
runtime_map = {
    "docker": DockerClientAdapter,
    "podman": PodmanClientAdapter,
}

def get_adapter(cfg):
    runtimes_to_try = [cfg.runtime] if cfg.runtime else ["docker", "podman"]

    last_error = None
    for runtime_name in runtimes_to_try:
        if runtime_name not in runtime_map:
            continue
        try:
            adapter = runtime_map[runtime_name](cfg.container_host)
            adapter.ping()
            logger.info(f"Using {runtime_name} as container runtime")
            return adapter
        except Exception as e:
            logger.debug(f"{runtime_name} initialization failed: {e}")
            last_error = e

    raise RuntimeError(
        "Neither Docker nor Podman is available. "
        "Please install Docker or Podman, or use LocalProcessBackendConfig instead."
    ) from last_error
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,716 @@
...
+        ]
+
+        logs = self._adapter.run_oneoff_container(image=image, command=command)
+        print(logs)
+
+    def train(
+        self,
+        runtime: types.Runtime | None = None,
+        initializer: types.Initializer | None = None,
+        trainer: types.CustomTrainer | types.BuiltinTrainer | None = None,
```
andreyvelich: Please don't use `|` since we still support Python 3.9 for now. Let's be consistent across backends:
https://github.com/kubeflow/sdk/blob/24f0bbd68b00286e34d8bb8ab84edc9d78480de7/kubeflow/trainer/backends/kubernetes/backend.py#L183

kubeflow/trainer/config/training_runtimes/torch_distributed.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: trainer.kubeflow.org/v1alpha1
```
andreyvelich: Instead of installing the runtimes, can we just read the image version from GitHub dynamically ?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,716 @@
...
+    ) -> str:
+        if runtime is None:
+            runtime = self.get_runtime("torch-distributed")
+
+        if not isinstance(trainer, types.CustomTrainer):
+            raise ValueError(f"{self.__class__.__name__} supports only CustomTrainer in v1")
+
+        # Generate job name
+        job_name = random.choice(string.ascii_lowercase) + uuid.uuid4().hex[:11]
+        logger.info(f"Starting training job: {job_name}")
```
andreyvelich: Can we use the `.debug()` logging for now, here we can discuss how we want to expose logging: https://github.com/kubeflow/sdk/pull/110 in Kubeflow SDK.
cc @anencore94

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,716 @@
...
+        # Remove network (best-effort)
+        if network_id:
+            with suppress(Exception):
+                self._adapter.delete_network(network_id)
+
+        # Remove working directory if configured
+        if self.cfg.auto_remove and workdir_host and os.path.isdir(workdir_host):
+            shutil.rmtree(workdir_host, ignore_errors=True)
+
+    # Helper methods
```
andreyvelich: Shall we move those helper functions to `backends/container/utils.py` to make it consistent with other directories?

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,56 @@
...
+from typing import Literal, Optional, Union
+
+from pydantic import BaseModel, Field
+
+
+class ContainerBackendConfig(BaseModel):
+    image: Optional[str] = Field(default=None)
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    gpus: Optional[Union[int, bool]] = Field(default=None)
```
andreyvelich: How this is used ?

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,56 @@
...
+    env: Optional[dict[str, str]] = Field(default=None)
+    container_host: Optional[str] = Field(default=None)
+    workdir_base: Optional[str] = Field(default=None)
```
andreyvelich: Can we initially use the default dir and if users require to configure it we will give them such option ?

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,56 @@
...
+
+from pydantic import BaseModel, Field
+
+
+class ContainerBackendConfig(BaseModel):
+    image: Optional[str] = Field(default=None)
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    gpus: Optional[Union[int, bool]] = Field(default=None)
+    env: Optional[dict[str, str]] = Field(default=None)
```
andreyvelich: Do we start a new container every time Job is submitted ?
If yes, this might be controlled via `train()` API.

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,56 @@
...
+    container_host: Optional[str] = Field(default=None)
+    workdir_base: Optional[str] = Field(default=None)
+    runtime: Optional[Literal["docker", "podman"]] = Field(default=None)
```
andreyvelich: to make it less confusing with TrainingRuntime, can we name it:
```suggestion
    container_runtime: Optional[Literal["docker", "podman"]] = Field(default="docker")
```

kubeflow/trainer/backends/container/client_adapter.py
```diff
@@ -0,0 +1,606 @@
...
+a common interface.
+"""
+
+from __future__ import annotations
+
+import abc
+from collections.abc import Iterator
+
+
+class ContainerClientAdapter(abc.ABC):
```
andreyvelich: Can we call it as BaseContainerClientAdapter(), similar to: https://github.com/kubeflow/sdk/blob/24f0bbd68b00286e34d8bb8ab84edc9d78480de7/kubeflow/trainer/backends/base.py#L23
 I would suggest, we move them to subdirectory:
 ```
container/adapters/base.py
container/adapters/docker.py
container/adapters/podman.py
```

WDYT @Fiona-Waters ?

**Review by andreyvelich — COMMENTED** [review]
Thank you @Fiona-Waters!
I left my initial messages.

File: kubeflow/trainer/config/training_runtimes/torch_distributed.yaml (hunk 7 above)
Comment (Fiona-Waters) [reply kubeflow/trainer/config/training_runtimes/torch_distributed.yaml]: Let me look into that. For offline support should we fall back to providing this?

File: kubeflow/trainer/backends/container/types.py (hunk 10 above)
Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/types.py]: It allows you to override the default container image specified in the ClusterTrainingRuntime.

Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/types.py]: I think actually you are referring to `gpus`. It was pulled over from a previous iteration but isn't being used currently. I can update to include GPU support.

Comment (astefanutti) [reply kubeflow/trainer/backends/container/types.py]: I'm a bit confused by this as there can be multiple ClusterTraininingRuntimes and it's defined by TrainJobs.

Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/types.py]: Apologies I understand now that it doesn't make sense to allow the image to be updated here as it is not per job. Will remove this. Thank you.

File: kubeflow/trainer/backends/container/types.py (hunk 11 above)
Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/types.py]: I can remove it for now and use the default dir.

File: kubeflow/trainer/backends/container/types.py (hunk 12 above)
Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/types.py]: Yes, you're right - similar to the image param. Will remove this. Thanks

File: kubeflow/trainer/backends/container/client_adapter.py (hunk 14 above)
Comment (Fiona-Waters) [reply kubeflow/trainer/backends/container/client_adapter.py]: Yes good idea, will do.
BTW thank you for your review. Will update the PR tomorrow hopefully.

File: README.md (hunk 3 above)
Comment (Fiona-Waters) [reply README.md:74]: WIP PR for this https://github.com/kubeflow/website/pull/4221

File: kubeflow/trainer/backends/container/utils.py [hunk 15]
```diff
@@ -0,0 +1,266 @@
...
+from kubeflow.trainer.types import types
+
+logger = logging.getLogger(__name__)
+
+
+def create_workdir(job_name: str, backend_class_name: str) -> str:
+    """
+    Create per-job working directory on host.
+
+    Working directories are created under ~/.kubeflow_trainer/localcontainer/<job_name>
```
astefanutti: Maybe `~/.kubeflow/trainer/containers/...` ?

kubeflow/trainer/backends/container_runtime_loader.py
```diff
@@ -0,0 +1,417 @@
...
+import urllib.request
+import urllib.error
+
+import yaml
+
+from kubeflow.trainer.types import types as base_types
+
+logger = logging.getLogger(__name__)
+
+CONTAINER_RUNTIMES_DIR = Path(__file__).parents[1] / "config" / "container_runtimes"
```
astefanutti: I think I've suggested "container_runtimes" before, but looking at it maybe `training_runtimes` would be more appropriate?

kubeflow/trainer/backends/container_runtime_loader.py
```diff
@@ -0,0 +1,417 @@
...
+CACHE_DIR = Path.home() / ".kubeflow_trainer" / "runtime_cache"
```
astefanutti: Maybe `~/.kubeflow/trainer/cache`?

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,417 @@
...
+
+logger = logging.getLogger(__name__)
+
+CONTAINER_RUNTIMES_DIR = Path(__file__).parents[1] / "config" / "container_runtimes"
+CACHE_DIR = Path.home() / ".kubeflow_trainer" / "runtime_cache"
+CACHE_DURATION = timedelta(hours=24)
+
+# GitHub runtimes configuration
+GITHUB_RUNTIMES_BASE_URL = (
+    "https://raw.githubusercontent.com/kubeflow/trainer/master/manifests/base/runtimes"
```
astefanutti: Note for later, we should probably find a way rely on released runtimes.

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,417 @@
+# Copyright 2025 The Kubeflow Authors.
```
astefanutti: `training_runtime_loader.py`?

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,53 @@
...
+from pydantic import BaseModel, Field
+
+
+class ContainerBackendConfig(BaseModel):
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    gpus: Optional[Union[int, bool]] = Field(default=None)
+    container_host: Optional[str] = Field(default=None)
+    container_runtime: Optional[Literal["docker", "podman"]] = Field(default=None)
+    use_github_runtimes: bool = Field(default=True)
```
astefanutti: Maybe we can be more structured here and have a `training_runtimes` argument, that could have different options, like one to some URLs maybe.

kubeflow/trainer/backends/container/runtime_loader.py
```diff
@@ -0,0 +1,27 @@
...
+# limitations under the License.
+
+"""
+Runtime loader for the Container backend.
+
+This module provides container-agnostic imports for the shared container runtime loader.
+"""
+
+from kubeflow.trainer.backends.container_runtime_loader import (
+    CONTAINER_RUNTIMES_DIR,
```
astefanutti:
```suggestion
    TRAINING_RUNTIMES_DIR,
```

kubeflow/trainer/backends/container/runtime_loader.py
```diff
@@ -0,0 +1,27 @@
...
+    get_container_runtime,
```
astefanutti:
```suggestion
    get_training_runtime,
```

kubeflow/trainer/backends/container/runtime_loader.py
```diff
@@ -0,0 +1,27 @@
...
+    list_container_runtimes,
```
astefanutti:
```suggestion
    list_training_runtimes,
```

kubeflow/trainer/backends/container_runtime_loader.py
Fiona-Waters: I think this would be confused with the training runtime definition like "torch_distributed". What about `engine_runtimes` or `container_engine`?
or `training_runtime_engine` but that might be too long

astefanutti: My understanding is that there is "container runtime" or "runtime backend" and "training runtime".

In that file here, my understanding is that it's about "training runtimes" right? or maybe there are training runtimes configuration per container runtimes?

kubeflow/trainer/config/training_runtimes/torch_distributed.yaml
andreyvelich: @Fiona-Waters Do we require only image ?
We can fallback to the constant, that we define somewhere.

kubeflow/trainer/backends/container_runtime_loader.py
Fiona-Waters: Oh sorry I totally confused the whole thing. Yes this should be training_runtimes. Will update where relevant.

kubeflow/trainer/backends/container/types.py
Fiona-Waters: To allow maybe github urls, local files, and other urls. Something like this?
```
class TrainingRuntimeSource(BaseModel):
      # GitHub
      github_enabled: bool = Field(default=True)
      github_base_url: Optional[str] = Field(default=None)

      # Local filesystem directories
      custom_paths: list[str] = Field(default_factory=list)

      # Remote HTTP(S) URLs
      http_urls: list[str] = Field(default_factory=list)
```

astefanutti: Yes, maybe a list of URLs where different scheme would be supported, e.g. `file://`, `https://`, and that would be more extensible?

kubeflow/trainer/config/training_runtimes/torch_distributed.yaml
Fiona-Waters: Is it okay to hardcode constants that will then need to be updated manually later? Something like this:
```
DEFAULT_FRAMEWORK_IMAGES = {
      "torch": "pytorch/pytorch:2.7.1-cuda12.8-cudnn9-runtime",
  }
```
This would then be used to create a default runtime, if the user doesn't provide a runtime url (and we have removed the sample one).

kubeflow/trainer/backends/container/types.py
Fiona-Waters: Ok I've pushed an update, ptal. Thank you.

kubeflow/trainer/config/training_runtimes/torch_distributed.yaml
Fiona-Waters: I've pushed changes to use a default image if a user doesn't provide a URL and have removed the yaml. Please review. Thanks

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
...
+
+# Auto-detects Docker or Podman
+config = ContainerBackendConfig()
+client = TrainerClient(backend_config=config)
+
+# Your training runs in isolated containers
+job_id = client.train(trainer=CustomTrainer(func=train_fn))
+```
+
+For detailed configuration options and platform-specific setup (macOS, Linux), see the [ContainerBackend documentation](kubeflow/trainer/backends/container/README.md).
```
briangallagher: Does this file exist?

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
 print("\n".join(TrainerClient().get_job_logs(name=job_id)))
 ```
 
+## Local Development
+
+Kubeflow SDK provides first-class support for local development, allowing you to test and iterate on your models without needing a Kubernetes cluster.
```
kramaranya:
```suggestion
Kubeflow Trainer client supports local development without needing a Kubernetes cluster:
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,576 @@
...
+                adapter = runtime_map[runtime_name](self.cfg.container_host)
+                adapter.ping()
+                logger.debug(f"Using {runtime_name} as container runtime")
+                return adapter
+            except Exception as e:
+                logger.debug(f"{runtime_name} initialization failed: {e}")
+                last_error = e
+
+        raise RuntimeError(
+            "Neither Docker nor Podman is available. "
```
briangallagher: From my testing it seems that this error is thrown when Docker is available and installed but the socket is not in the default location. I'm testing on a mac with docker/colima. Can we assume that a high percentage of users will be using mac/coilma? In which case we could attempt to find the socket at the default colima location:
`host=f"unix://{os.path.expanduser('~')}/.colima/default/docker.sock"` after initially attempting without providing a host. 

Similarly, with podman, try with the host provided by the user, then the default, then some "known locations" such as `host=f"unix://{os.path.expanduser('~')}/.local/share/containers/podman/machine/podman.sock"`

At the moment the error message is a bit misleading, `Neither Docker nor Podman is available.` They may well be available, just that the socket is not in the expected/default location. Something like: "Verify the location of the socket........" 

We could also advise in the documentation on how to find/confirm the socket for docker and podman.

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
...
+
+### Execution Backends
+
+Choose the backend that fits your development workflow:
+
+| Backend | Description | Use Case |
+|---------|-------------|----------|
+| **KubernetesBackend** | Run jobs on Kubernetes cluster | Production, multi-node distributed training |
+| **ContainerBackend** | Auto-detects Docker or Podman | Local development with container isolation |
+| **LocalProcessBackend** | Run as local Python subprocesses | Quick prototyping, debugging |
```
kramaranya: I'd simplify this:
```suggestion
### Available Backends

- **KubernetesBackend** (default) - Production training on Kubernetes
- **ContainerBackend** - Local development with Docker/Podman isolation  
- **LocalProcessBackend** - Quick prototyping with Python subprocesses
```

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
...
+
+The **ContainerBackend** automatically detects and uses either Docker or Podman:
+
+```bash
+# Install with Docker support
+pip install kubeflow[docker]
+
+# Or install with Podman support
+pip install kubeflow[podman]
+```
```
kramaranya: Since this is Trainer-specific, I'd keep this high-level, just mention availability and point to the docs:
```suggestion
**Quick Start:**

Install container support: `pip install kubeflow[docker]` or `pip install kubeflow[podman]`
```

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
...
+
+```python
+from kubeflow.trainer import TrainerClient, ContainerBackendConfig, CustomTrainer
+
+# Auto-detects Docker or Podman
+config = ContainerBackendConfig()
+client = TrainerClient(backend_config=config)
```
kramaranya:
```suggestion
# Switch to local container execution
client = TrainerClient(backend_config=ContainerBackendConfig())
```

README.md
```diff
@@ -71,6 +71,45 @@ TrainerClient().wait_for_job_status(job_id)
...
+
+# Your training runs in isolated containers
+job_id = client.train(trainer=CustomTrainer(func=train_fn))
```
kramaranya:
```suggestion
# Your training runs locally in isolated containers
job_id = client.train(trainer=CustomTrainer(func=train_fn))
```

README.md
kramaranya: Since we publish Kubeflow SDK to PyPI, we need to use the full link, so PyPI can generate it https://github.com/kubeflow/sdk/tree/main/kubeflow/trainer/...

kramaranya: Also, do we still want to create readme for that? I'd prefer pointing out at the guide in Kubeflow website

kubeflow/trainer/__init__.py
```diff
@@ -15,6 +15,7 @@
 
 # Import the Kubeflow Trainer client.
 from kubeflow.trainer.api.trainer_client import TrainerClient  # noqa: F401
+from kubeflow.trainer.backends.container.types import ContainerBackendConfig
```
kramaranya: Shall this be under https://github.com/Fiona-Waters/kubeflow-sdk/blob/podman-backend/kubeflow/trainer/__init__.py#L20?

kubeflow/trainer/backends/container/adapters/base.py
```diff
@@ -0,0 +1,196 @@
...
+        self,
+        image: str,
+        command: list[str],
+        name: str,
+        network_id: str,
+        environment: dict[str, str],
+        labels: dict[str, str],
+        volumes: dict[str, dict[str, str]],
+        working_dir: str,
+        gpus: int | bool | None = None,
```
kramaranya: This requires python 3.10, since we still support 3.9, could you use optional/union instead?
https://peps.python.org/pep-0604/

kubeflow/trainer/backends/container/utils.py
```diff
@@ -0,0 +1,237 @@
...
+            if not adapter.image_exists(image):
+                raise RuntimeError(f"Image '{image}' not found locally and pull policy is Never")
+            return
+        if policy == "always":
+            logger.debug(f"Pulling image (Always): {image}")
+            adapter.pull_image(image)
+            return
+        # IfNotPresent
+        if not adapter.image_exists(image):
+            logger.debug(f"Pulling image (IfNotPresent): {image}")
```
briangallagher, line 200: I would consider making this a warning.
Reason being, most people will not run with debug logging when running trainer locally and for the first time  so they won't see this log.
When running locally and the image does not exist it can take some time for it to download given that we are using a pytorch image. It can seem like it is hanging, it might be enough for people just to give up. 
I would suggest changing to a warning and a message something like "Pulling image, this might take some time......."

briangallagher (changes requested):
@Fiona-Waters Looks great, I've made some suggestions to improve the user experience a bit. In particular, first time users.

kubeflow/trainer/backends/container/adapters/base.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            List of container info dictionaries with keys:
+            - id: Container ID
+            - name: Container name
+            - labels: Dictionary of labels
+            - status: Container status
+            - created: Creation timestamp
+        """
+        raise NotImplementedError()
```
kramaranya: I think signature should be `dict[str, list[str]]` since it accepts `{"label": ["key=value"]}`, right?

README.md
Fiona-Waters: No we moved it to guides PR. Will update. Thanks

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: I can update this with more validation for sure, and try those 2 common sockets before showing an error. The instructions for using colima and podman machine on mac are in the guides PR.

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,71 @@
...
+        ),
+    )
+
+
+class ContainerBackendConfig(BaseModel):
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    gpus: Optional[Union[int, bool]] = Field(default=None)
+    container_host: Optional[str] = Field(default=None)
+    container_runtime: Optional[Literal["docker", "podman"]] = Field(default="docker")
```
kramaranya: What would happen if I have only podman installed? Would this actually fallback to podman if no docker is installed?

kubeflow/trainer/backends/container/utils.py
Fiona-Waters: We want to use the debug logging for now but can update later. See comment [here](https://github.com/kubeflow/sdk/pull/119#discussion_r2445619351)

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,576 @@
...
+                labels = {
+                    f"{self.label_prefix}/trainjob-name": job_name,
+                    f"{self.label_prefix}/step": f"node-{rank}",
+                    f"{self.label_prefix}/network-id": network_id,
+                    f"{self.label_prefix}/num-nodes": str(num_nodes),
+                }
+
+                volumes = {
+                    workdir: {
+                        "bind": "/workspace",
```
kramaranya: we can use WORKSPACE_PATH from constants

kubeflow/trainer/backends/container/utils.py
```diff
@@ -0,0 +1,237 @@
...
+    script_path = Path(workdir) / "train.py"
+    import inspect
+    import textwrap
+
+    code = inspect.getsource(trainer.func)
+    code = textwrap.dedent(code)
+    if trainer.func_args is None:
+        code += f"\n{trainer.func.__name__}()\n"
+    else:
+        code += f"\n{trainer.func.__name__}({trainer.func_args})\n"
```
kramaranya: We should unpack args, right? cc @briangallagher @andreyvelich 
```suggestion
        code += f"\n{trainer.func.__name__}(**{trainer.func_args})\n"
```

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: https://github.com/kubeflow/sdk/pull/119/commits/bdcac824779c7b97dc02ea0c9cecd825ada893a1 PTAL, thanks

README.md
Fiona-Waters: The readme has been removed and info is added in the guides PR.

kubeflow/trainer/backends/container/utils.py
briangallagher: good catch, yes, similar to [default behaviour](https://github.com/kubeflow/sdk/blob/f26d91480b3c4c96b84f2100cb4b3d19beee0b02/kubeflow/trainer/utils/utils.py#L336)

kubeflow/trainer/backends/container/types.py
Fiona-Waters: There shouldn't be a default, good catch! Thank you!

kubeflow/trainer/backends/container/adapters/base.py
Fiona-Waters: Would it make sense to include these in the ignore block in the pyproject.toml? - since running `make verify` currently catches errors relating to these.
```
"UP006", # Use list instead of List (requires Python 3.10+)
"UP007", # Use X | Y instead of Union[X, Y] (requires Python 3.10+)
"UP035", # typing.List is deprecated (requires Python 3.10+)
"UP045", # Use X | None instead of Optional[X] (requires Python 3.10+)
```

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,71 @@
...
+            "If a runtime is not found in configured sources, built-in runtimes "
+            "packaged with kubeflow-trainer are used as default."
+        ),
+    )
+
+
+class ContainerBackendConfig(BaseModel):
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    gpus: Optional[Union[int, bool]] = Field(default=None)
```
astefanutti: Should `gpus` also be per-job?

kubeflow/trainer/backends/container/types.py
astefanutti: Thanks, looks great!

kubeflow/trainer/backends/container/types.py
Fiona-Waters: Yes you're right, I've removed it. Thank you.

kubeflow/trainer/backends/container/types.py
```diff
@@ -0,0 +1,67 @@
...
+
+class ContainerBackendConfig(BaseModel):
+    pull_policy: str = Field(default="IfNotPresent")
+    auto_remove: bool = Field(default=True)
+    container_host: Optional[str] = Field(default=None)
+    container_runtime: Optional[Literal["docker", "podman"]] = Field(default=None)
+    runtime_source: TrainingRuntimeSource = Field(
+        default_factory=TrainingRuntimeSource,
+        description="Configuration for training runtime sources",
+    )
```
andreyvelich, line 67: Why you need another class: TrainingRuntimeSource ?
You can directly define the list in the runtime_source.

@Fiona-Waters @astefanutti Are we planning to extend TrainingRuntimeSource in the future to support additional parameters (e.g. for auth)?

kubeflow/trainer/backends/container/adapters/base.py
```diff
@@ -0,0 +1,197 @@
...
+"""
+
+from __future__ import annotations
+
+import abc
+from collections.abc import Iterator
+from typing import Dict, List, Optional, Tuple
+
+
+class BaseContainerClientAdapter(abc.ABC):
```
andreyvelich, line 30: Shall we rely on Pydantic Baseclass everywhere ?
```suggestion
class BaseContainerClientAdapter(BaseModel):
```

kubeflow/trainer/backends/container/adapters/base.py
```diff
@@ -0,0 +1,197 @@
...
+This module implements the adapter pattern to abstract away differences between
+Docker and Podman APIs, allowing the backend to work with either runtime through
+a common interface.
+"""
+
+from __future__ import annotations
+
+import abc
+from collections.abc import Iterator
+from typing import Dict, List, Optional, Tuple
```
andreyvelich: Please can you use the default types when you can (dict, list, tuple), since these are deprecated
```suggestion
from typing import Optional
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,648 @@
...
+        job_name = random.choice(string.ascii_lowercase) + uuid.uuid4().hex[:11]
+        logger.debug(f"Starting training job: {job_name}")
+
+        try:
+            # Create per-job working directory on host
+            workdir = container_utils.create_workdir(job_name, self.__class__.__name__)
+            logger.debug(f"Created working directory: {workdir}")
+
+            _ = container_utils.write_training_script(workdir, trainer)
+            logger.debug(f"Wrote training script to {workdir}/train.py")
```
andreyvelich: Don't we want to write training script as part of container command like we do in Kubernetes ?
https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/constants/constants.py#L132-L136

I guess, right now we create this `train.py` file is user's workspace which is unnecessary.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,648 @@
...
+                        f"echo 'Timeout waiting for master node'; exit 1; fi; "
+                        f"  sleep 2; "
+                        f"done; "
+                    )
+
+                entry_cmd = (
+                    f"{pre_install_cmd}"
+                    f"{wait_for_master}"
+                    "if command -v torchrun >/dev/null 2>&1; then "
+                    f"  torchrun --nproc_per_node=1 --nnodes={num_nodes} "
```
andreyvelich: Why nproc_per_node always = 1?

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
...
+    Args:
+        runtime_file: YAML filename to fetch
+        owner: GitHub repository owner (default: "kubeflow")
+        repo: GitHub repository name (default: "trainer")
+        branch: Git branch name (default: "master")
+        path: Path to runtimes directory (default: "manifests/base/runtimes")
+
+    Returns None if fetch fails (network error, timeout, etc.)
+    """
+    url = f"https://raw.githubusercontent.com/{owner}/{repo}/{branch}/{path}/{runtime_file}"
```
andreyvelich: you should read runtime image versions from the overlays: https://github.com/kubeflow/trainer/blob/master/manifests/overlays/runtimes/kustomization.yaml#L7

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
...
+        return None
+
+    try:
+        with open(cache_file) as f:
+            data = json.load(f)
+
+        cached_time = datetime.fromisoformat(data["cached_at"])
+        if datetime.now() - cached_time > CACHE_DURATION:
+            logger.debug("Runtime list cache expired")
+            return None
```
andreyvelich: Do we really need to expire Cache ?

kubeflow/trainer/backends/training_runtime_loader_test.py
```diff
@@ -0,0 +1,430 @@
+# Copyright 2025 The Kubeflow Authors.
```
andreyvelich: Please refactor this test file, with test-cases similar to this: 
https://github.com/kubeflow/sdk/blob/0449cd8957602f29df941aa81192a636654e0ad4/kubeflow/trainer/backends/kubernetes/backend_test.py#L779


We should just create single test function with test cases (e.g. test_discover_github_runtime_files)

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
...
+CACHE_DURATION = timedelta(hours=24)
+
+# GitHub runtimes configuration
+GITHUB_RUNTIMES_BASE_URL = (
+    "https://raw.githubusercontent.com/kubeflow/trainer/master/manifests/base/runtimes"
+)
+GITHUB_RUNTIMES_TREE_URL = "https://github.com/kubeflow/trainer/tree/master/manifests/base/runtimes"
+
+
+def _load_runtime_from_yaml(path: Path) -> dict[str, Any]:
```
andreyvelich: Why do you want to keep these functions private if all functions in this file are private ?
```suggestion
def load_runtime_from_yaml(path: Path) -> dict[str, Any]:
```

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
+# Copyright 2025 The Kubeflow Authors.
```
andreyvelich: Can you move these functions to the `backend/container/runtime_loader.py` for now ?

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
...
+    replicated = templ.get("replicatedJobs", [])
+    node_jobs = [j for j in replicated if j.get("name") == "node"]
+    if not node_jobs:
+        raise ValueError(
+            f"Runtime {name} from {source} must define replicatedJobs with a 'node' entry"
+        )
+    node_spec = node_jobs[0].get("template", {}).get("spec", {}).get("template", {}).get("spec", {})
+    containers = node_spec.get("containers", [])
+    if not containers or not containers[0].get("image"):
+        raise ValueError(f"Runtime {name} from {source} 'node' must specify containers[0].image")
```
andreyvelich: You can remove all of these, if you read image from the overlays.

kubeflow/trainer/backends/training_runtime_loader.py
```diff
@@ -0,0 +1,657 @@
...
+
+    Raises:
+        ValueError: If the runtime is not found
+    """
+    for rt in list_training_runtimes(use_github=use_github):
+        if rt.name == name:
+            return rt
+    raise ValueError(
+        f"Runtime '{name}' not found. Available runtimes: "
+        f"{[rt.name for rt in list_training_runtimes(use_github=use_github)]}"
```
andreyvelich: This is unused.

kubeflow/trainer/backends/container/backend_test.py
```diff
@@ -0,0 +1,779 @@
...
+
+        assert test_case.expected_status == SUCCESS
+        assert job.status == test_case.config["expected_job_status"]
+
+    except Exception as e:
+        assert type(e) is test_case.expected_error
+    print("test execution complete")
+
+
+class TestSocketLocationDetection:
```
andreyvelich: Please can you also use @pytest.mark.parametrize here ?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,648 @@
...
+        try:
+            job_runtime = self.get_runtime(runtime_name) if runtime_name else None
+        except Exception as e:
+            raise ValueError(f"Runtime {runtime_name} not found for job {name}") from e
+
+        if not job_runtime:
+            raise ValueError(f"Runtime {runtime_name} not found for job {name}")
+
+        # Parse creation timestamp
+        created_str = network_labels.get(f"{self.label_prefix}/created", "")
```
andreyvelich: Does container API have creation timestamp ?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,648 @@
...
+                    name=step_name,
+                    pod_name=container["name"],
+                    status=container_utils.get_container_status(self._adapter, container["id"]),
+                )
+            )
+
+        # Get num_nodes from labels
+        num_nodes = int(
+            containers[0]["labels"].get(f"{self.label_prefix}/num-nodes", len(containers))
+        )
```
andreyvelich: Can we read num_nodes as number of containers created for the appropriate TrainJob ?

kubeflow/trainer/backends/container/utils.py
```diff
@@ -0,0 +1,236 @@
...
+import os
+from pathlib import Path
+
+from kubeflow.trainer.constants import constants
+from kubeflow.trainer.types import types
+
+logger = logging.getLogger(__name__)
+
+
+def create_workdir(job_name: str, backend_class_name: str) -> str:
```
andreyvelich:
```suggestion
def create_workdir(job_name: str) -> str:
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -0,0 +1,648 @@
...
+                continue
+
+            # Get runtime object
+            try:
+                job_runtime = self.get_runtime(runtime_name) if runtime_name else None
+            except Exception:
+                job_runtime = None
+
+            if not job_runtime:
+                continue
```
andreyvelich: Can you just consolidate code between get_job and list_jobs API ?
Check example: https://github.com/kubeflow/sdk/blob/57c0a40392a0c35c712918c3867020e8bdc29eac/kubeflow/optimizer/backends/kubernetes/backend.py#L194

kubeflow/trainer/backends/container/types.py
Fiona-Waters: Yes the idea here is for future extensibility.

kubeflow/trainer/backends/container/adapters/base.py
Fiona-Waters: `BaseContainerClientAdapter` has no fields to validate—it only defines method signatures like ping() and create_network() that
  Docker/Podman adapters must implement, which is what abc.ABC is designed for according to Python's official documentation.
  https://peps.python.org/pep-3119/

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: Yes you're right, will refactor this. Thanks

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: It shouldn't be. Refactoring for it to be determined based on GPU count and default to 1 if no GPU.

kubeflow/trainer/backends/training_runtime_loader.py
Fiona-Waters: Should there be a ghcr.io/kubeflow/trainer/torch-runtime entry? 
I can map to the image versions from the overlays and at the moment for torch-distributed it will use the `DEFAULT_FRAMEWORK_IMAGES` torch constant. Is that okay?
```
DEFAULT_FRAMEWORK_IMAGES = {
    "torch": "pytorch/pytorch:2.7.1-cuda12.8-cudnn9-runtime",
}
```

andreyvelich: Oh, that is good point, for torch we just use the upstream image.
In that case, let's just take the image version from the `runtime` manifests as you do now.
We can refactor it later.

kubeflow/trainer/backends/training_runtime_loader.py
Fiona-Waters: No but if runtime versions are updated in the repo, they won't get to the user unless they clear the cache themselves. Let me know how to proceed with this one please. Thanks

kubeflow/trainer/backends/training_runtime_loader.py
Fiona-Waters: I will move these to runtime_loader.py where some will be public and some private.

kubeflow/trainer/backends/container/adapters/base.py
kramaranya: Yes, we should ignore UP007 and UP045 for now

kramaranya: Regarding UP006 & UP035, please use built in generic types such as list, dict, tuple etc instead of importing them from typing

andreyvelich: @Fiona-Waters @kramaranya We should not use `|` in Kubeflow SDK since we still support Python 3.9.
Please can you change it back to `Union` for now.

kubeflow/trainer/backends/training_runtime_loader.py
Fiona-Waters: Leaving these since we will stick with reading images from the manifests.

kubeflow/trainer/backends/container/adapters/base.py
kramaranya: Yes, exactly. The issue is that Ruff is suggesting python 3.10+ syntax (including | ), so we should ignore that in pyproject while we still support 3.9

kubeflow/trainer/backends/training_runtime_loader.py
Fiona-Waters: Will remove it as well as list_training_runtimes - they were replaced but not removed.

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: Yes, will update to use the first containers timestamp. Thanks

kubeflow/trainer/backends/container/adapters/base.py
Fiona-Waters: @andreyvelich listing `UP007` will allow us to ignore linting errors asking us to use `|` instead of `union`.
@kramaranya thanks. Will include `UP007` and `UP045` in the pyproject.toml

andreyvelich: @Fiona-Waters The problem is that users who are still on Python 3.9 will get errors while running that code, isn't ?

kramaranya: > The problem is that users who are still on Python 3.9 will get errors while running that code, isn't ?

Why would they? We agreed to use `Union` instead of `|` operator in our codebase. Because of that, ruff is giving some suggestions that we need to ignore.

andreyvelich: That's right. @Fiona-Waters Did you push the change to replace all `|` with `Union` ?

Fiona-Waters: Yes I did. Thank you

kubeflow/trainer/backends/training_runtime_loader.py
andreyvelich: It's fine to keep it for now.

kubeflow/trainer/backends/container/utils.py
```diff
@@ -0,0 +1,236 @@
...
+
+    Args:
+        adapter: Container client adapter (DockerClientAdapter or PodmanClientAdapter).
+        image: Container image name.
+        pull_policy: Pull policy ("IfNotPresent", "Always", or "Never").
+
+    Raises:
+        RuntimeError: If image is not found or pull fails.
+    """
+    policy = (pull_policy or "IfNotPresent").lower()
```
andreyvelich: you don't need this since pull_policy can't be None at this point.
```suggestion
    policy = pull_policy.lower()
```

https://github.com/kubeflow/sdk/blob/9bf4cb7cfd56f5e45c3ceee89023733a6be650db/kubeflow/trainer/backends/container/types.py#L60

kubeflow/trainer/backends/container/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kubeflow.trainer.backends.container.backend import ContainerBackend
+from kubeflow.trainer.backends.container.types import (
+    ContainerBackendConfig,
+    TrainingRuntimeSource,
+)
+
+__all__ = ["ContainerBackend", "ContainerBackendConfig", "TrainingRuntimeSource"]
```
andreyvelich: We don't really need it since we try not to import * from the package, right ?
```suggestion
```

kubeflow/trainer/backends/container/adapters/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+from kubeflow.trainer.backends.container.adapters.base import BaseContainerClientAdapter
+from kubeflow.trainer.backends.container.adapters.docker import DockerClientAdapter
+from kubeflow.trainer.backends.container.adapters.podman import PodmanClientAdapter
+
+__all__ = [
+    "BaseContainerClientAdapter",
+    "DockerClientAdapter",
+    "PodmanClientAdapter",
+]
```
andreyvelich:
```suggestion
```

andreyvelich:
Thanks for this great contribution @Fiona-Waters!
/lgtm
/assign @astefanutti @Electronic-Waste @kramaranya

astefanutti:
Thanks @Fiona-Waters for this awesome work!

/lgtm
/approve

Fiona-Waters:
Thanks @andreyvelich @astefanutti, delighted to have made this contribution. Thank you, and to all reviewers for their help.
The [PR for website guides](https://github.com/kubeflow/website/pull/4221) and the [PR for local mode notebook example](https://github.com/kubeflow/trainer/pull/2907) can be reviewed/merged now. Thanks 🥳

Issue #46, opened 2025-07-23
Support Hyperparameter Optimization in Kubeflow SDK


kramaranya:
We want to migrate [Katib SDK](https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py) to the Unified SDK, similar to what did with [Trainer SDK](https://github.com/kubeflow/sdk/tree/main/python/kubeflow/trainer).

This is the tracking issue:

- [x] Investigate Katib SDK and see how can we migrate it before creating new APIs
- [x] Investigate whether we can release Kubeflow SDK to PyPI with git reference for modules
- [x] https://github.com/kubeflow/katib/pull/2579
- [x] #124 
- [x] `get_job()` API for OptimizerClient
- [x] `list_jobs()` API OptimizerClient
- [x] `delete_job()` API OptimizerClient
- [ ] Testing

andreyvelich:
/retitle Support Hyperparameter Optimization in Kubeflow SDK

PR #124, opened 2025-10-13
feat: Hyperparameter Optimization APIs in Kubeflow SDK

andreyvelich:
Part of: https://github.com/kubeflow/sdk/issues/46
Depends on: https://github.com/kubeflow/katib/pull/2579

I've added initial support for hyperparameter optimization with `OptimizerClient()` into Kubeflow SDK.
This PR also introduced some refactoring to re-use code across `TrainerClient()` and `OptimizerClient()`.

Working example:

```py
def train_func(lr: str, num_epochs: str):
    import time
    import random

    for i in range(10):
        time.sleep(1)
        print(f"Training {i}, lr: {lr}, num_epochs: {num_epochs}")

    print(f"loss={round(random.uniform(0.77, 0.99), 2)}")

OptimizerClient().optimize(
    TrainJobTemplate(
        trainer=CustomTrainer(train_func, num_nodes=2),
    ),
    search_space={
        "lr": Search.loguniform(0.01, 0.05),
        "num_epochs": Search.choice([2, 4, 5]),
    },
)
```

<details>
<summary>Katib Experiment</summary>

```yaml
apiVersion: kubeflow.org/v1beta1
kind: Experiment
metadata:
  name: o759d77408d2
  namespace: default
spec:
  algorithm:
    algorithmName: random
  maxTrialCount: 10
  metricsCollectorSpec:
    collector:
      kind: StdOut
  objective:
    metricStrategies:
      - name: loss
        value: min
    objectiveMetricName: loss
    type: minimize
  parallelTrialCount: 1
  parameters:
    - feasibleSpace:
        distribution: logUniform
        max: "0.05"
        min: "0.01"
      name: lr
      parameterType: double
    - feasibleSpace:
        distribution: uniform
        list:
          - "2"
          - "4"
          - "5"
      name: num_epochs
      parameterType: categorical
  resumePolicy: Never
  trialTemplate:
    failureCondition: status.conditions.#(type=="Failed")#|#(status=="True")#
    primaryContainerName: node
    primaryPodLabels:
      batch.kubernetes.io/job-completion-index: "0"
      jobset.sigs.k8s.io/replicatedjob-name: node
    retain: true
    successCondition: status.conditions.#(type=="Complete")#|#(status=="True")#
    trialParameters:
      - name: lr
        reference: lr
      - name: num_epochs
        reference: num_epochs
    trialSpec:
      apiVersion: trainer.kubeflow.org/v1alpha1
      kind: TrainJob
      spec:
        runtimeRef:
          name: torch-distributed
        trainer:
          command:
            - bash
            - -c
            - |2-

              read -r -d '' SCRIPT << EOM

              def train_func(lr: str, num_epochs: str):
                  import time
                  import random

                  for i in range(10):
                      time.sleep(1)
                      print(f"Training {i}, lr: {lr}, num_epochs: {num_epochs}")

                  print(f"loss={round(random.uniform(0.77, 0.99), 2)}")

              train_func(**{'lr': '${trialParameters.lr}', 'num_epochs': '${trialParameters.num_epochs}'})

              EOM
              printf "%s" "$SCRIPT" > "test-iceberg.py"
              torchrun "test-iceberg.py"
          numNodes: 2
status:
  completionTime: "2025-10-13T21:48:41Z"
  conditions:
    - lastTransitionTime: "2025-10-13T21:45:37Z"
      lastUpdateTime: "2025-10-13T21:45:37Z"
      message: Experiment is created
      reason: ExperimentCreated
      status: "True"
      type: Created
    - lastTransitionTime: "2025-10-13T21:48:41Z"
      lastUpdateTime: "2025-10-13T21:48:41Z"
      message: Experiment is running
      reason: ExperimentRunning
      status: "False"
      type: Running
    - lastTransitionTime: "2025-10-13T21:48:41Z"
      lastUpdateTime: "2025-10-13T21:48:41Z"
      message: Experiment has succeeded because max trial count has reached
      reason: ExperimentMaxTrialsReached
      status: "True"
      type: Succeeded
  currentOptimalTrial:
    bestTrialName: o759d77408d2-lfcqff79
    observation:
      metrics:
        - latest: "0.85"
          max: "0.98"
          min: "0.77"
          name: loss
    parameterAssignments:
      - name: lr
        value: "0.018571949792818013"
      - name: num_epochs
        value: "5"
  startTime: "2025-10-13T21:45:37Z"
  succeededTrialList:
    - o759d77408d2-lfcqff79
    - o759d77408d2-qwbkwc9n
    - o759d77408d2-jhqgmnm6
    - o759d77408d2-xjk86z66
    - o759d77408d2-g8mr72v7
    - o759d77408d2-5s2mqftm
    - o759d77408d2-86p9bw4r
    - o759d77408d2-28d5gd8f
    - o759d77408d2-m8gq4pcn
    - o759d77408d2-kxg6f45v
  trials: 10
  trialsSucceeded: 10
```
</details>

/assign @kubeflow/kubeflow-sdk-team @akshaychitneni 

TODO items:

- [x] Add `get_job()` API
- [x] Add `list_jobs()` API
- [x] Add `delete_job()` API

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -425,6 +425,7 @@ dependencies = [
 [package.dev-dependencies]
 dev = [
     { name = "coverage" },
+    { name = "kubeflow-katib-api" },
     { name = "kubeflow-trainer-api" },
     { name = "pre-commit" },
     { name = "pygithub" },
@@ -435,7 +436,7 @@ dev = [
 
 [package.metadata]
 requires-dist = [
-    { name = "kubeflow-katib-api", git = "https://github.com/kramaranya/katib.git?subdirectory=api%2Fpython_api&rev=separate-models-from-sdk" },
+    { name = "kubeflow-katib-api", specifier = ">=0.19.0" },
     { name = "kubeflow-trainer-api", specifier = ">=2.0.0" },
     { name = "kubernetes", specifier = ">=27.2.0" },
     { name = "pydantic", specifier = ">=2.10.0" },
@@ -444,6 +445,7 @@ requires-dist = [
 [package.metadata.requires-dev]
 dev = [
     { name = "coverage", specifier = ">=7.0" },
+    { name = "kubeflow-katib-api", git = "https://github.com/kubeflow/katib.git?subdirectory=api%2Fpython_api&rev=master" },
     { name = "kubeflow-trainer-api", git = "https://github.com/kubeflow/trainer.git?subdirectory=api%2Fpython_api&rev=master" },
     { name = "pre-commit", specifier = ">=4.2.0" },
     { name = "pygithub", specifier = ">=2.7.0" },
@@ -454,8 +456,8 @@ dev = [
 
 [[package]]
 name = "kubeflow-katib-api"
-version = "0.18.0"
-source = { git = "https://github.com/kramaranya/katib.git?subdirectory=api%2Fpython_api&rev=separate-models-from-sdk#97265fe9175114ef0293285e1a6e3deee7b727f4" }
+version = "0.19.0"
+source = { git = "https://github.com/kubeflow/katib.git?subdirectory=api%2Fpython_api&rev=master#38982e298d9c8d003d8964eaa067b71cb432fb7e" }
 dependencies = [
     { name = "pydantic" },
 ]


kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -0,0 +1,125 @@
...
+            raise ValueError(f"Invalid backend config '{backend_config}'")
+
+    def optimize(
+        self,
+        trial_template: common_types.TrainJobTemplate,
+        *,
+        trial_config: Optional[TrialConfig] = None,
+        search_space: dict[str, Any],
+        objectives: Optional[list[Objective]] = None,
+        algorithm: Optional[RandomSearch] = None,
```
astefanutti: Should we consider adding options already?

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,345 @@
...
+                name=name,
+            )
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(f"Timeout to delete OptimizationJob: {self.namespace}/{name}") from e
+        except Exception as e:
+            raise RuntimeError(f"Failed to delete OptimizationJob: {self.namespace}/{name}") from e
+
+        logger.debug(f"OptimizationJob {self.namespace}/{name} has been deleted")
+
+    def __get_optimization_job_from_crd(
```
astefanutti:
```suggestion
    def __get_optimization_job_from_custom_resource(
```

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,345 @@
...
+        self,
+        optimization_job_crd: models.V1beta1Experiment,
```
astefanutti:
```suggestion
        optimization_job_cr: models.V1beta1Experiment,
```

kubeflow/optimizer/backends/base.py
```diff
@@ -0,0 +1,46 @@
...
+
+import abc
+from typing import Any, Optional
+
+from kubeflow.common import types as common_types
+from kubeflow.optimizer.types.algorithm_types import RandomSearch
+from kubeflow.optimizer.types.optimization_types import Objective, OptimizationJob, TrialConfig
+
+
+class ExecutionBackend(abc.ABC):
```
astefanutti:
```suggestion
class RuntimeBackend(abc.ABC):
```

astefanutti: Or:
```suggestion
class OptimizerBackend(abc.ABC):
```

kubeflow/optimizer/api/optimizer_client.py
andreyvelich: Let's add it in the followup PR, since we want to limit number of APIs user can configure initially for Experiment CR.

kubeflow/optimizer/backends/base.py
andreyvelich: We previously agreed on the ExecutionBackend here: https://github.com/kubeflow/sdk/pull/34#discussion_r2299391985 with @kramaranya and @szaher.
Do you prefer to find better name for it @astefanutti ?

astefanutti: @andreyvelich that's not a big deal, ExecutionBackend is fine. RuntimeBackend seems more general as it also covers resources and not only the "execution", like the job "registry" (ETCD for Kubernetes).

andreyvelich: That sounds good!

kubeflow/optimizer/api/optimizer_client.py
astefanutti: Sounds good!

kubeflow/optimizer/constants/constants.py
```diff
@@ -0,0 +1,53 @@
...
+EXPERIMENT_KIND = "Experiment"
+
+# The plural for the Experiment.
+EXPERIMENT_PLURAL = "experiments"
+
+# The succeeded condition for the Experiment.
+EXPERIMENT_SUCCEEDED = "Succeeded"
+
+# Label to identify Experiment's resources.
+EXPERIMENT_LABEL = "katib.kubeflow.org/experiment"
```
astefanutti, line 31: Should we start using `optimizer.kubeflow.org`?

andreyvelich: Since we rely on Katib Experiment CRD for now, we can't use the new labels yet.

kubeflow/trainer/types/types.py
```diff
@@ -238,20 +239,19 @@ class Step:
...
-# TODO (andreyvelich): Discuss what fields users want to get.
 @dataclass
 class TrainJob:
     name: str
-    creation_timestamp: datetime
     runtime: Runtime
     steps: list[Step]
     num_nodes: int
-    status: str = constants.UNKNOWN
+    creation_timestamp: datetime
```
Electronic-Waste, line 253: Why do we need `creation_timestamp`? Shouldn't it be added automatically in the creation phase?

kubeflow/optimizer/constants/constants.py
Electronic-Waste: So, do we have plans to implement `OptimizerRuntime` and let `OptimizerJob` override it in the future?

andreyvelich: I don't think we need `OptimizerRuntime`, since `OptimizerJob` should natively integrate with `TrainingRuntime`

kubeflow/trainer/types/types.py
andreyvelich: It does. We set this property from the Experiment.metadata.creation_timestamp: https://github.com/kubeflow/sdk/blob/9b3700a07514f9d2b5e0bc4664cf202e47a11ee9/kubeflow/optimizer/backends/kubernetes/backend.py#L278

andreyvelich:
@kramaranya @Electronic-Waste @astefanutti Any additional comments before we move forward with the initial support of HPO in Kubeflow SDK ?

pyproject.toml
```diff
@@ -30,6 +29,8 @@ dependencies = [
   "kubernetes>=27.2.0",
   "pydantic>=2.10.0",
   "kubeflow-trainer-api>=2.0.0",
+  # TODO (andreyvelich): Switch to kubeflow-katib-api once it is published.
+  "kubeflow_katib_api@git+https://github.com/kramaranya/katib.git@separate-models-from-sdk#subdirectory=api/python_api",
```
kramaranya: Since it has been merged, we can update that with katib ref instead of the fork. Or shall we cut a new Katib release and publish those models to PyPI?

kubeflow/trainer/types/types.py
```diff
@@ -330,3 +330,27 @@ class Initializer:
...
+
+    Args:
+        trainer (`CustomTrainer`): Configuration for a CustomTrainer.
+        runtime (`Optional[Runtime]`): Optional, reference to one of the existing runtimes. Defaults
+            to the torch-distributed runtime if not provided.
+        initializer (`Optional[Initializer]`): Optional configuration for the dataset and model
+            initializers.
+    """
+
+    trainer: CustomTrainer
```
kramaranya, line 348: Why don't we support BuiltinTrainer initially? Is it due to metrics collection?

kubeflow/optimizer/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+
+# Import the Kubeflow Optimizer types.
+from kubeflow.optimizer.types.algorithm_types import RandomSearch
+from kubeflow.optimizer.types.optimization_types import Objective, OptimizationJob, TrialConfig
+from kubeflow.optimizer.types.search_types import Search
+
+# Import the Kubeflow Trainer types.
+from kubeflow.trainer.types.types import TrainJobTemplate
+
+__all__ = [
```
kramaranya, line 29: Shall we add GridSearch here?

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -0,0 +1,126 @@
...
+            backend_config: Backend configuration. Either KubernetesBackendConfig or None to use
+                default config class. Defaults to KubernetesBackendConfig.
+
+        Raises:
+            ValueError: Invalid backend configuration.
+
+        """
+        # Set the default backend config.
+        if not backend_config:
+            backend_config = KubernetesBackendConfig()
```
kramaranya, line 44: nit, just for consistency shall we match [trainer](https://github.com/andreyvelich/sdk/blob/hpo-support/kubeflow/trainer/api/trainer_client.py#L19) and use the same import style:
```python
if not backend_config:
    backend_config = common_types.KubernetesBackendConfig()
```

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,346 @@
...
+                name=name,
+            )
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(f"Timeout to delete OptimizationJob: {self.namespace}/{name}") from e
+        except Exception as e:
+            raise RuntimeError(f"Failed to delete OptimizationJob: {self.namespace}/{name}") from e
+
+        logger.debug(f"OptimizationJob {self.namespace}/{name} has been deleted")
+
+    def __get_optimization_job_from_custom_resource(
```
kramaranya: To align with trainer, should we update this?
```suggestion
    def __get_optimization_job_from_cr(
```

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,346 @@
...
+
+            if not optimization_job_list:
+                return result
+
+            for optimization_job in optimization_job_list.items:
+                result.append(self.__get_optimization_job_from_custom_resource(optimization_job))
+
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(
+                f"Timeout to list OptimizationJobs in namespace: {self.namespace}"
```
kramaranya: Can we add OptimizationJob to constants instead?

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -0,0 +1,346 @@
...
+        algorithm = algorithm or RandomSearch()
+        trial_config = trial_config or TrialConfig()
+
+        # Iterate over search space to build the following values:
+        # experiment.spec.parameters to define distribution and feasible space.
+        # experiment.spec.trialTemplate.trialParameters to reference parameters in Trials.
+        # Trainer function arguments for the appropriate substitution.
+        parameters_spec = []
+        trial_parameters = []
+        trial_template.trainer.func_args = {}
```
kramaranya: Would this not overwrite existing func_args?

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -0,0 +1,126 @@
...
+            raise ValueError(f"Invalid backend config '{backend_config}'")
+
+    def optimize(
+        self,
+        trial_template: TrainJobTemplate,
+        *,
+        trial_config: Optional[TrialConfig] = None,
+        search_space: dict[str, Any],
+        objectives: Optional[list[Objective]] = None,
+        algorithm: Optional[RandomSearch] = None,
```
kramaranya: I wonder whether we should accept a base type instead so any algorithm works without changing api in the future?
```suggestion
        algorithm: Optional[BaseAlgorithm] = None,
```

kubeflow/optimizer/types/optimization_types.py
```diff
@@ -0,0 +1,119 @@
...
+from kubeflow.optimizer.types.search_types import CategoricalSearchSpace, ContinuousSearchSpace
+from kubeflow.trainer.types.types import TrainJob
+
+
+# Direction for optimization objective
+class Direction(Enum):
+    """Direction for optimization objective."""
+
+    MAXIMIZE = "maximize"
+    MINIMIZE = "minimize"
```
kramaranya, line 31: What do you think about adding "max" and "min" aliases?

pyproject.toml
andreyvelich: Yes, I will replace it once we release 0.19

kubeflow/trainer/types/types.py
andreyvelich: Yes, since we don't have access to the BuiltinTrainer scripts, we can't guarantee in which format metrics will be printed. In the future integrations, let's talk how we can integrate BuiltinTrainers as well.

kubeflow/optimizer/__init__.py
andreyvelich: Good catch!

kubeflow/optimizer/api/optimizer_client.py
andreyvelich: Let me go other way around, tho.

kubeflow/optimizer/backends/kubernetes/backend.py
andreyvelich: It will, it is a good point.
I am curious whether we want to merge them? E.g. we should allow users to do something like this:
```py
func_args={
  "lr": 0.1
  "num_epochs": 5
},
search_space={
  "num_epochs": Serach.choice(1,2)
}
```
WDYT @kramaranya @astefanutti @Electronic-Waste ?

kubeflow/optimizer/api/optimizer_client.py
andreyvelich: Good point!

kubeflow/optimizer/types/optimization_types.py
andreyvelich: I think, this makes us consistent with the Optuna naming: https://optuna.readthedocs.io/en/stable/tutorial/20_recipes/002_multi_objective.html#run-multi-objective-optimization

kubeflow/trainer/types/types.py
kramaranya: I see, thanks for explaining

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: That sounds good to me

kubeflow/optimizer/types/optimization_types.py
kramaranya: Sure, sounds good to me. I was just wondering if it makes sense to support both "max" and "maximize" as options, since Ray Tune uses "max" for example https://docs.ray.io/en/latest/tune/api/doc/ray.tune.run.html

kubeflow/optimizer/backends/kubernetes/backend.py
andreyvelich: Change it to:
```py
  if trial_template.trainer.func_args is None:
    trial_template.trainer.func_args = {}
```

kramaranya:
I'm thinking whether we should introduce OWNERS files for sub-packages like optimizer and trainer? I think this was mentioned in the KEP. Doesn't need to be in this PR, but wanted to raise it since we now have two subpackages

andreyvelich:
Thanks everyone!
/approve

kubeflow/optimizer/types/optimization_types.py
andreyvelich: Sounds good, maybe we should do something like this for [this enum](https://github.com/kubeflow/sdk/blob/1353fc9e856fcd6df3f6a5a316dc68fdaa033ffa/kubeflow/optimizer/types/optimization_types.py#L27) ?
```py
    @classmethod
    def from_str(cls, value: str):
        value = value.lower()
        aliases = {
            "max": "maximize",
            "maximize": "maximize",
            "min": "minimize",
            "minimize": "minimize",
        }
        return cls(aliases[value])
```

@kramaranya Can you create an issue to track this improvement ?

Issue #116, opened 2025-10-03
Support creating TrainJob using image entry point


sutaakar:
### What you would like to be added?

I would like to have a simple way to create a `TrainJob` using entry point from image specified in TrainingRuntime, without a need to specify explicit train function and related properties (`func_args`, `packages_to_install`, `pip_index_urls`).
It should still be possible to provide job configuration like `num_nodes`, `resources_per_node`, `env`...

Optionally it can be also achieved by https://github.com/kubeflow/sdk/issues/47

### Why is this needed?

Some training images (for example https://github.com/foundation-model-stack/fms-hf-tuning/tree/main/build) already provide training scripts by themselves, referenced them in the image entry point.

It would be good to allow creation of `TrainJob` to run those images, providing possibility to specify training properties like environment variables, number of nodes, resources...

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

andreyvelich:
This is resolved by: https://github.com/kubeflow/sdk/pull/127
/close

PR #127, opened 2025-10-27
feat(trainer): Add CustomTrainerContainer to create TrainJobs from image

andreyvelich:
Fixes: https://github.com/kubeflow/sdk/issues/47, https://github.com/kubeflow/sdk/pull/49, https://github.com/kubeflow/sdk/issues/116, https://github.com/kubeflow/sdk/issues/26

As we discussed in the Slack, I introduced `CustomTrainerContainer()` to create TrainJobs using container image: https://cloud-native.slack.com/archives/C0742LDFZ4K/p1759849253406959

/assign @kubeflow/kubeflow-sdk-team @franciscojavierarceo @akshaychitneni @sutaakar @briangallagher @abhijeet-dhumal @Fiona-Waters @CrashCumber

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -112,8 +112,9 @@ class TrainerClient:
             runtime: Optional reference to one of the existing runtimes. Defaults to the
                 torch-distributed runtime if not provided.
             initializer: Optional configuration for the dataset and model initializers.
-            trainer: Optional configuration for a CustomTrainer or BuiltinTrainer. If not specified,
-                the TrainJob will use the runtime's default values.
+            trainer: Optional configuration for a CustomTrainer, CustomTrainerContainer, or
+                BuiltinTrainer. If not specified, the TrainJob will use the
+                runtime's default values.
 
         Returns:
             The unique name of the TrainJob that has been generated.


kubeflow/trainer/types/types.py
```diff
@@ -51,6 +51,25 @@ class CustomTrainer:
     env: Optional[dict[str, str]] = None
 
 
+# Configuration for the Custom Trainer Container.
+@dataclass
+class CustomTrainerContainer:
+    """Custom Trainer configuration. Configure the container image
```
kramaranya:
```suggestion
    """Custom Trainer Container configuration. Configure the container image
```

kramaranya:
> Looks fine in general. Just one question - does the image parameter need to be mandatory? Wouldn't it have sense to make it optional, taking default from training runtime?

Good question! But isn't the whole purpose of CustomTrainerContainer to let you bring your own image? I feel like making it optional might blur the distinction with CustomTrainer

cc @andreyvelich @astefanutti

andreyvelich:
> Looks fine in general. Just one question - does the image parameter need to be mandatory? Wouldn't it have sense to make it optional, taking default from training runtime?

@kramaranya Is right. It doesn't make sense to use CustomTrainer if user doesn't want to set an image or a training function. 

The main differences between CustomTrainer and BuiltinTrainer is that in BuiltinTrainer, the training script is pre-packaged for users, so they just need to configure input parameters.

The default runtime images for CustomTrainer or CustomTrainerContainer don't have any script to execute: https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/mlx_distributed.yaml#L34

andreyvelich:
@kramaranya Updated.

astefanutti:
Thanks @andreyvelich!

/lgtm
/approve
/hold @andreyvelich when you want it to go

PR #130, opened 2025-10-29
feat(trainer): Support namespaced TrainingRuntime in the SDK

### LINKED ISSUE #88: Support namespaced TrainingRuntime in the SDK

We want the SDK to work with both cluster-scoped and namespaced runtimes, not just `ClusterTrainingRuntime`. Users should be able to list and select a `TrainingRuntime` from their namespace.

### Why is this needed?

Right now, users can only reference `ClusterTrainingRuntime` for their TrainJob via the SDK. Users who need `TrainingRuntime` per team or per project in their own namespaces can’t use the SDK directly, they have to use YAMLs instead. 

### Love this feature?

shaikmoeed:
Add support to list/get namespaced TrainingRuntime.

Fixes #88 

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -184,21 +184,12 @@ class KubernetesBackend(RuntimeBackend):
 
         except multiprocessing.TimeoutError as e:
             raise TimeoutError(
-                f"Timeout while getting namespaced TrainingRuntime {self.namespace}/{name}"
+                f"Timeout to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
             ) from e
-
-        except client.exceptions.ApiException as e:
-            if e.status != 404:
-                raise RuntimeError(
-                    f"Failed to get namespaced TrainingRuntime "
-                    f"{self.namespace}/{name}: {e.status} {e.reason}"
-                ) from e
-
-            logger.info(
-                "Namespaced TrainingRuntime %s/%s not found, falling back to cluster-scoped",
-                self.namespace,
-                name,
-            )
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+            ) from e
 
         try:
             cluster_thread = self.custom_api.get_cluster_custom_object(
@@ -214,9 +205,13 @@ class KubernetesBackend(RuntimeBackend):
             return self.__get_runtime_from_cr(runtime)
 
         except multiprocessing.TimeoutError as e:
-            raise TimeoutError(f"Timeout while getting cluster TrainingRuntime '{name}'") from e
+            raise TimeoutError(
+                f"Timeout to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+            ) from e
         except Exception as e:
-            raise RuntimeError(f"Failed to get cluster TrainingRuntime: '{name}'") from e
+            raise RuntimeError(
+                f"Failed to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+            ) from e
 
     def get_runtime_packages(self, runtime: types.Runtime):
         if runtime.trainer.trainer_type == types.TrainerType.BUILTIN_TRAINER:


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -88,33 +100,55 @@ def list_runtimes(self) -> list[types.Runtime]:
...
+                "Failed to list "
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
                 f"in namespace: {self.namespace}"
             ) from e
 
         return result
 
     def get_runtime(self, name: str) -> types.Runtime:
-        """Get the the Runtime object"""
+        """Get the the Runtime object prefer namespaced, fall-back to cluster-scoped"""
```
abhijeet-dhumal:
```suggestion
        """Get the Runtime object prefer namespaced, fall-back to cluster-scoped"""
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -59,21 +59,33 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+            namespace_thread = self.custom_api.list_namespaced_custom_object(
+                constants.GROUP,
+                constants.VERSION,
+                self.namespace,
+                constants.TRAINING_RUNTIME_PLURAL,
+                async_req=True,
+            )
+
+            cluster_runtime_list = models.TrainerV1alpha1ClusterTrainingRuntimeList.from_dict(
+                cluster_thread.get(constants.DEFAULT_TIMEOUT)
```
abhijeet-dhumal:
```suggestion
                cluster_thread.get(common_constants.DEFAULT_TIMEOUT)
```

kubeflow/trainer/backends/kubernetes/backend.py
abhijeet-dhumal: Same change goes for each occurence here

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -490,6 +500,37 @@ def create_cluster_training_runtime(
...
+                metadata=models.IoK8sApimachineryPkgApisMetaV1ObjectMeta(
+                    name=name,
+                    namespace=namespace,
+                ),
+                spec=models.JobsetV1alpha2JobSetSpec(replicatedJobs=[get_replicated_job()]),
+            ),
+        ),
+    )
+
+
```
szaher, line 639: did you mean to create this in `kubernetes/backend_test.py`? 
this is not a test function and I believe it should be added to the `TrainerClient` and propagated to the different backends.

shaikmoeed: Similar to [create_train_job](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend_test.py#L446), I thought to use `create_training_runtime` to pass [here ](https://github.com/shaikmoeed/sdk/blob/fix/namespace-trainingruntime-list/kubeflow/trainer/backends/kubernetes/backend_test.py#L361-L369) instead of empty list. Please correct me, if this was not intended?

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Done! Reason for test case failures!

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -87,28 +99,56 @@ def list_runtimes(self) -> list[types.Runtime]:
...
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
+                f"in namespace: {self.namespace}"
+            ) from e
         except Exception as e:
-            raise RuntimeError(f"Failed to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
+            raise RuntimeError(
+                "Failed to list "
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
+                f"in namespace: {self.namespace}"
+            ) from e
```
kramaranya, line 192: What If only cluster runtimes exist (no TrainingRuntime CRD)? The entire list_runtimes will fail with RuntimeError, which we don't want, right?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -87,28 +99,56 @@ def list_runtimes(self) -> list[types.Runtime]:
                 result.append(self.__get_runtime_from_cr(runtime))
 
         except multiprocessing.TimeoutError as e:
-            raise TimeoutError(f"Timeout to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
+            raise TimeoutError(
+                "Timeout to list "
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
+                f"in namespace: {self.namespace}"
+            ) from e
```
kramaranya, line 188: What if one type times out and the other succeeds? I think we still should return partial results

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -87,28 +99,56 @@ def list_runtimes(self) -> list[types.Runtime]:
                 result.append(self.__get_runtime_from_cr(runtime))
 
         except multiprocessing.TimeoutError as e:
-            raise TimeoutError(f"Timeout to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
+            raise TimeoutError(
+                "Timeout to list "
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
+                f"in namespace: {self.namespace}"
```
kramaranya: nit: cluster scoped runtimes are not really in a namespace

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -87,28 +99,56 @@ def list_runtimes(self) -> list[types.Runtime]:
...
+            # Try namespaced runtime first, fall back to cluster-scoped one
+            try:
+                runtime = models.TrainerV1alpha1TrainingRuntime.from_dict(
+                    namespace_thread.get(common_constants.DEFAULT_TIMEOUT)  # type: ignore
+                )
+            except Exception as e:
+                logger.warning(
+                    f"Namespaced TrainingRuntime '{self.namespace}/{name}' not found "
+                    f"({type(e).__name__}: {e}); falling back to cluster-scoped runtime."
+                )
```
kramaranya: What if it was for example TimeoutError? We will still silently fallback to cluster scoped runtimes, right? I would suggest only treating not found / missing CRD as fall back.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -87,28 +99,56 @@ def list_runtimes(self) -> list[types.Runtime]:
...
-            raise RuntimeError(f"Failed to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
+            raise RuntimeError(
+                "Failed to list "
+                f"{constants.CLUSTER_TRAINING_RUNTIME_KIND}s/{constants.TRAINING_RUNTIME_KIND}s "
+                f"in namespace: {self.namespace}"
+            ) from e
 
         return result
 
     def get_runtime(self, name: str) -> types.Runtime:
```
kramaranya: Similar to list_runtimes(), either case will cause a full failure

kubeflow/trainer/constants/constants.py
```diff
@@ -26,6 +26,12 @@
 # The plural for the ClusterTrainingRuntime.
 CLUSTER_TRAINING_RUNTIME_PLURAL = "clustertrainingruntimes"
 
+# The Kind name for the TrainingRuntime.
+TRAINING_RUNTIME_KIND = "TrainingRuntime"
+
+# The plural for the ClusterTrainingRuntime.
```
kramaranya:
```suggestion
# The plural for the TrainingRuntime.
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -59,21 +59,33 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
-            if not runtime_list:
+            namespace_runtime_list = models.TrainerV1alpha1TrainingRuntimeList.from_dict(
+                namespace_thread.get(common_constants.DEFAULT_TIMEOUT)
+            )
+
+            if not (cluster_runtime_list or namespace_runtime_list):
                 return result
 
-            for runtime in runtime_list.items:
+            for runtime in namespace_runtime_list.items + cluster_runtime_list.items:
```
abhijeet-dhumal: @shaikmoeed  
Quick question : What if runtimes with the same name exists in both cluster and namespace scoped ?
IIUC you have implemented namespace scoped priority in get_runtime() where trainingRuntimes get's first priority..

thinking should it be same case for list_runtimes method too ? 
And one more thing, In case of list_runtimes it's just appending both even if duplicates comes in.. 
So for end user how user will be able to know the kind of runtime via list_runtimes's list items ?

abhijeet-dhumal: What if we introduce kind and namespace(optional) params in Runtime dataclass here : https://github.com/kubeflow/sdk/blob/49a5087bb1154c0505cbe2a004db17d4d373c82a/kubeflow/trainer/types/types.py#L251
So that 
```
for runtime in runtimes:
    print(f"{runtime.name} ({runtime.kind}, ns={runtime.namespace})")
    
# Output:
# torch-runtime (TrainingRuntime, ns=team-a)
# torch-runtime (ClusterTrainingRuntime, ns=None)
# custom-runtime (TrainingRuntime, ns=team-a)
```

abhijeet-dhumal: WDYT? @kramaranya @szaher ?

kramaranya: I'd be in favor of adding a scope field instead, with values like "project" and "cluster", so we don't expose any k8s terminology to users.
Regarding namespace I don't think AI practitioners need to see it, since they already know they're working in their project context (set at client init time).
cc @andreyvelich @astefanutti

astefanutti: +1 for a scope field, though I wonder whether it could be more symmetrical to the resolution logic in the `get_runtime` and not returning the ClusterTrainingRuntimes that are "overwritten" in the client namespace. This way I don't think any additional field would be needed.

andreyvelich: Shall we configure this setting during `TrainerClient()` initialization?
I am curious if we ever have a use-case when users want to create a single `TrainerClient()` which can work with `TrainingRuntime` and `ClusterTrainingRuntime` ?

shaikmoeed:
> Hey, @shaikmoeed! Are you still working on this PR? Did you have a chance to address the feedback?

@kramaranya Haven't got a chance to address the feedback. Will update the PR by next week.

shaikmoeed:
> Hey @shaikmoeed! Are you still working on this PR? It would be great if we could include this feature in 0.3 release (we aim to cut the release this week)

@kramaranya  Sorry for the delay, I have addressed all the comments! Can you please provide your review?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -412,8 +492,17 @@ def delete_job(self, name: str):
 
     def __get_runtime_from_cr(
         self,
-        runtime_cr: models.TrainerV1alpha1ClusterTrainingRuntime,
+        runtime_cr: Union[
+            models.TrainerV1alpha1ClusterTrainingRuntime, models.TrainerV1alpha1TrainingRuntime
+        ],
     ) -> types.Runtime:
+        crd_kind = getattr(runtime_cr, "kind", "UnknownKind")
```
abhijeet-dhumal: QQ: It seems relying on kind field being present. K8s API doesn't always return kind on retrieved objects (it's often omitted). If missing, scope stays None.
what if we use type check which doesn't rely on K8s api response fileds.. something like :.
```
scope = (
    "cluster"
    if isinstance(runtime_cr, models.TrainerV1alpha1ClusterTrainingRuntime)
    else "project"
)
```

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -546,9 +593,13 @@ def create_runtime_type(
         image="example.com/test-runtime",
     )
     trainer.set_command(constants.TORCH_COMMAND)
+    # Namespaced TrainingRuntime objects and default torch runtime use project scope;
+    # other runtimes created as cluster-scoped use cluster scope.
+    scope = "project" if name.startswith("ns-") or name == TORCH_RUNTIME else "cluster"
```
abhijeet-dhumal: Thinking .. Scope is derived from naming convention. Adding a runtime like "my-custom-runtime" silently gets scope="cluster" even if it's namespaced.

Consider suggestion provided by @kramaranya from here : 
https://github.com/kubeflow/sdk/pull/130#discussion_r2583080641, Something like : 
```
create_runtime_type(name="my-custom-runtime", scope="project")
```
In this case explicit is better than implicit

abhijeet-dhumal: wdyt @shaikmoeed ?

abhijeet-dhumal:
Following up on @astefanutti's [comment](https://github.com/kubeflow/sdk/pull/130#discussion_r2584044912) about symmetry between list_runtimes() and get_runtime().

According to current behavior and if we consider separate scope arg : 
If Namespace has "torch-runtime", cluster also has "torch-runtime"
 client.list_runtimes() will return : 
```
- torch-runtime (scope=project)
- torch-runtime (scope=cluster)  ...duplicate name
```
client.get_runtime("torch-runtime")
Returns one:
```
- torch-runtime (scope=project)  .. namespaced wins
```
so user sees duplicate names in list, but get_runtime() silently picks one. 
This creates confusion — "which one will user actually get?"

so what if : 

client.list_runtimes()
Returns:
```
- torch-runtime (scope=project)  ... namespaced, shadows cluster
```
client.get_runtime("torch-runtime")
Returns:
```
- torch-runtime (scope=project)  ... matches what list shows
```

here are some edge cases i could think of on top of my head : 
- Same name in both scopes  -  Namespaced returned, cluster hidden 
- Only cluster runtime exists - Returned with scope="cluster"
- Only namespaced runtime exists - Returned with scope="project" 
- User needs shadowed cluster runtime - Use kubectl or rename .. rare admin case

WDYT? @shaikmoeed

kubeflow/trainer/backends/kubernetes/backend_test.py
shaikmoeed: Yeah, agree. Made the relevant changes!

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Thanks, it simplifies code as well! Updated with changes!

kubeflow/trainer/types/types.py
```diff
@@ -245,6 +245,13 @@ class TrainerType(Enum):
     BUILTIN_TRAINER = BuiltinTrainer.__name__
 
 
+class RuntimeScope(Enum):
+    """Scope for training runtimes."""
+
+    NAMESPACE = "project"
+    CLUSTER = "cluster"
```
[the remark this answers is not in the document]
shaikmoeed: This is to stay inline with domain usage.
cc: @kramaranya

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,161 @@ def verify_backend(self) -> None:
...
+        """List available runtimes, preferring namespaced over cluster-scoped for duplicates.
+
+        If a TrainingRuntime with the same name exists in both the namespace and cluster scope,
+        only the namespaced runtime is returned. Cluster-scoped runtimes are still returned
+        when there is no namespaced runtime with the same name.
+        """
+        result: list[types.Runtime] = []
+
+        cluster_err = None
+        namespace_err = None
```
andreyvelich: Can you return errors directly like we do in other places.

```suggestion
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,161 @@ def verify_backend(self) -> None:
...
-                return result
+        if cluster_runtime_list is None and namespace_runtime_list is None:
+            # If both timed out → timeout
+            if isinstance(cluster_err, multiprocessing.TimeoutError) and isinstance(
+                namespace_err, multiprocessing.TimeoutError
+            ):
+                raise TimeoutError(
+                    f"Timeout while retrieving both cluster and namespace runtimes "
+                    f"(namespace={self.namespace})"
+                ) from cluster_err
```
andreyvelich: Can you refactor this to simple raise timeout error like we do right now: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L117-L121

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,161 @@ def verify_backend(self) -> None:
...
+        runtimes = []
+        ns_names = {}
+        if namespace_runtime_list:
+            runtimes.extend(namespace_runtime_list.items)
+            ns_names = {r.metadata.name for r in runtimes}
+
+        if cluster_runtime_list:
+            runtimes.extend(
+                r for r in cluster_runtime_list.items if r.metadata.name not in ns_names
+            )
```
andreyvelich: It would be easier to create set of runtimes, and if item is already in the set (e.g. namespace runtime), we don't add cluster runtime there.

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Could get namespace runtimes time out while get cluster runtimes does not?

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Does the order of runtimes really matter? I don’t think so — I agree with you, I’ll use a set.

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: I don't think it matters for users, just run `list_namespaced_custom_object` and `list_cluster_custom_object`, and raise Exception if any of it fails.

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Do we need to make `Runtime` as hashable object to store in set?

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: Updated as per the suggestion

kubeflow/trainer/backends/kubernetes/backend.py
shaikmoeed: I have made some changes. LMK, if this can be improved a bit more?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,143 @@ def verify_backend(self) -> None:
...
+            async_req=True,
+        )
+
+        namespace_thread = self.custom_api.list_namespaced_custom_object(
+            constants.GROUP,
+            constants.VERSION,
+            self.namespace,
+            constants.TRAINING_RUNTIME_PLURAL,
+            async_req=True,
+        )
```
andreyvelich, line 109: You can do something like this to de-duplicate code and keep runtime names in the Map:
```py
      # Fetch both cluster and namespaced runtimes in parallel
        cluster_thread = self.custom_api.list_cluster_custom_object(
            constants.GROUP,
            constants.VERSION,
            constants.CLUSTER_TRAINING_RUNTIME_PLURAL,
            async_req=True,
        )
        namespace_thread = self.custom_api.list_namespaced_custom_object(
            constants.GROUP,
            constants.VERSION,
            self.namespace,
            constants.TRAINING_RUNTIME_PLURAL,
            async_req=True,
        )

        # Helper to fetch and convert runtime list
        def fetch_runtime_list(thread, kind, model_class):
            try:
                return model_class.from_dict(thread.get(common_constants.DEFAULT_TIMEOUT))
            except multiprocessing.TimeoutError as e:
                raise TimeoutError(f"Timeout to list {kind}s") from e
            except Exception as e:
                raise RuntimeError(f"Failed to list {kind}s") from e

        cluster_runtimes = fetch_runtime_list(
            cluster_thread,
            constants.CLUSTER_TRAINING_RUNTIME_KIND,
            models.TrainerV1alpha1ClusterTrainingRuntimeList,
        )
        namespace_runtimes = fetch_runtime_list(
            namespace_thread,
            constants.TRAINING_RUNTIME_KIND,
            models.TrainerV1alpha1TrainingRuntimeList,
        )

        # Collect runtimes in a map, preferring namespaced over cluster-scoped
        runtimes_by_name = {}

        # Add namespaced runtimes first (they have priority)
        if namespace_runtimes:
            for runtime in namespace_runtimes.items:
                if runtime.metadata and runtime.metadata.name:
                    runtimes_by_name[runtime.metadata.name] = runtime

        # Add cluster runtimes only if not already present
        if cluster_runtimes:
            for runtime in cluster_runtimes.items:
                if runtime.metadata and runtime.metadata.name:
                    runtimes_by_name.setdefault(runtime.metadata.name, runtime)


```

kubeflow/trainer/types/types.py
```diff
@@ -245,6 +245,13 @@ class TrainerType(Enum):
     BUILTIN_TRAINER = BuiltinTrainer.__name__
 
 
+class RuntimeScope(Enum):
```
andreyvelich: I don't think that is needed for now.
As namespace runtime takes priority over cluster-scope runtime, I would suggest we don't show user the scope of the runtime.
In the future, we might want to introduce `scope` API to the `get_runtime()` and `list_runtimes()` APIs if users want it.

shaikmoeed: Makes sense for UX. Only question: does hiding scope reduce observability when precedence kicks in? If yes, we might want a future debug hook.

cc: @andreyvelich @kramaranya @abhijeet-dhumal for your thoughts on the same.

andreyvelich: I would suggest to leave it as future improvement. So we can remove RuntimeScope from this PR.

shaikmoeed: Noted! Removed RuntimeScope, can you please provide review again? Thank you!

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
+                    f"{self.namespace}/{name}: {e.status} {e.reason}"
+                ) from e
 
-    def get_runtime(self, name: str) -> types.Runtime:
-        """Get the Runtime object"""
+            logger.info(
+                "Namespaced TrainingRuntime %s/%s not found, falling back to cluster-scoped",
+                self.namespace,
+                name,
+            )
```
andreyvelich: Let's not give this `info` log.

```suggestion
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
-                thread.get(common_constants.DEFAULT_TIMEOUT)  # type: ignore
+                cluster_thread.get(common_constants.DEFAULT_TIMEOUT)
             )
+            return self.__get_runtime_from_cr(runtime)
 
         except multiprocessing.TimeoutError as e:
-            raise TimeoutError(
-                f"Timeout to get {constants.CLUSTER_TRAINING_RUNTIME_PLURAL}: {name}"
-            ) from e
+            raise TimeoutError(f"Timeout while getting cluster TrainingRuntime '{name}'") from e
```
andreyvelich:
```suggestion
            raise TimeoutError(f"Timeout to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}") from e
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
         except Exception as e:
-            raise RuntimeError(
-                f"Failed to get {constants.CLUSTER_TRAINING_RUNTIME_PLURAL}: {name}"
-            ) from e
-
-        return self.__get_runtime_from_cr(runtime)  # type: ignore
+            raise RuntimeError(f"Failed to get cluster TrainingRuntime: '{name}'") from e
```
andreyvelich:
```suggestion
            raise RuntimeError(f"Failed to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}") from e
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
+            raise TimeoutError(
+                f"Timeout while getting namespaced TrainingRuntime {self.namespace}/{name}"
+            ) from e
 
-        return result
+        except client.exceptions.ApiException as e:
+            if e.status != 404:
+                raise RuntimeError(
+                    f"Failed to get namespaced TrainingRuntime "
+                    f"{self.namespace}/{name}: {e.status} {e.reason}"
```
andreyvelich:
```suggestion
                    f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}" from e
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
+                ns_thread.get(common_constants.DEFAULT_TIMEOUT)
+            )
+            return self.__get_runtime_from_cr(runtime)
 
         except multiprocessing.TimeoutError as e:
-            raise TimeoutError(f"Timeout to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
-        except Exception as e:
-            raise RuntimeError(f"Failed to list {constants.CLUSTER_TRAINING_RUNTIME_KIND}s") from e
+            raise TimeoutError(
+                f"Timeout while getting namespaced TrainingRuntime {self.namespace}/{name}"
```
andreyvelich:
```suggestion
                f"Timeout to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -85,68 +85,138 @@ def verify_backend(self) -> None:
...
+            ) from e
 
-        return result
+        except client.exceptions.ApiException as e:
+            if e.status != 404:
```
andreyvelich: Can you process this error similar to get_job() API: 
https://github.com/shaikmoeed/sdk/blob/a42180ca15fe7a8075a5da6566c41ce41b5c649b/kubeflow/trainer/backends/kubernetes/backend.py#L415-L423
```suggestion
```

PR #136, opened 2025-10-30
fix: add --user when install python packages

### LINKED ISSUE #135: Install python packages with --user to ensure success when container not running as root

Always install additional python packages defined in `packages_to_install` with the `--user `argument. 

### Why is this needed?

When installing the packages listed in `packages_to_install` they are currently installed system wide with `pip install <packages>. `
When the training container is running as non root and the HOME location is not accessible to this user, the install fails. 

By always installing with --user we will ensure the install will not fail for write permissions reasons. Users can use this in conjunction with setting a HOME environment variable. 

See [issues/84](https://github.com/kubeflow/sdk/issues/84) for more details.

### Love this feature?

briangallagher:
Add --user argument when installing additional python packages to ensure avoid permissions errors when the container is running as non root

Fixes #135

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -247,9 +247,8 @@ def get_script_for_python_packages(
     # first url will be the index-url.
     options = [f"--index-url {pip_index_urls[0]}"]
     options.extend(f"--extra-index-url {extra_index_url}" for extra_index_url in pip_index_urls[1:])
-    # For the OpenMPI, the packages must be installed for the mpiuser.
-    if is_mpi:
-        options.append("--user")
+    # Always install for the user to avoid permission issues across environments.
+    options.append("--user")
 
     header_script = textwrap.dedent(
         """
--- a/kubeflow/trainer/backends/localprocess/constants.py
+++ b/kubeflow/trainer/backends/localprocess/constants.py
@@ -43,7 +43,7 @@ local_runtimes = [
 # The exec script to embed training function into container command.
 DEPENDENCIES_SCRIPT = textwrap.dedent(
     """
-        PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET \
+        PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET --user \
     --no-warn-script-location $PIP_INDEX $PACKAGE_STR
     """
 )
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Kubeflow Trainer client supports local development without needing a Kubernetes 
 ### Available Backends
 
 - **KubernetesBackend** (default) - Production training on Kubernetes
-- **ContainerBackend** - Local development with Docker/Podman isolation  
+- **ContainerBackend** - Local development with Docker/Podman isolation
 - **LocalProcessBackend** - Quick prototyping with Python subprocesses
 
 **Quick Start:**
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -217,6 +217,7 @@ def get_custom_trainer(
     """
     pip_command = [f"--index-url {pip_index_urls[0]}"]
     pip_command.extend([f"--extra-index-url {repo}" for repo in pip_index_urls[1:]])
+    pip_command.append("--user")
     pip_command = " ".join(pip_command)
 
     packages_command = " ".join(packages_to_install)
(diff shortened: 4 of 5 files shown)


astefanutti:
@briangallagher Thanks!

/lgtm

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -269,9 +269,8 @@ def get_script_for_python_packages(
     # first url will be the index-url.
     options = [f"--index-url {pip_index_urls[0]}"]
     options.extend(f"--extra-index-url {extra_index_url}" for extra_index_url in pip_index_urls[1:])
-    # For the OpenMPI, the packages must be installed for the mpiuser.
-    if is_mpi:
-        options.append("--user")
+    # Always install for the user to avoid permission issues across environments.
+    options.append("--user")
```
andreyvelich, line 251: This can be removed since you always add --user to the constants, right ?
```suggestion
```
https://github.com/kubeflow/sdk/blob/b74f60e12f5841a4f099c5af547dc234fe73ea73/kubeflow/trainer/backends/localprocess/constants.py#L45

andreyvelich (commented):
@briangallagher Thanks!

kubeflow/trainer/backends/kubernetes/utils.py
briangallagher: That constant is just used for local process so `options.append("--user")`  is still required.

andreyvelich: Oh, you are right, we don't use it for k8s.
@briangallagher please rebase this PR, so we can merge it.

kubeflow/trainer/backends/localprocess/constants.py
```diff
@@ -42,7 +42,7 @@
 # The exec script to embed training function into container command.
 DEPENDENCIES_SCRIPT = textwrap.dedent(
     """
-        PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET \
+        PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET --user \
```
kramaranya, line 46: Do we actually need this for local execution mode?

andreyvelich: @szaher maybe knows more.

szaher: for local-exec we need to only pass the mappings and it should work https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/localprocess/utils.py#L171

```
def get_dependencies_command(
    runtime_packages: list[str],
    pip_index_urls: str,
    trainer_packages: list[str],
    quiet: bool = True,
) -> str:
    # resolve runtime dependencies and trainer dependencies.
    packages = get_install_packages(
        runtime_packages=runtime_packages,
        trainer_packages=trainer_packages,
    )

    """
           PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET $AS_USER \
       --no-warn-script-location $PIP_INDEX $PACKAGE_STR
       """
    mapping = {
        "QUIET": "--quiet" if quiet else "",
        "PIP_INDEX": " ".join(options),
        "PACKAGE_STR": '"{}"'.format('" "'.join(packages)),  # quote deps
        "AS_USER": "--user"
    }
    t = Template(local_exec_constants.DEPENDENCIES_SCRIPT)
    result = t.substitute(**mapping)
    return result
```

or 

```
def get_dependencies_command(
    runtime_packages: list[str],
    pip_index_urls: str,
    trainer_packages: list[str],
    quiet: bool = True,
) -> str:
    # resolve runtime dependencies and trainer dependencies.
    packages = get_install_packages(
        runtime_packages=runtime_packages,
        trainer_packages=trainer_packages,
    )

    """
           PIP_DISABLE_PIP_VERSION_CHECK=1 pip install $QUIET --user\
       --no-warn-script-location $PIP_INDEX $PACKAGE_STR
       """
    mapping = {
        "QUIET": "--quiet" if quiet else "",
        "PIP_INDEX": " ".join(options),
        "PACKAGE_STR": '"{}"'.format('" "'.join(packages)),  # quote deps
    }
    t = Template(local_exec_constants.DEPENDENCIES_SCRIPT)
    result = t.substitute(**mapping)
    return result
```

briangallagher: @szaher So I think your saying yes, we can leave it? I think it's good to be consistent with the kubernetes backend approach, which uses `--user` by default. Your second approach needs to update `local_exec_constants.DEPENDENCIES_SCRIPT`, which is what the PR currently does.

andreyvelich (commented):
@briangallagher Please can you fix the pre-commit, so we can merge this PR ?

kubeflow/trainer/backends/localprocess/constants.py
andreyvelich: I think, it's fine to use `--user` for local backend for now.

PR #140, opened 2025-11-04
fix: Support custom images in ClusterTrainingRuntime for container backend

Fiona-Waters:
While testing use of a custom image in the container backend I noticed that the image was not being used and instead the default one was being picked up. This PR will fix this and ensure that the TrainingRuntimeSource will correctly use the image specified in the users ClusterTrainingRuntime yaml.

cc @andreyvelich @astefanutti @kramaranya 

Fixes #

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/container/runtime_loader.py
+++ b/kubeflow/trainer/backends/container/runtime_loader.py
@@ -329,6 +329,7 @@ def _create_default_runtimes() -> list[base_types.Runtime]:
                 num_nodes=1,
             ),
             pretrained_model=None,
+            image=image,
         )
         default_runtimes.append(runtime)
         logger.debug(f"Created default runtime: {runtime.name} with image {image}")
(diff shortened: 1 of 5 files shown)


szaher (approved):
/lgtm

kubeflow/trainer/backends/container/runtime_loader.py
```diff
@@ -388,6 +389,9 @@ def _parse_runtime_yaml(data: dict[str, Any], source: str = "unknown") -> base_t
     if not containers or not containers[0].get("image"):
         raise ValueError(f"Runtime {name} from {source} 'node' must specify containers[0].image")
 
+    # Extract the container image
+    image = containers[0].get("image")
```
astefanutti: Might not be that important in practice, but maybe more robust to select the container named "node".

Fiona-Waters: Thanks, I've updated to select the container named "node". PTAL.

kubeflow/trainer/types/types.py
```diff
@@ -250,6 +250,7 @@ class Runtime:
     name: str
     trainer: RuntimeTrainer
     pretrained_model: Optional[str] = None
+    image: Optional[str] = None
```
andreyvelich, line 253: @astefanutti @Fiona-Waters, shall we move this image under `RuntimeTrainer` ?
Since we also have initializer in the Runtime.
Also, we might need to update Kubernetes backend to also populate this field.

astefanutti: @andreyvelich right that makes sense.

Fiona-Waters: Sure, I can look at creating a follow on PR to do that.

Fiona-Waters: I see you already got there :-)

PR #143, opened 2025-11-04
fix(trainer): Fix empty image for Runtime trainer

andreyvelich:
As we discussed here, we should move `image` to the Runtime trainer: https://github.com/kubeflow/sdk/pull/140#discussion_r2491185418

/assign @kubeflow/kubeflow-sdk-team @Fiona-Waters

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -21,6 +21,7 @@ from unittest.mock import Mock, patch
 import pytest
 
 from kubeflow.trainer.backends.localprocess.backend import LocalProcessBackend
+from kubeflow.trainer.backends.localprocess.constants import LOCAL_RUNTIME_IMAGE
 from kubeflow.trainer.backends.localprocess.types import (
     LocalProcessBackendConfig,
     LocalRuntimeTrainer,
@@ -78,9 +79,10 @@ def mock_train_environment():
             trainer_type=types.TrainerType.CUSTOM_TRAINER,
             framework="torch",
             num_nodes=1,
-            device_count=1,
+            device_count="1",
             device="cpu",
             packages=["torch"],
+            image=LOCAL_RUNTIME_IMAGE,
         )
         mock_trainer.set_command = Mock()
         mock_get_trainer.return_value = mock_trainer
@@ -152,6 +154,7 @@ def test_get_runtime(local_backend, test_case):
                         trainer_type=types.TrainerType.CUSTOM_TRAINER,
                         framework="torch",
                         num_nodes=1,
+                        image=LOCAL_RUNTIME_IMAGE,
                     ),
                 ),
             },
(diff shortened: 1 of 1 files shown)


kubeflow/trainer/backends/container/utils.py
```diff
@@ -150,38 +150,6 @@ def aggregate_status_from_containers(container_statuses: list[str]) -> str:
     return UNKNOWN
 
 
-def resolve_image(runtime: types.Runtime) -> str:
```
andreyvelich, line 153: @Fiona-Waters It looks like we don't need image resolver, since we fallback to the default runtime in case we can't get it online: https://github.com/kubeflow/sdk/blob/a9c5244abe6299c8b26461b10b70546283beba79/kubeflow/trainer/backends/container/runtime_loader.py#L314

Is that correct ?

kubeflow/trainer/types/types.py
```diff
@@ -231,6 +231,7 @@ class TrainerType(Enum):
 class RuntimeTrainer:
     trainer_type: TrainerType
     framework: str
+    image: str
```
andreyvelich, line 235: @astefanutti @Fiona-Waters I made `image` mandatory for the RuntimeTrainer.
The container should always has an image, but for the local subprocess backend, we can populate some const value there.
What do you think about it ?

kubeflow/trainer/backends/container/utils.py
Fiona-Waters: Yes if image is not optional then we don't need this function.

kubeflow/trainer/types/types.py
Fiona-Waters: I think it makes sense for the image to be mandatory. Adding a dummy const to local process backend seems fine to me as long as we make sure it's purpose is mentioned clearly in a comment.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -220,14 +219,17 @@ def get_runtime_packages(self, runtime: types.Runtime):
...
 
     def train(
         self,
         runtime: Optional[types.Runtime] = None,
         initializer: Optional[types.Initializer] = None,
-        trainer: Optional[Union[types.CustomTrainer, types.BuiltinTrainer]] = None,
+        trainer: Optional[
+            Union[types.CustomTrainer, types.CustomTrainerContainer, types.BuiltinTrainer]
+        ] = None,
+        options: Optional[list] = None,
```
andreyvelich, line 230: This should fix E2Es in this PR: https://github.com/kubeflow/trainer/pull/2907

cc @Fiona-Waters @astefanutti @kramaranya

kubeflow/trainer/backends/base.py
```diff
@@ -62,7 +62,7 @@ def get_job(self, name: str) -> types.TrainJob:
     def get_job_logs(
         self,
         name: str,
-        follow: Optional[bool] = False,
+        follow: bool = False,
```
kramaranya, line 65: Why do we need this change?

andreyvelich: Properties should not be Optional if they can't be `None`.
By default `follow=False`

PR #145, opened 2025-11-05
feat(optimizer): Add wait_for_job_status and get_best_trial APIs to OptimizerClient

kramaranya:
I've added `wait_for_job_status` and `get_best_trial` APIs to the OptimizerClient. 

Working example:

```python
from kubeflow.optimizer import OptimizerClient, Search, Objective, TrialConfig, TrainJobTemplate
from kubeflow.trainer import TrainerClient, CustomTrainer

def train_model(learning_rate, num_epochs):
    import os
    import torch
    import torch.distributed as dist
    
    lr = float(learning_rate)
    epochs = int(num_epochs)
    
    dist.init_process_group(backend="gloo")
    print("PyTorch Distributed Environment")
    print(f"WORLD_SIZE: {dist.get_world_size()}")
    print(f"RANK: {dist.get_rank()}")
    print(f"LOCAL_RANK: {os.environ['LOCAL_RANK']}")
    
    if dist.get_rank() == 0:
        print(f"Training with lr={lr}, epochs={epochs}")
    
    accuracy = 0.8 + (lr * 10) + (epochs * 0.01)
    
    if dist.get_rank() == 0:
        print(f"accuracy={accuracy}")

template = TrainJobTemplate(
    trainer=CustomTrainer(
        func=train_model,
        func_args={"learning_rate": "0.01", "num_epochs": "5"},
        resources_per_node={"cpu": 1},
    ),
    runtime=TrainerClient().get_runtime("torch-distributed"),
)

job_id = OptimizerClient().optimize(
    trial_template=template,
    trial_config=TrialConfig(num_trials=10, parallel_trials=2),
    search_space={
        "learning_rate": Search.loguniform(0.001, 0.1),
        "num_epochs": Search.choice([5, 10, 15]),
    },
    objectives=[Objective(metric="accuracy", direction="maximize")],
)

print(f"OptimizationJob created: {job_id}")

OptimizerClient().wait_for_job_status(job_id)

OptimizerClient().get_best_trial(job_id)
```

### DIFF
--- a/kubeflow/optimizer/backends/kubernetes/backend.py
+++ b/kubeflow/optimizer/backends/kubernetes/backend.py
@@ -209,31 +209,8 @@ class KubernetesBackend(RuntimeBackend):
 
     def get_job(self, name: str) -> OptimizationJob:
         """Get the OptimizationJob object"""
-
-        try:
-            thread = self.custom_api.get_namespaced_custom_object(
-                constants.GROUP,
-                constants.VERSION,
-                self.namespace,
-                constants.EXPERIMENT_PLURAL,
-                name,
-                async_req=True,
-            )
-
-            optimization_job = models.V1beta1Experiment.from_dict(
-                thread.get(common_constants.DEFAULT_TIMEOUT)  # type: ignore
-            )
-
-        except multiprocessing.TimeoutError as e:
-            raise TimeoutError(
-                f"Timeout to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
-            ) from e
-        except Exception as e:
-            raise RuntimeError(
-                f"Failed to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
-            ) from e
-
-        return self.__get_optimization_job_from_cr(optimization_job)  # type: ignore
+        optimization_job = self.__get_experiment_cr(name)
+        return self.__get_optimization_job_from_cr(optimization_job)
 
     def wait_for_job_status(
         self,
@@ -281,28 +258,7 @@ class KubernetesBackend(RuntimeBackend):
 
     def get_best_trial(self, name: str) -> Optional[Trial]:
         """Get the best current Trial for the OptimizationJob"""
-        try:
-            thread = self.custom_api.get_namespaced_custom_object(
-                constants.GROUP,
-                constants.VERSION,
-                self.namespace,
-                constants.EXPERIMENT_PLURAL,
-                name,
-                async_req=True,
-            )
-
-            optimization_job = models.V1beta1Experiment.from_dict(
-                thread.get(common_constants.DEFAULT_TIMEOUT)  # type: ignore
-            )
-
-        except multiprocessing.TimeoutError as e:
-            raise TimeoutError(
-                f"Timeout to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
-            ) from e
-        except Exception as e:
-            raise RuntimeError(
-                f"Failed to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
-            ) from e
+        optimization_job = self.__get_experiment_cr(name)
 
         # Get the best trial from currentOptimalTrial
         if (
@@ -366,6 +322,33 @@ class KubernetesBackend(RuntimeBackend):
             ) from e
 
         logger.debug(f"{constants.OPTIMIZATION_JOB_KIND} {self.namespace}/{name} has been deleted")
+
+    def __get_experiment_cr(self, name: str) -> models.V1beta1Experiment:
+        """Get the Experiment CR from Kubernetes API"""
+        try:
+            thread = self.custom_api.get_namespaced_custom_object(
+                constants.GROUP,
+                constants.VERSION,
+                self.namespace,
+                constants.EXPERIMENT_PLURAL,
+                name,
+                async_req=True,
+            )
+
+            optimization_job = models.V1beta1Experiment.from_dict(
+                thread.get(common_constants.DEFAULT_TIMEOUT)  # type: ignore
+            )
+
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(
+                f"Timeout to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
+            ) from e
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to get {constants.OPTIMIZATION_JOB_KIND}: {self.namespace}/{name}"
+            ) from e
+
+        return optimization_job
 
     def __get_optimization_job_from_cr(
         self,


kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -206,6 +207,116 @@ def list_jobs(self) -> list[OptimizationJob]:
 
         return result
 
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: set[str] = {constants.OPTIMIZATION_JOB_COMPLETE},
+        timeout: int = 3600,
+        polling_interval: int = 2,
```
astefanutti, line 220: Out of curiosity, I wonder whether there is a reason why it does not rely on a watch request?

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -206,6 +207,116 @@ def list_jobs(self) -> list[OptimizationJob]:
...
+    def get_best_trial(self, name: str) -> Optional[Trial]:
+        """Get the best current Trial for the OptimizationJob"""
+        try:
+            thread = self.custom_api.get_namespaced_custom_object(
+                constants.GROUP,
+                constants.VERSION,
+                self.namespace,
+                constants.EXPERIMENT_PLURAL,
+                name,
+                async_req=True,
```
astefanutti: Out of curiosity, I wonder why async is used here?

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -17,8 +17,14 @@
 
 from kubeflow.common.types import KubernetesBackendConfig
 from kubeflow.optimizer.backends.kubernetes.backend import KubernetesBackend
+from kubeflow.optimizer.constants import constants as optimizer_constants
```
andreyvelich: Do you need this change since we don't import common constants ?
```suggestion
from kubeflow.optimizer.constants import constants
```

kubeflow/optimizer/backends/base.py
```diff
@@ -44,3 +50,17 @@ def get_job(self, name: str) -> OptimizationJob:
...
+        name: str,
+        status: set[str] = {constants.OPTIMIZATION_JOB_COMPLETE},
+        timeout: int = 3600,
+        polling_interval: int = 2,
+    ) -> OptimizationJob:
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def get_best_trial(self, name: str) -> Optional[Trial]:
+        raise NotImplementedError()
```
andreyvelich, line 62: Please can we keep the order of APIs similar to Trainer: https://github.com/kubeflow/sdk/blob/b5626160d7e376e1ba662501df569b35fadb61e1/kubeflow/trainer/backends/base.py

```py
get_job()
wait_for_job_status()
get_best_trial()
delete_job()
```

kubeflow/optimizer/backends/base.py
```diff
@@ -44,3 +50,17 @@ def get_job(self, name: str) -> OptimizationJob:
     @abc.abstractmethod
     def delete_job(self, name: str):
         raise NotImplementedError()
+
+    @abc.abstractmethod
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: set[str] = {constants.OPTIMIZATION_JOB_COMPLETE},
+        timeout: int = 3600,
```
andreyvelich, line 55: Let's use same timeout as for Trainer for now?
```suggestion
        timeout: int = 600,
```

kubeflow/optimizer/backends/kubernetes/backend.py
andreyvelich: I think, we discussed it before, we have to watch for Pods in that case: https://github.com/kubeflow/sdk/pull/52#discussion_r2245398574

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -206,6 +207,116 @@ def list_jobs(self) -> list[OptimizationJob]:
...
+
+        raise TimeoutError(
+            f"Timeout waiting for {constants.OPTIMIZATION_JOB_KIND} {name} to reach status: "
+            f"{status}"
+        )
+
+    def get_best_trial(self, name: str) -> Optional[Trial]:
+        """Get the best current Trial for the OptimizationJob"""
+        try:
+            thread = self.custom_api.get_namespaced_custom_object(
```
andreyvelich: Can you create helper function to reduce code duplication to get Experiment object ?
https://github.com/kubeflow/sdk/blob/a5807e9f35a072b322aaf7099e109e14b7c0441f/kubeflow/optimizer/backends/kubernetes/backend.py#L323-L344

E.g.
```
self.__get_experiment_cr(name: str) -> models.V1beta1Experiment
```

andreyvelich (commented):
Thanks @kramaranya!

kubeflow/optimizer/backends/kubernetes/backend.py
astefanutti: @andreyvelich yes I remember now, but it's different in this case, we have all the statuses on the experiment / optimization job?

kramaranya: Yeah, but we actually still rely on Trials/TrainJobs via polling to know when it’s running

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: This has already been used to get experiment from cr, I think it lets us set a timeout for the API call.
@andreyvelich might have a better understanding

kubeflow/optimizer/backends/kubernetes/backend.py
andreyvelich: Yes, it might be hard to detect that OptimizationJob is running since we don't expose all events to the Experiment CRD.

kubeflow/optimizer/backends/base.py
kramaranya: I increased it since the default num_trials is 10, meaning optimization will take about 10 times longer than just training, right?

kubeflow/optimizer/backends/kubernetes/backend.py
andreyvelich: IIRC, that's right.
In the future we can set timeout to the Kubernetes API Server if we want to.

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: Sure, updated in https://github.com/kubeflow/sdk/pull/145/commits/5cb996672b305fcc4d273240a31318cfa3aa64fa

kubeflow/optimizer/backends/base.py
andreyvelich: I see, it should be fine in that case.

kubeflow/optimizer/backends/kubernetes/backend.py
astefanutti: OK that's clear thanks.

andreyvelich (commented):
Thanks @kramaranya!
/lgtm
/approve

PR #146, opened 2025-11-05
fix: Update Kubeflow SDK diagram

kramaranya:
The Kubeflow SDK diagram was missing some icons, so I updated it

@kubeflow/kubeflow-sdk-team

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ ML applications rather than managing complex infrastrutcure.
 - **Seamless Integration**: Designed to work together with all Kubeflow projects for end-to-end ML pipelines
 - **Local Development**: First-class support for local development requiring only `pip` installation
 
-![Kubeflow SDK Diagram](docs/images/kubeflow-sdk.drawio.svg)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,14 +19,7 @@ ML applications rather than managing complex infrastrutcure.
...
 
-<div style="text-align: center;">
-  <img
-    src="https://raw.githubusercontent.com/kubeflow/sdk/main/docs/images/kubeflow-sdk.drawio.svg"
-    width="600"
-    title="Kubeflow SDK Diagram"
-    alt="Kubeflow SDK Diagram"
-  />
-</div>
+![Kubeflow SDK Diagram](https://raw.githubusercontent.com/kubeflow/sdk/main/docs/images/kubeflow-sdk.drawio.svg)
```
andreyvelich, line 22: Do you know if this render works in PyPI ?
Shall we still use `<div>` ?

kramaranya: Yes, it'll work, for example feast does the same in [readme](https://github.com/feast-dev/feast/blob/master/README.md?plain=1#L41) and [pypi](https://pypi.org/project/feast/)
I used div there because I needed to adjust the size of the diagram, but we no longer need to do that

PR #147, opened 2025-11-05
fix(trainer): Remove --user flag from packages install in local subprocess

andreyvelich:
As we discussed, local subprocess fails when packages install with `--user` flag.


kramaranya (commented):
Thanks @andreyvelich!
/lgtm

PR #148, opened 2025-11-06
feat(optimizer): Add get_job_logs API to OptimizerClient

kramaranya:
I've added `get_job_logs` API to the OptimizerClient

Working example:
```python
from kubeflow.optimizer import OptimizerClient, Search, Objective, TrialConfig

def get_torch_dist(learning_rate: str, num_epochs: str):
    import os
    import torch
    import torch.distributed as dist

    dist.init_process_group(backend="gloo")
    print("PyTorch Distributed Environment")
    print(f"WORLD_SIZE: {dist.get_world_size()}")
    print(f"RANK: {dist.get_rank()}")
    print(f"LOCAL_RANK: {os.environ['LOCAL_RANK']}")

    lr = float(learning_rate)
    epochs = int(num_epochs)
    loss = 1.0 - (lr * 2) - (epochs * 0.01)

    if dist.get_rank() == 0:
        print(f"loss={loss}")
    
    dist.barrier()

template = TrainJobTemplate(
    trainer=CustomTrainer(
        func=get_torch_dist,
        func_args={"learning_rate": "0.01", "num_epochs": "5"},
        num_nodes=2,
        resources_per_node={"gpu": 1},
    ),
    runtime=TrainerClient().get_runtime("torch-distributed"),
)

job_id = OptimizerClient().optimize(
    trial_template=template,
    trial_config=TrialConfig(num_trials=10, parallel_trials=2),
    search_space={
        "learning_rate": Search.loguniform(0.001, 0.1),
        "num_epochs": Search.choice([5, 10, 15]),
    },
)

print(f"OptimizationJob created: {job_id}")

print("\n".join(OptimizerClient().get_job_logs(name=job_id)))
```

### DIFF
--- a/kubeflow/optimizer/api/optimizer_client.py
+++ b/kubeflow/optimizer/api/optimizer_client.py
@@ -26,7 +26,6 @@ from kubeflow.optimizer.types.optimization_types import (
     Trial,
     TrialConfig,
 )
-from kubeflow.trainer.constants import constants as trainer_constants
 from kubeflow.trainer.types.types import TrainJobTemplate
 
 logger = logging.getLogger(__name__)
@@ -125,7 +124,6 @@ class OptimizerClient:
         self,
         name: str,
         trial_name: Optional[str] = None,
-        step: str = trainer_constants.NODE + "-0",
         follow: bool = False,
     ) -> Iterator[str]:
         """Get logs from a specific trial of an OptimizationJob.
@@ -150,7 +148,6 @@ class OptimizerClient:
             trial_name: Optional name of a specific Trial. If not provided, logs from the
                 current best trial are returned. If no best trial is available yet, logs
                 from the first trial are returned.
-            step: Step of the Trial to collect logs from, like node-0.
             follow: Whether to stream logs in realtime as they are produced.
 
         Returns:
@@ -161,7 +158,7 @@ class OptimizerClient:
             TimeoutError: Timeout to get an OptimizationJob.
             RuntimeError: Failed to get an OptimizationJob.
         """
-        return self.backend.get_job_logs(name=name, trial_name=trial_name, follow=follow, step=step)
+        return self.backend.get_job_logs(name=name, trial_name=trial_name, follow=follow)
 
     def wait_for_job_status(
         self,
--- a/kubeflow/optimizer/backends/base.py
+++ b/kubeflow/optimizer/backends/base.py
@@ -54,7 +54,6 @@ class RuntimeBackend(abc.ABC):
         name: str,
         trial_name: Optional[str],
         follow: bool,
-        step: str,
     ) -> Iterator[str]:
         raise NotImplementedError()
 
--- a/kubeflow/optimizer/backends/kubernetes/backend.py
+++ b/kubeflow/optimizer/backends/kubernetes/backend.py
@@ -218,7 +218,6 @@ class KubernetesBackend(RuntimeBackend):
         name: str,
         trial_name: Optional[str] = None,
         follow: bool = False,
-        step: str = trainer_constants.NODE + "-0",
     ) -> Iterator[str]:
         """Get the OptimizationJob logs from a Trial"""
         # Determine what trial to get logs from.
@@ -237,6 +236,7 @@ class KubernetesBackend(RuntimeBackend):
 
         # Get the Trial's Pod name.
         pod_name = None
+        step = trainer_constants.NODE + "-0"
         for c in self.trainer_backend.get_job(trial_name).steps:
             if c.status != trainer_constants.POD_PENDING and c.name == step:
                 pod_name = c.pod_name


kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -119,6 +121,47 @@ def get_job(self, name: str) -> OptimizationJob:
 
         return self.backend.get_job(name=name)
 
+    def get_job_logs(
+        self,
+        name: str,
+        trial: Optional[str] = None,
```
andreyvelich: Can we use `trial_name` here ?
```suggestion
        trial_name: Optional[str] = None,
```

kubeflow/optimizer/backends/base.py
```diff
@@ -47,6 +48,16 @@ def list_jobs(self) -> list[OptimizationJob]:
     def get_job(self, name: str) -> OptimizationJob:
         raise NotImplementedError()
 
+    @abc.abstractmethod
+    def get_job_logs(
+        self,
+        name: str,
+        trial: Optional[str],
```
andreyvelich:
```suggestion
        trial_name: Optional[str],
```

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -212,6 +213,60 @@ def get_job(self, name: str) -> OptimizationJob:
...
+        trial: Optional[str] = None,
+        follow: bool = False,
+        step: str = trainer_constants.NODE + "-0",
+    ) -> Iterator[str]:
+        """Get the OptimizationJob logs from a Trial"""
+        # Determine what trial to get logs from.
+        if trial is None:
+            # Get logs from the best current trial.
+            best_trial = self.get_best_trial(name)
+            if best_trial is None:
```
andreyvelich, line 227: As we discussed, if the best Trial is empty, let's take the first Trial from the OptimizationJob if list is not empty: https://github.com/kubeflow/sdk/blob/63effe0ca460733e83524434dc947820aa646000/kubeflow/optimizer/types/optimization_types.py#L117

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -212,6 +213,60 @@ def get_job(self, name: str) -> OptimizationJob:
...
+                # TODO (kramaranya): Consider waiting for best trial when follow=True
+                return
+            trial = best_trial.name
+            logger.info(f"Getting logs from best trial: {trial}")
```
andreyvelich: Let's use `.debug` here, since we don't use `.info` logging in the SDK for now.
```suggestion
            logger.debug(f"Getting logs from best trial: {trial}")
```

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -212,6 +213,60 @@ def get_job(self, name: str) -> OptimizationJob:
         optimization_job = self.__get_experiment_cr(name)
         return self.__get_optimization_job_from_cr(optimization_job)
 
+    def get_job_logs(
+        self,
+        name: str,
+        trial: Optional[str] = None,
+        follow: bool = False,
+        step: str = trainer_constants.NODE + "-0",
```
andreyvelich: Maybe for now, we should remove step from this API ?
The problem is that for other Steps (e.g. Pods), the container name is not `metrics-logger-and-collector`.
As a workaround, users can always use `TrainerClient()` to get logs for other steps of TrainJob
(e.g. TrainJob name == Trial name)

kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -212,6 +213,60 @@ def get_job(self, name: str) -> OptimizationJob:
...
+        pod_name = None
+        for c in self.trainer_backend.get_job(trial).steps:
+            if c.status != trainer_constants.POD_PENDING and c.name == step:
+                pod_name = c.pod_name
+                break
+        if pod_name is None:
+            return
+
+        container_name = constants.METRICS_COLLECTOR_CONTAINER
+        try:
```
andreyvelich: To reduce code duplication can you wrap this code under helper function: `self.__read_pod_logs(pod_name: str, container_name: str, follow: bool) in `, the Trainer client and use it as:
```py
yield from self.trainer_backend.__read_pod_logs(
        pod_name=pod_name,
        container_name=container_name,
        follow=follow
    )
```

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: Sure, updated in https://github.com/kubeflow/sdk/pull/148/commits/66927dc6affc44cd9c74ae16c91d642c3ea02ae8

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: I was thinking about this too, makes sense to me, updated in https://github.com/kubeflow/sdk/pull/148/commits/e1a00b821e2b87ce688e9d3d9aa1e85982184c67

kubeflow/optimizer/backends/kubernetes/backend.py
kramaranya: Updated in https://github.com/kubeflow/sdk/pull/148/commits/49b4da0e00280e988f79d3bb9fa1968c6a9434f9

astefanutti:
Thanks @kramaranya 

/lgtm
/approve

I think we need to improve how Katib instruments the train nodes for StdOutCollector and FileCollector metrics to avoid depending on the metric collector sidecar and also make it work for other configurations like TfEventCollector or PrometheusMetricCollector.

Issue #161, opened 2025-11-17
pip install with --user argument fails with image running in python virtual environment


briangallagher:
### What happened?

When the training runtime image has a python virtual environment and it attempts a pip install with --user, we get the following error: 

```
(app-root) bash-5.1$ pip install --user datasets
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.
```

### What did you expect to happen?

pip install should always succeed regardless of container runtime image running as root or non root and in a virtual environment or not.

PR #162, opened 2025-11-17
fix: pip install with --user argument fails with image running in python virtual environment

briangallagher:
pip install with --user argument fails with image running in python virtual environment

I've added some conditions to adding the **--user** argument on pip install 
Currently we add **--user** for all pip installs
I've reverted the previous condition that was in place i.e. an mpi job
And I've added 2 additional conditions:
1. **Not** in a virtual environment **and**
2. **Not** root user

Fixes # 161

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -22,7 +22,6 @@ It tests KubernetesBackend's behavior across job listing, resource creation etc
 from dataclasses import asdict
 import datetime
 import multiprocessing
-import os
 import random
 import string
 from typing import Optional
@@ -218,15 +217,12 @@ def get_custom_trainer(
     env: Optional[list[models.IoK8sApiCoreV1EnvVar]] = None,
     pip_index_urls: Optional[list[str]] = constants.DEFAULT_PIP_INDEX_URLS,
     packages_to_install: list[str] = ["torch", "numpy"],
-    add_user: bool = False,
 ) -> models.TrainerV1alpha1Trainer:
     """
     Get the custom trainer for the TrainJob.
     """
     pip_command = [f"--index-url {pip_index_urls[0]}"]
     pip_command.extend([f"--extra-index-url {repo}" for repo in pip_index_urls[1:]])
-    if add_user:
-        pip_command.append("--user")
     pip_command = " ".join(pip_command)
 
     packages_command = " ".join(packages_to_install)
@@ -236,7 +232,10 @@ def get_custom_trainer(
             "-c",
             '\nif ! [ -x "$(command -v pip)" ]; then\n    python -m ensurepip '
             "|| python -m ensurepip --user || apt-get install python-pip"
-            "\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet"
+            "\nfi\n\n"
+            "PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet"
+            f" --no-warn-script-location {pip_command} --user {packages_command}"
+            " ||\nPIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet"
             f" --no-warn-script-location {pip_command} {packages_command}"
             "\n\nread -r -d '' SCRIPT << EOM\n\nfunc=lambda: "
             'print("Hello World"),\n\n<lambda>(**'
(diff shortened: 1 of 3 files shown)


kramaranya:
Fixes #161

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -272,6 +274,37 @@ def get_script_for_python_packages(
...
+
+def is_root_user() -> bool:
+    try:
+        return os.geteuid() == 0  # type: ignore[attr-defined]
+    except AttributeError:
+        # Platforms without geteuid (e.g., Windows). Treat as non-root.
+        return False
+
+
+def is_running_in_venv() -> bool:
```
astefanutti: That would have to run in the TrainJob pod?

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -248,8 +249,9 @@ def get_script_for_python_packages(
     # first url will be the index-url.
     options = [f"--index-url {pip_index_urls[0]}"]
     options.extend(f"--extra-index-url {extra_index_url}" for extra_index_url in pip_index_urls[1:])
-    # Always install for the user to avoid permission issues across environments.
-    options.append("--user")
+    # Add --user when appropriate.
+    if append_user(is_mpi):
```
astefanutti: I think this must also be removed from `python -m ensurepip --user` below.

andreyvelich (commented):
Thanks for this @briangallagher!
Do we know how KFP solves this problem ?
Do they always install pip packages without `--user` ?

kubeflow/trainer/backends/kubernetes/utils.py
briangallagher: Yeah, you're right, I need to refactor

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -262,10 +259,43 @@ def get_script_for_python_packages(
         """
     )
 
+    # Decide whether to append --user at container runtime:
+    # - Always for MPI
+    # - Or when running as non-root and not inside a virtual environment
+    runtime_logic = textwrap.dedent(
+        f"""\
+MPI_FLAG={1 if is_mpi else 0}
```
andreyvelich: I am curious wether it would be easier to just run `pip install --user` if that fails, run `pip install` without `--user`?
Something like:
```
pip install --user .... || pip install ....
```

briangallagher: We could but we would be trying pip install without **_--user_** regardless of the error. There could be multiple reasons for pip install failure. We could parse the error for permissions but that also seems a bit flaky. 

The PR is quite verbose but also more deterministic imo. Open to other suggestions.

andreyvelich: But if pip fails because of other issues, it eventually fails in the second call as well (`pip install package`), right ?

Which means users will see the underlying error like 404.

briangallagher: @andreyvelich Yes, that's probably sufficient. 
I've updated the PR to reflect that.

kubeflow/trainer/backends/kubernetes/utils.py
briangallagher: I think this is ok. Follows same pattern now as below suggested by Andrey

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -802,6 +804,27 @@ def test_get_runtime_packages(kubernetes_backend, test_case):
...
+            },
+            expected_output=get_train_job(
+                runtime_name=TORCH_RUNTIME,
+                train_job_name=TRAIN_JOB_WITH_CUSTOM_TRAINER,
+                train_job_trainer=get_custom_trainer(
+                    pip_index_urls=constants.DEFAULT_PIP_INDEX_URLS,
+                    packages_to_install=["torch", "numpy"],
+                ),
+            ),
+        ),
```
kramaranya: This seems to be identical to "valid flow with custom trainer"

kramaranya (commented):
Thank you, @briangallagher!
/lgtm

andreyvelich (commented):
Thank you @briangallagher!
/lgtm
/approve
@kubeflow/kubeflow-sdk-team Do we want to cherry-pick this to release-2.0 and release v0.2.1 ?

PR #158, opened 2025-11-11
feat: Support for Spark Client in Kubeflow SDK

Shekharrajak:
This PR introduces the Kubeflow Spark Client - a cloud-native
  Python client for managing Apache Spark applications on
  Kubernetes. It provides a unified, Pythonic interface
   for submitting, monitoring, and managing Spark jobs using the
  Kubeflow Spark Operator.

KEP: https://github.com/kubeflow/sdk/pull/163

Few examples added in examples/spark directory to play with. 

```
# Setup the env using 
cd examples/spark

./setup_test_environment.sh

# run simple example 
python test_spark_client_integration.py

# spark connect 
./setup_spark_connect.sh
python ipython_spark_connect_demo.py 

```

Slack thread: https://cloud-native.slack.com/archives/C074588U7EG/p1763656387742729?thread_ts=1763568656.642239&cid=C074588U7EG

PR #163, opened 2025-11-17
feat(docs): KEP- Spark Client for Kubeflow SDK

Shekharrajak:
Ref https://docs.google.com/document/d/1l57bBlpxrW4gLgAGnoq9Bg7Shre7Cglv4OLCox7ER_s/edit?tab=t.0

PR: https://github.com/kubeflow/sdk/pull/158

### DIFF
--- a/docs/proposals/107-spark-client/README.md
+++ b/docs/proposals/107-spark-client/README.md
@@ -240,7 +240,7 @@ class SparkClientBuilder:
 
         Args:
             conf: Dictionary of S3A configuration (endpoint, access.key, secret.key,
-                region, path.style.access, connection.ssl.enabled, etc.)
+                region, path.style.access, etc.)
         """
 
     def seaweedfs_config(
@@ -299,7 +299,6 @@ class SparkClient:
         self,
         url: Optional[str] = None,
         token: Optional[str] = None,
-        use_ssl: bool = False,
         name: Optional[str] = None,
         app_name: Optional[str] = None,
         num_executors: Optional[int] = None,
@@ -319,7 +318,6 @@ class SparkClient:
             url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
                  If provided, connects to existing server. If None, creates new session.
             token: Optional authentication token for existing server.
-            use_ssl: Whether to use SSL for connection (only for existing servers).
             name: Optional session name. Auto-generated if not provided (create mode only).
             app_name: Optional Spark application name (create mode only).
             num_executors: Number of executor instances (create mode only).


docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,791 @@
...
+
+#### Backend Modularity
+
+The SDK is designed for easy extension with new backends:
+
+```python
+# Custom backend implementation
+from kubeflow.spark.backends.base import SparkBackend
+
+class CustomBackend(SparkBackend):
```
Shekharrajak: Users can extend the backend, if they want to have any specific changes or different way to connect or submit spark job

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+- Write YAML configurations with extensive boilerplate
+- Separately handle application submission, monitoring, and log retrieval
+- Manage different Spark deployment patterns (standalone, client/server mode)
+
+This creates friction for data engineers and ML practitioners who want to focus on their data processing logic rather than infrastructure management.
+
+### Goals
+
+- Design a unified, Pythonic SDK for managing Spark applications on Kubernetes
+- Support multiple backends (Kubernetes Operator, REST Gateway, Spark Connect) following the Trainer pattern
```
andreyvelich: For Trainer, backends represent various job submission (local subprocess, container, and Kubernetes). I am not sure if we can replicate it for Spark.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+└──────────────────────┬──────────────────────────────────────┘
+                       │ implemented by
+           ┌───────────┴─────────────┬──────────────────┬────────────────┐
+           ▼                         ▼                  ▼                ▼
+┌──────────────────┐   ┌──────────────────┐   ┌──────────────────┐   ┌──────────────────┐
+│OperatorBackend   │   │  GatewayBackend  │   │ ConnectBackend   │   │  LocalBackend    │
+│(Spark Operator   │   │  (REST Gateway)  │   │  (Spark Connect/ │   │  (Future)        │
+│ CRDs on K8s)     │   │                  │   │   Interactive)   │   │                  │
+└──────────────────┘   └──────────────────┘   └──────────────────┘   └──────────────────┘
+     Batch Jobs            Batch Jobs            Sessions              Local Dev
```
andreyvelich: Can you explain reason of creating various backends? Can we just have an API: `SparkClient().connect()` which creates SparkConnect CR and connects to the existing cluster as we discussed?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+```python
+from kubeflow.spark import SparkClient
+
+# Create client (uses Operator backend by default)
+client = SparkClient()
+
+# Submit a Spark application
+response = client.submit_application(
+    app_name="spark-pi",
+    main_application_file="local:///opt/spark/examples/src/main/python/pi.py",
```
andreyvelich: I am wondering if there is a way to allow users bypass function to SparkApplication similar to Trainer API: https://github.com/kubeflow/sdk?tab=readme-ov-file#run-your-first-pytorch-distributed-job

That might be interesting to explore how we can allow to submit SparkApplication without building an image.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+
+```python
+from kubeflow.spark import SparkClient
+
+# Create client (uses Operator backend by default)
+client = SparkClient()
+
+# Submit a Spark application
+response = client.submit_application(
+    app_name="spark-pi",
```
andreyvelich: Can name be automatically generated similar to TrainJob and OptimizationJob by default ?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+    driver_memory="512m",
+    executor_cores=1,
+    executor_memory="512m",
+    num_executors=2
+)
+
+print(f"Submitted: {response.submission_id}")
+
+# Wait for completion
+status = client.wait_for_completion(response.submission_id)
```
andreyvelich: As we discussed, please make APIs consistent with TrainerClient: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py
e.g.
```
wait_for_job_status
list_jobs
...
```

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+    driver_cores=2,
+    driver_memory="4g",
+    executor_cores=2,
+    executor_memory="8g",
+
+    # Enable dynamic allocation (auto-scaling)
+    enable_dynamic_allocation=True,
+    initial_executors=2,
+    min_executors=1,
+    max_executors=10,
```
andreyvelich: You don't need enable flag, if you add optional parameter in `submit_job()` API:
```py
dynamic_allocation=DynamicAllocation(
initial_executors=2,
min_executors=1,
max_executors=10
)

File: docs/proposals/107-spark-client/README.md [hunk 8]
```diff
@@ -0,0 +1,1023 @@
...
+
+    # Configure volumes for data access
+    volumes=[{
+        "name": "data-volume",
+        "persistentVolumeClaim": {"claimName": "my-pvc"}
+    }],
+    driver_volume_mounts=[{
+        "name": "data-volume",
+        "mountPath": "/data"
+    }],
```
andreyvelich: Can we configure these parameters via options, like how we did if for Trainer: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L110

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+client = SparkClient(backend_config=config)
+
+response = client.submit_application(
+    app_name="advanced-job",
+    main_application_file="local:///app/job.py",
+    spark_version="4.0.0",
+    driver_cores=2,
+    driver_memory="4g",
+    executor_cores=2,
+    executor_memory="8g",
```
andreyvelich: I would move these values under `resources=Resources()` class.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+    executor_gpu={"name": "nvidia.com/gpu", "quantity": 2},
+
+    # Node selector and tolerations
+    node_selector={"node-type": "compute"},
+    tolerations=[{
+        "key": "spark",
+        "operator": "Equal",
+        "value": "true",
+        "effect": "NoSchedule"
+    }],
```
andreyvelich: Same suggestion to move this in options.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+
+```python
+from kubeflow.spark.backends import OperatorBackendConfig
+
+config = OperatorBackendConfig(
+    namespace="default",                     # K8s namespace
+    context=None,                            # kubeconfig context
+    service_account="spark-operator-spark",  # Service account
+    image_pull_policy="IfNotPresent",        # Image pull policy
+    default_spark_image="gcr.io/spark-operator/spark-py",
```
andreyvelich: So you suggest that user creates a new SparkClient every time they want to change image?
For example, in TrainerClient users can specify image for every TrainJob submission: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L116-L117

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+    enable_monitoring=True,                  # Prometheus monitoring
+    enable_ui=True,                          # Spark UI
```
andreyvelich: Why do we need this flag ?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+- Manages lifecycle through Kubernetes API
+- Supports all Spark Operator features (dynamic allocation, monitoring, GPU)
+- Provides log streaming from driver and executor pods
+- Enables Spark UI access through services/ingress
+
+#### 3. GatewayBackend (REST API)
+
+Implements the backend for managed Spark gateways:
+
+- Submits applications via REST endpoints
```
andreyvelich: Who is managing this gateway service?
IIUC, Spark Operator doesn't provide such feature out of the box.
Right @ChenYi015 @vara-bonthu @nabuskey @yuchaoran2011 ?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+- Provides session-based and batch job execution
+
+
+#### 3. ConnectBackend (Spark Connect / Interactive Sessions)
+
+Implements session-oriented backend using Spark Connect protocol (gRPC):
+
+**Key Features**:
+- Remote connectivity to existing Spark clusters via Spark Connect (gRPC)
+- Session-based interactive workloads (exploratory analysis, notebooks)
```
andreyvelich: We also need to figure out whether this capability might be useful for Notebooks use-cases.
I know that @lresende and @fresende did some work to integrate Spark Operator with Jupyter Enterprise gateway: https://www.kubeflow.org/docs/components/spark-operator/user-guide/notebooks-spark-operator/

cc @ederign @andyatmiami @akshaychitneni @shravan-achar

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+
+config = ConnectBackendConfig(
+    connect_url="sc://spark-cluster.default.svc:15002",
+    token="bearer-token",  # Optional authentication
+    use_ssl=True,          # SSL/TLS for secure communication
+)
+client = SparkClient(backend_config=config)
+
+# Create interactive session
+session = client.create_session(app_name="data-analysis")
```
andreyvelich: We need to see how we can leverage SparkConnect CR, created by @ChenYi015

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+
+with dev_client.create_session("dev") as session:
+    # Test and validate query
+    test_df = session.sql("SELECT * FROM data LIMIT 1000")
+    test_df.show()
+    # Iterate and refine...
+
+# Step 2: Production batch job with OperatorBackend
+prod_config = OperatorBackendConfig(namespace="production")
+prod_client = SparkClient(backend_config=prod_config)
```
andreyvelich: This is pretty interesting experience for dev -> prod Spark lifecycle.
cc @shravan-achar @akshaychitneni @bigsur0 to explore.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+    state: ApplicationState  # Enum: NEW, RUNNING, COMPLETED, FAILED, etc.
+    app_id: Optional[str]    # Spark application ID
+    driver_info: Optional[Dict]
+    executor_info: Optional[Dict]
+    submission_time: Optional[str]
+    completion_time: Optional[str]
+    message: Optional[str]
+```
+
+### User Stories
```
andreyvelich: Could you move these user stories to the beginning of the KEP?
Here is the example: https://github.com/kubeflow/sdk/tree/main/docs/proposals/46-hyperparameter-optimization#user-stories

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+)
+spark_client.wait_for_completion(features.submission_id)
+
+# Step 2: Model training with Trainer
+trainer_client = TrainerClient()
+trainer_client.train(
+    name="train-model",
+    func=train_func,
+    num_nodes=4,
+)
```
andreyvelich:
```suggestion
trainer_client.train(
    trainer=CustomTrainer(
        func=train_func,
        num_nodes=4
        )
  )
```

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1023 @@
...
+```
+
+#### Integration with Pipelines
```
andreyvelich: cc @kubeflow/wg-pipeline-leads to explore

docs/proposals/107-spark-client/README.md
Shekharrajak: Yes, it is following the same pattern. The user have to create the Spark Connect server and pass the url into spark session backend. 

yaml which worked locally : https://github.com/kubeflow/sdk/pull/158/files#diff-09efc73ad2dcad6d7c80cdaa710e3c49d1dcdb5b4f15d91ecb82a2c6053d22b5R82

docs/proposals/107-spark-client/README.md
Shekharrajak: The batch backend will have APIs like `submit_application`, `wait_for_completion`, `get_logs`, .. where user will just submit the job and can check the logs/results.
example:  https://github.com/kubeflow/sdk/pull/158/files#diff-e692a5819ee6b1dc00cba3b58e91f058c0022d3ca9aa6f3ee468126f245eef89R89

But with interactive session spark client user will be able to  run interactive SQL queries and DataFrame operations. 
example: https://github.com/kubeflow/sdk/pull/158/files#diff-a5011f48c9d6d16ff6ddd65588f65a7c78abf5fbeb121cccb693c0892ce3a5aeR275

docs/proposals/107-spark-client/README.md
Shekharrajak: Updated in https://github.com/kubeflow/sdk/pull/158/commits/5fc0308f7f29a3d768aa626e73052aa7a889b113

docs/proposals/107-spark-client/README.md
Shekharrajak: Updated in commit https://github.com/kubeflow/sdk/pull/158/commits/5fc0308f7f29a3d768aa626e73052aa7a889b113

Shekharrajak: Updated the KEP https://github.com/kubeflow/sdk/pull/163/commits/5cadd5c1e5a59b4306650f29d2bc97d0fdfb3642

docs/proposals/107-spark-client/README.md
Shekharrajak: Updated the KEP https://github.com/kubeflow/sdk/pull/163/commits/5cadd5c1e5a59b4306650f29d2bc97d0fdfb3642

docs/proposals/107-spark-client/README.md
Shekharrajak: Done https://github.com/kubeflow/sdk/pull/163/commits/90ba54e2b2f4fac1bd3817e41e66de14b54f96c7

docs/proposals/107-spark-client/README.md
Shekharrajak: Here we have job submission using K8S Operator backend, Spark Connector backend , Gateway backend (not implemented - we just have abstract class).

docs/proposals/107-spark-client/README.md
Shekharrajak: Updated https://github.com/kubeflow/sdk/pull/163/commits/db6f3b3d3553b0a7fe04dca00c5f6800d9dc66a4

Shekharrajak:
```
client.submit_application(
    app_name="custom-job",
    image="company-registry.io/spark-ml:3.5.0",
    main_application_file="s3a://bucket/ml_job.py",
)

```
User should be able to pass the image in each submission.

docs/proposals/107-spark-client/README.md
andreyvelich: I am wondering what is the main motivation to separate `SessionClient()` and `BatchClient()`?
Alternatively, we can just have unified `SparkClient()` which has sessions and batch APIs:
```py
submit_job() <-- to create Spark Application and submit batch job
connect() <-- to create session and connect to existing Spark cluster
```

Shekharrajak: - BatchSparkClient users never see connect(), create_session() methods
- SparkSessionClient users never see submit_application(), wait_for_job_status() methods
- This prevents runtime errors: Can't call wait_for_job_status() on a session object.

Shekharrajak: This helps user to know which methods are available clearly : 

```
 client = SparkClient()
  # User sees BOTH batch AND session methods
  client.submit_job(...)        # For batch
  client.connect(...)           # For session
  client.get_job(...)           # Works with connect() or batch()  ? 

 # event if we take arg in config: 
  client = SparkClient(mode="batch")
  client.create_session(...)  # IDE will not show error, but runtime error 
```

andreyvelich: Do we know how the Spark session management will work with the new CRD: SparkCluster proposed by @ChenYi015 here: https://github.com/kubeflow/spark-operator/issues/2744 ?

Shekharrajak: Added my view: https://github.com/kubeflow/spark-operator/issues/2744#issuecomment-3574094597

Shekharrajak: If we want to create the spark cluster through the SDK then we need API like :

```
 from kubeflow.spark import SparkClusterManager

 manager = SparkClusterManager()

 # Create persistent cluster
 cluster = manager.create_cluster(
     name="prod",
     spark_version="4.0.1",
     init_workers=5,
     min_workers=2,
     max_workers=10,
     enable_hpa=True,
 )

# Cluster url to use for spark job submission: cluster.master_url
```

docs/proposals/107-spark-client/README.md
Shekharrajak: Are we thinking an API like this ?

```
def my_job(input_path: str, output_path: str):                                                              
    from pyspark.sql import SparkSession
    spark = SparkSession.builder.getOrCreate()                                                         
    df = spark.read.csv(input_path)                                                                    
    df.groupBy("col").count().write.parquet(output_path)                                               
    spark.stop()                                                                                       
                                                                                                                    
client.submit_function(                                                                                
    func=my_job,                                                                                       
    func_args={"input_path": "s3://in", "output_path": "s3://out"}                                     
) 
```

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1125 @@
...
+
+- Design specialized Pythonic SDKs for managing Spark applications on Kubernetes
+- Support multiple backends (Kubernetes Operator, REST Gateway, Spark Connect) following the Trainer pattern
+- Provide separate clients for batch jobs (BatchSparkClient) and interactive sessions (SparkSessionClient)
+- Follow Interface Segregation Principle to prevent runtime errors and improve type safety
+- Enable seamless integration with Kubeflow ML pipelines and workflows
+- Provide comprehensive monitoring, logging, and debugging capabilities
+- Support production features: dynamic allocation, GPU scheduling, volumes, monitoring
+- Support interactive data exploration and notebook-style workflows via Spark Connect
+- Deliver excellent developer experience with minimal boilerplate
```
aagumin: Is it possible to expect an Airflow provider to be available? Or cncf provider with SparkSubmitKubernetesOperator it's main flow?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1125 @@
...
+
+# Configure connection
+config = ConnectBackendConfig(
+    connect_url="sc://spark-cluster:15002",
+    use_ssl=True,
+)
+
+# Create client and session
+client = SparkSessionClient(backend_config=config)
+session = client.create_session(app_name="analysis")
```
aagumin: How will this be better than a native SparkConnect call? I can see that you can’t replace a Connect client with a regular one without changing interfaces, considering that the options for configuring a Connect session are significantly more limited.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,1125 @@
...
+    num_executors=5,
+    spark_conf={
+        "spark.sql.shuffle.partitions": "200",
+        "spark.hadoop.fs.s3a.endpoint": "s3.amazonaws.com",
+    },
+    env_vars={
+        "AWS_ACCESS_KEY_ID": "your-key",
+        "AWS_SECRET_ACCESS_KEY": "your-secret",
+    }
+)
```
aagumin: Maybe consider allowing an .env file to be passed? Since the list can be very long and reduce overall readability.

docs/proposals/107-spark-client/README.md
Shekharrajak: Here the scope is how we can have spark job submission and spark job life cycle being in kubeflow ecosystem. 
If user using airflow for spark jobs then Airflow  PythonOperator can help running the python code (import kubeflow ).

docs/proposals/107-spark-client/README.md
aagumin: If a [PR](https://github.com/kubeflow/spark-operator/pull/2629) is added for gRPC submit support and the ability to implement a custom submit mechanism, then this becomes relevant.

docs/proposals/107-spark-client/README.md
andreyvelich: That's right. @aagumin you can easily leverage Airflow Python Operator to use `SparkClient()`.

docs/proposals/107-spark-client/README.md
andreyvelich: Yeah, that might be interesting to explore.
We can also extend `submit_job()` API to accept main_application_file or function.

Shekharrajak: Let me add to KEP and will work on it.

docs/proposals/107-spark-client/README.md
Shekharrajak: Yes, we can think of such enhancement after initial version and MVP.

docs/proposals/107-spark-client/README.md
Shekharrajak: Spark  requires pre-allocated cluster resources (driver/executor specifications), so I could not find such way for running spark job. But we can explore in next version.

docs/proposals/107-spark-client/README.md
Shekharrajak: this is added in KEP.

docs/proposals/107-spark-client/README.md
Shekharrajak: We will have more kubeflow compatible APIs and infra.

docs/proposals/107-spark-client/README.md
Shekharrajak: Spark related configs.

docs/proposals/107-spark-client/README.md
Shekharrajak: We can take this as next phase implementation or simplification.

andreyvelich: Could we design this from the start, so we introduce less breaking changes in the future?

Do you have any objections to move these parameters under `Resources()` instance?

docs/proposals/107-spark-client/README.md
andreyvelich: @vara-bonthu @ChenYi015 @nabuskey Do you have any ideas on how we can dynamically inject the users' code into Spark driver without re-building an image or put Python file into Pod's filesystem?

I have one idea if we have can leverage [PodTemplate spec](https://github.com/kubeflow/spark-operator/blob/master/examples/spark-pi-pod-template.yaml).
Let's say we define InitContainer that simply creates file in the Pod's filesystem with the Spark code in the pre-defined location (e.g. `local:///opt/spark/spark_job.py`). This initContainer can share emptyDir with the driver container, so driver container can see the file. To create the file, we can use the same script as for Trainer:
```yaml
container:
  command:
    - bash
    - -c
    - |
      read -r -d '' SCRIPT << EOM
      def my_job(input_path: str, output_path: str):                                                           
          from pyspark.sql import SparkSession
          spark = SparkSession.builder.getOrCreate()                                                         
          df = spark.read.csv(input_path)                                                                    
          df.groupBy("col").count().write.parquet(output_path)                                               
          spark.stop()    

      EOM
      printf "%s" "$SCRIPT" > "/opt/spark/spark_job.py"
```

WDYT ?

cc @shravan-achar @akshaychitneni @bigsur0 for feedback on this.

aagumin: @andreyvelich 

Sorry to interrupt. I implemented a decorator that uses inspect to get the function's body and form an executable file. It also parses the function's signature into submit parameters (they must be JSON-serializable). I could implement this if you're interested.

Example
```python
@spark_submit
def my_task(first, second):
    ...

```

andreyvelich: @aagumin Yeah, this looks similar to how we do this in `TrainerClient()`: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/utils.py#L297-L301
I think, having the consistent UX across Data Processing (e.g. SparkJobs) and Training (e.g. TrainJob) workloads would be really cool.
As long as SparkApplication CRD can support initContainer via PodTemplate, that should be relatively easy to implement.

cc @franciscojavierarceo @kubeflow/kubeflow-sdk-team

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+client.stop()
+```
+
+### Full Configuration (Builder Pattern)
+
+```python
+from kubeflow.spark import SparkClient
+
+client = (
+    SparkClient.builder()
```
andreyvelich: I like the idea of builder(), but shall this comes in the 2nd phase?
We have similar ideas with @kubeflow/kubeflow-sdk-team  for TrainerClient and OptimizerClient, so maybe we can unify the experience?

Initially, we can just allow users to specify minimal number of parameters in the `connect()` API.

But I am happy to debate it during our next SDK call on Dec 17th: https://docs.google.com/document/d/1jH2WAX2ePxOfI4JuiVK9nPlesDMiyg67xzLwhpR7wTQ/edit?tab=t.0#heading=h.jarjwwjkqmul

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+### Connect to Existing Server
+
+```python
+from kubeflow.spark import SparkClient
+
+# Connect to existing Spark Connect server (no auto-provisioning)
+client = SparkClient.connect("sc://spark-server:15002")
+spark = client.session()
+df = spark.read.parquet("s3a://bucket/data/")
+client.stop()
```
andreyvelich: Can the `connect()` API first check if the SparkConnect CRD exists, if it doesn't exist create one and connect into it?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+## SparkClient API
+
+### SparkClientBuilder
+
+The Builder Pattern provides a fluent API for configuration - readable, discoverable, and extensible.
+
+```python
+class SparkClientBuilder:
+    """Builder for SparkClient configuration."""
```
andreyvelich: Will SparkApplication and SparkConnect CRD always have the same APIs? I am wondering if we can re-use the same builder for them.
cc @ChenYi015

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+    # Kubernetes settings
+    def namespace(self, ns: str) -> "SparkClientBuilder":
+        """Set Kubernetes namespace."""
```
andreyvelich: I think, namespace should be part of KubernetesBackendConfig in SparkClient to align with other clients: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L42-L43

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+    def logs(self, follow: bool = False) -> Iterator[str]:
+        """Get server logs."""
+
+    # Context manager
+    def __enter__(self) -> "SparkClient": ...
+    def __exit__(self, ...): ...
+
+    # === Batch Job Support ===
+    def submit_batch(
```
andreyvelich: Maybe calling it: submit_job()?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+```python
+from kubeflow.spark import SparkClient
+from kubeflow.trainer import TrainerClient, CustomTrainer
+from kubeflow.trainer.types import S3DatasetInitializer
+
+# Step 1: Run Spark feature engineering
+spark_client = SparkClient.builder().namespace("ml-jobs").build()
+
+feature_job = spark_client.submit_batch(
+    main_file="s3a://ml/feature_pipeline.py",
```
andreyvelich, line 571: As we discussed, it would be nice to add samples in this proposal on how users can pass the function with Spark transformations directly into `submit_job()` API.
After the 1st phase with `connect()` API support, we can add such functionality.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+## Architecture
+
+```
+SparkClient
+    │
+    ├── session() ──► creates Spark Connect server (if needed)
+    │                     │
+    │                     ▼
+    │               SparkApplication CRD
```
andreyvelich: Does SparkConnect CRD create SparkApplication CRD ?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+        main_file="s3a://ml/etl.py",
+        executor_instances=20,
+        executor_memory="8g",
+    )
+    
+    # Training step depends on ETL completion
+    train = TrainOp(
+        dataset_path=etl.outputs["output_path"],
+    )
+    train.after(etl)
```
andreyvelich, line 809: We should whether we need to move it under: `kubeflow.pipelines.components`.
We can also leverage KFP lightweight Python tasks to submit such pipelines.

cc @franciscojavierarceo @kubeflow/wg-pipeline-leads

andreyvelich (commented):
Thanks for the update @Shekharrajak!
I left a few comments.

docs/proposals/107-spark-client/README.md
franciscojavierarceo: +1

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+```
+
+### Full Configuration (Builder Pattern)
+
+```python
+from kubeflow.spark import SparkClient
+
+client = (
+    SparkClient.builder()
+    .namespace("spark-jobs")
```
juliusvonkohout: Will this integrate well with our current multi-tenancy architecture ? https://github.com/kubeflow/manifests#architecture

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+    def service_account(self, sa: str) -> "SparkClientBuilder":
+        """Set service account for Spark pods."""
+
+    # Resource configuration
+    def driver(self, cores: int = 1, memory: str = "1g") -> "SparkClientBuilder":
+        """Configure driver resources."""
+
+    def executor(self, cores: int = 1, memory: str = "1g") -> "SparkClientBuilder":
+        """Configure executor resources."""
+
```
juliusvonkohout: I am missing a bit the CPU and memory requests vs Limit . I only see

``` 
.driver(cores=2, memory="4g")
.executor(cores=4, memory="8g")
```

which is somehow mixing up requests with limits. Furthermore we should make sure that it also works with the integrated seaweedfs S3.

docs/proposals/107-spark-client/README.md
juliusvonkohout: So if i execute this from a jupyterlab will it just use the current namespace?

docs/proposals/107-spark-client/README.md
kramaranya: Agree, I'd like us to start with constructor and config objects aligning with existing clients and later expand it with builder approach

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+        """Shortcut to session()."""
+
+    def stop(self) -> None:
+        """Stop session and cleanup server."""
+
+    def status(self) -> ServerStatus:
+        """Get Spark Connect server status."""
+
+    def logs(self, follow: bool = False) -> Iterator[str]:
+        """Get server logs."""
```
kramaranya: This might be confusing comparing to get_job_logs() API. Could we rename it to get_server_logs() for example?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+    @classmethod
+    def connect(cls, url: str, token: str = None, use_ssl: bool = False) -> "SparkClient":
+        """Connect to an existing Spark Connect server (no auto-provisioning)."""
+
+    def session(self, app_name: Optional[str] = None) -> SparkSession:
+        """Get or create SparkSession. Creates Spark Connect server if needed."""
+
+    @property
+    def spark(self) -> SparkSession:
+        """Shortcut to session()."""
```
kramaranya: So why do we need this shortcut? What are the use cases when you'd choose spark over session?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,597 @@
...
+    ├── session() ──► creates Spark Connect server (if needed)
+    │                     │
+    │                     ▼
+    │               SparkApplication CRD
+    │                     │
+    │                     ▼
+    │               Spark Driver Pod
+    │               (with Spark Connect enabled)
+    │                     │
+    └── connect() ──────► sc://server:15002 ──► SparkSession
```
kramaranya: So how batch jobs differ from interactive session?

docs/proposals/107-spark-client/README.md
kramaranya: Hmm currently SparkClient is stateful and it remembers the session, so `client.session()` stores a reference, and then `client.stop()` knows which session to stop. But batch jobs work differently since they're managed by name `client.delete_job(name)` and the client doesn't track them
This means that you can only have one session per client, right? and `client.stop()` is confusing when you have both sessions and jobs running

What do you think if we make the client stateless like TrainerClient? So `create_session()` would return a session object that manages itself (so you can `session.stop()`) and `submit_job()` returns a job name that you manage through the client:

```python
session = client.create_session()
session.stop()

job_name = client.submit_job(...)
client.delete_job(job_name)
```
wdyt @Shekharrajak @andreyvelich ?

docs/proposals/107-spark-client/README.md
Shekharrajak: As of now explicit way of asking connect end point looks fine otherwise we will not be knowing which spark connect  node it is connected and running.

docs/proposals/107-spark-client/README.md
Shekharrajak: Yes

docs/proposals/107-spark-client/README.md
Shekharrajak: Batch (SparkApplication CRD) : Run a predefined Spark application to completion (ETL pipelines, ML training, data processing)

Interactive Sessions ((SparkConnect CRD)): Long-running server that accepts ad-hoc queries from clients (notebooks, BI tools, data exploration)

docs/proposals/107-spark-client/README.md
Shekharrajak: SparkConnect bypasses SparkApplication entirely. The executor spawning is done by Spark itself (the JVM process which is running in spark connect pod ), not by the Spark Operator.

Shekharrajak: Let me correct it. Thanks for pointing it out.

docs/proposals/107-spark-client/README.md
Shekharrajak: Updated

docs/proposals/107-spark-client/README.md
Shekharrajak: This would be internal changes, we can take care of it later as well.

docs/proposals/107-spark-client/README.md
Shekharrajak: Added .

docs/proposals/107-spark-client/README.md
Shekharrajak: Thanks for pointing it out. There are lot of conf is spark so anyways we need wider key value pair set.

docs/proposals/107-spark-client/README.md
Shekharrajak: https://github.com/kubeflow/sdk/pull/163/commits/e378e42e239998d2d4350a967838f44d013081a0 - I find that key, value dict map as arg makes it extendable , so we should have spark_conf, driver_conf and executor_conf as dict and user can pass any valid key value pair that works in CRD .

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,757 @@
...
+```python
+class SparkClientBuilder:
+    """Builder for advanced SparkClient configuration."""
+
+    # Backend configuration
+    def backend(self, config: Union[KubernetesBackendConfig, GatewayBackendConfig]) -> "SparkClientBuilder":
+        """Set backend configuration (namespace, context, etc.)."""
+
+    # Kubernetes settings
+    def service_account(self, sa: str) -> "SparkClientBuilder":
```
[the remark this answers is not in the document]
juliusvonkohout: Yes also pronounceability is important

docs/proposals/107-spark-client/README.md
andreyvelich: @Shekharrajak I am not sure that introducing unstructured CRD fields via `executor_conf` and `driver_conf` is clear for user.
They should always understand what fields are supported via CRD which introduce unnecessary complexity.
If we need to separate Driver from Executor, we can do something like this (similar as Trainer, Runtime, Initilizer in `train` API):
```py
client.connect(
    driver=Driver(
        image="spark-image"
    ),
    executor=Executor(
        resources_per_executor=Resources(
            "cpu": 5,
            "memory": 10,
        ),
        num_instances=5
    ),
)
```

However, I am trying to understand whether users really require to have control over Driver spec in the `connect()` API? Can we start with something very simple initially ?
```py
client.connect(
    num_executors=5,
    resources_per_executor=Resources(
        "cpu": 5,
        "memory": 10,
    ),
    spark_conf={
        "spark.sql.adaptive.enabled": true
    }
)
```
And introduce additional complex Kubernetes configurations via `options` parameter.

WDYT @Shekharrajak @kubeflow/kubeflow-sdk-team ?

Shekharrajak: Thanks, yes this is looking good. Let's go for it. Let me update it accordingly.

Shekharrajak: Updated the KEP: [https://github.com/kubeflow/sdk/pull/163/changes/c2a72fa27137f16f1f4b89542706b31645163d16#diff-4715d58efaf92eebc6656b[…]919a77f821c23eddb715045bR103](https://github.com/kubeflow/sdk/pull/163/changes/c2a72fa27137f16f1f4b89542706b31645163d16#diff-4715d58efaf92eebc6656b58201fc98804a1ed81919a77f821c23eddb715045bR103)

docs/proposals/107-spark-client/README.md
```diff
@@ -89,31 +89,60 @@ spark.stop()
...
 
-### Key-Value Configuration Design
+### Structured Types
+
+```python
+from dataclasses import dataclass
+from typing import Optional
+
+@dataclass
+class Resources:
```
Shekharrajak: This is used for Driver#resources  and Executor#resources_per_executor - we are following composite design pattern.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+```python
+resources_per_executor = {
+    "cpu": "5",
+    "memory": "10Gi",
+    "cpu_limit": "8",
+    "memory_limit": "12Gi"
+}
+```
+
+### Structured Types
```
andreyvelich, line 159: I guess, we can remove this from the version 1, right?
And implement this in the 2nd phase if needed.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+### Resource Configuration
+
+Resources are specified as dictionaries:
+
+```python
+resources_per_executor = {
+    "cpu": "5",
+    "memory": "10Gi",
+    "cpu_limit": "8",
+    "memory_limit": "12Gi"
```
andreyvelich: Let's keep only cpu and memory where requests == limits always in the 1st version
```suggestion
```

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+
+options = [
+    Labels({"app": "spark"}),
+    Annotations({"description": "ETL job"}),
+    PodTemplateOverrides(...)
+]
+```
+
+```python
+class SparkClientBuilder:
```
andreyvelich, line 195: Can you move Builder to the future plans, since we will implement it later if needed.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+
+        This method supports two modes based on parameters:
+        - **Connect mode**: When `url` is provided, connects to an existing Spark Connect server
+        - **Create mode**: When `url` is not provided, creates a new Spark Connect session
+
+        Args:
+            url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
+                 If provided, connects to existing server. If None, creates new session.
+            token: Optional authentication token for existing server.
+            use_ssl: Whether to use SSL for connection (only for existing servers).
```
andreyvelich: Does SparkConnect CRD support ssl and auth?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+            name: Optional session name. Auto-generated if not provided (create mode only).
+            app_name: Optional Spark application name (create mode only).
```
andreyvelich, line 322: We don't need this, right ?

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+        This method supports two modes based on parameters:
+        - **Connect mode**: When `url` is provided, connects to an existing Spark Connect server
+        - **Create mode**: When `url` is not provided, creates a new Spark Connect session
+
+        Args:
+            url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
+                 If provided, connects to existing server. If None, creates new session.
+            token: Optional authentication token for existing server.
+            use_ssl: Whether to use SSL for connection (only for existing servers).
+            name: Optional session name. Auto-generated if not provided (create mode only).
```
andreyvelich, line 321: WIll name be equal to the SparkConnect CR name? Can we move it to options for now, to be consistent with Trainer.

docs/proposals/107-spark-client/README.md
```diff
@@ -0,0 +1,859 @@
...
+
+        Args:
+            name: Name of the Spark Connect session.
+            follow: Whether to stream logs in realtime.
+        """
+
+    def delete_session(self, name: str) -> None:
+        """Delete a Spark Connect session."""
+
+    def submit_job(
```
andreyvelich, line 370: Let's say that job submission APIs will be implemented in the phase 2.

andreyvelich (commented):
Thanks @Shekharrajak!
Mostly lgtm, I suggest that we move some topics to phase 2 as we discussed.
/assign @kramaranya @kubeflow/kubeflow-sdk-team for the final feedback.

Shekharrajak:
I understand there will be bunch of iterations on the API design and to improve user experience. But definitely we will focus on first MVP as simple as possible with **strong foundation** and phase wise implement features.

The more important point is to start the journey otherwise we will see no progress.

docs/proposals/107-spark-client/README.md
Shekharrajak: If user want to provide specific name - let's park it for phase 2

docs/proposals/107-spark-client/README.md
Shekharrajak: Yes +1

docs/proposals/107-spark-client/README.md
Shekharrajak: No built-in SSL/TLS or authentication support in the SparkConnect CRD. I was thinking for external servers or submitting spark job. But let me remove it for now  - we will think later on this.

Shekharrajak: Removed it for now https://github.com/kubeflow/sdk/pull/163/commits/a2a4837be19f4b506e3714c9a1c57dbe228d26ad

docs/proposals/107-spark-client/README.md
Shekharrajak: Yeah, we will start simpler in first MVP

andreyvelich (commented):
Yeah, I agree with you.
Thank you for this work @Shekharrajak, excited to see this moving forward!
/lgtm
/approve
/hold in case @kramaranya @astefanutti @szaher @Fiona-Waters wanted to give more comments.
/assign @kramaranya @astefanutti @szaher @Fiona-Waters

kramaranya (commented):
Looks good to me! Thanks @Shekharrajak for your work
/lgtm

PR #166, opened 2025-11-18
fix(trainer): Remove namespace from ClusterTrainingRuntime exception messages

astefanutti:

ClusterTrainingRuntime resources are not namespace-scoped so the namespace info is not relevant.

**Checklist:**


kramaranya (commented):
Thank you!
/lgtm

andreyvelich (commented):
/lgtm
/approve

PR #168, opened 2025-11-18
fix(trainer): Use PyTorch static rendezvous in container backend

### LINKED ISSUE #167: Node rank assigned by the container backend is not used
### What happened?

When running distributed training jobs with the container backend, the assigned node rank is not taken into account and the container name index doesn't match the actual PyTorch distributed rank.

For example, this gives:
```
$ podman logs rb0f235b3dc2-node-0
Distributed Training for WORLD_SIZE: 4, RANK: 2, LOCAL_RANK: 0
```

The assigned node rank should be respected so the container name indices and the node ranks match.

### Environment

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow
Name: kubeflow
Version: 0.2.0
```

### Impacted by this bug?

astefanutti:

This changes the container backend to use PyTorch distributed static rendezvous otherwise `--node-rank` is not taken into account by torchrun:  

https://github.com/pytorch/pytorch/blob/1df723e6f5a09b2be2f3d69add2800280daad0aa/torch/distributed/run.py#L869

Fixes #167 

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -356,7 +356,7 @@ class ContainerBackend(RuntimeBackend):
                     f"TRAINING_SCRIPT_EOF\n"
                     "if command -v torchrun >/dev/null 2>&1; then "
                     f"  torchrun --nproc_per_node={nproc_per_node} --nnodes={num_nodes} "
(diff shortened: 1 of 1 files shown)


kramaranya:
@kubeflow/kubeflow-sdk-team do we want to cherry-pick this in 0.2.1?

PR #181, opened 2025-11-25
fix: update permissions for welcome workflow to avoid 403 error

### LINKED ISSUE #179: Welcome new contributors workflow fails with 403 when posting comments
[Welcome new contributors](https://github.com/kubeflow/sdk/blob/main/.github/workflows/welcome-new-contributors.yaml) workflow is consistently failing with a 403 Resource not accessible by integration error when it tries to post a comment on new pull requests. 

In the [logs](https://github.com/kubeflow/sdk/actions/runs/19667782076/job/56328907638), GITHUB_TOKEN shows only read access for issues and pull_requests even though the workflow requests write permissions

/kind bug

aniketpati1121:
This PR fixes the 403 "Resource not accessible by integration" error 
in the "Welcome new contributors" workflow.

Changes:
- Updated `contents` permission from `read` to `write`
- Ensured `pull_requests` and `issues` have proper write access

This allows the `actions/first-interaction` action to successfully post 
welcome comments on new issues and PRs. 

fixes #179

### DIFF
--- a/.github/workflows/welcome-new-contributors.yaml
+++ b/.github/workflows/welcome-new-contributors.yaml
@@ -1,7 +1,7 @@
 name: Welcome new contributors
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
   issues:


.github/workflows/welcome-new-contributors.yaml
```diff
@@ -15,6 +15,10 @@ permissions:
 jobs:
   welcome:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
```
andreyvelich: Thanks @aniketpati1121!
I think @kramaranya originally tried to use PR target, but it doesn't work: https://github.com/kubeflow/sdk/pull/82

Did we miss the `contents: write` permission?

aniketpati1121: > Thanks @aniketpati1121! I think @kramaranya originally tried to use PR target, but it doesn't work: #82
> 
> Did we miss the `contents: write` permission?

Thanks @andreyvelich! Yes, we added `contents: write` permission and updated the workflow to use `pull_request_target`, so it should now successfully post comments on new PRs and issues.

andreyvelich: Do you need to remove the permission at the Workflow level (e.g. L11-L13)?

aniketpati1121: We are only using job-level permissions under `welcome` job, so no workflow-level permissions are set. This should be sufficient for the action to post comments successfully.

andreyvelich: SGTM, can you remove these lines in your PR: https://github.com/kubeflow/sdk/blob/8672d5f1c979204b39c55caec785a7fb51659438/.github/workflows/welcome-new-contributors.yaml#L11-L14 ?

kramaranya: Hm, that might work. I'm thinking whether we should just add `contents: write` at the workflow level instead and keep other permissions there as well? Similar to https://github.com/HugoBlox/hugo-blox-builder/blob/0ef515fbaeeda95259fa77e817b4538e84bf6a9e/.github/workflows/community-welcome.yml#L12-L15 and https://github.com/actions/first-interaction/blob/main/README.md?plain=1#L39-L41

aniketpati1121: Please let me know if anything else is needed.

andreyvelich: @aniketpati1121 Can you move `contents: write` to the workflow level permission (e.g. L40)?

aniketpati1121: > @aniketpati1121 Can you move `contents: write` to the workflow level permission (e.g. L40)?

Updated the workflow by moving `contents: write` to the workflow-level permissions

Issue #50, opened 2023-01-10
[SDK] Generate Docs for Kubeflow SDK


andreyvelich:
Currently, [the Katib Client doc](https://github.com/kubeflow/katib/blob/master/sdk/python/v1beta1/docs/KatibClient.md) is outdated as I mentioned in this PR: https://github.com/kubeflow/katib/pull/2075.
Each API in the Katib Client has [the appropriate `docstring`](https://github.com/kubeflow/katib/blob/45a474432ad0e71db789c35564dc3ad763cae203/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py#L138-L169).
We should automate doc creation based on that description.

This issue is also applied for [the Training Client](https://github.com/kubeflow/training-operator/tree/master/sdk/python/kubeflow/training).

We need to add documentation on how to use KatibClient outside of Kubernetes cluster using Service Account token: https://github.com/kubeflow/katib/issues/2046#issuecomment-1691659428 

This is tracking issue for generate docs for Kubeflow SDK.

/area sdk

---

Love this feature? Give it a 👍 We prioritize the features with the most 👍

kramaranya:
Related to https://github.com/kubeflow/katib/pull/2415

andreyvelich:
We are looking for contributors to pick this up.

/good-first-issue
/area docs

andreyvelich:
Thank you for your interest. @NarayanaSabari!
Please can you explore these PRs: https://github.com/kubeflow/katib/pull/2415, https://github.com/kubeflow/katib/issues/2373.
Previously, we decided to use Sphinx to auto-generate docs similar to Model Registry or KFP (cc @kubeflow/wg-data-leads @kubeflow/wg-pipeline-leads).
Try to see how markdown files are stored for Model Registry or Kubeflow Pipelines SDKs.


>How should we integrate the generated docs with kubeflow.org? Should we:

Let's initially have a separate website, just for Kubeflow SDK, and re-evaluate it later.

[...]

PR #186, opened 2025-12-02
feat: add model registry client

### LINKED ISSUE #59: Add Model Registry Client to Kubeflow SDK

jonburdo:
Add an api for model registry under a new kubeflow.hub submodule.

This takes the form of a ModelRegistryClient class which provides a kubeflow-style api for model-registry.

Fixes #59 

**Checklist:**

## details

This provides a possible api following [this discussion](https://www.youtube.com/watch?v=UWCeLq6RtmI&t=1763s
). Also see the linked issue and the following:
- https://github.com/kubeflow/sdk/issues/59#issuecomment-3160865638
- https://github.com/kubeflow/sdk/pull/60


Makefile
```diff
@@ -74,7 +75,7 @@ release: install-dev
  # make test-python will produce html coverage by default. Run with `make test-python report=xml` to produce xml report.
 .PHONY: test-python
 test-python: uv-venv  ## Run Python unit tests
-	@uv sync
+	@uv sync --extra dev
```
jonburdo: needed for coverage

README.md
```diff
@@ -109,6 +109,55 @@ optimization_id = OptimizerClient().optimize(
...
+
+```python
+from kubeflow.hub import ModelRegistryClient
+
+# Connect to Model Registry
+client = ModelRegistryClient(
+    server_address="https://model-registry.kubeflow.svc.cluster.local",
+    port=443,
+    author="your-name"
+)
```
jonburdo: We pass the init args to the "developer/internal client" (`ModelRegistry`) and create a new instance automatically by default, so we don't expose the `model_registry` package to users at all. If a user already uses `model_registry` or needs access to some of the methods it offers, we provide as an "advanced" option `ModelRegistryClient.from_registry(registry)` so they are not forced to use init and have a new `ModelRegistry` instance created.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,285 @@
...
+        *,
+        author: str,
+        is_secure: bool = True,
+        user_token: str | None = None,
+        user_token_envvar: str | None = None,
+        custom_ca: str | None = None,
+        custom_ca_envvar: str | None = None,
+        log_level: int = logging.WARNING,
+        async_runner: Callable[[Coroutine[Any, Any, T]], T] | None = None,
+    ):
```
jonburdo, line 47: At the moment, this has the same `__init__` args as `ModelRegistry`, but thoughts on what to change would be great.

I'm not a huge fan of the keyword-only args, although it does offer flexibility to re-order args later without it being a breaking change.

I'm not sure we need all of these. For instance, `async_runner` can probably be removed at the least.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,285 @@
...
+        Raises:
+            ImportError: If model-registry is not installed.
+        """
+        try:
+            from model_registry import ModelRegistry
+        except ImportError as e:
+            raise ImportError(
+                "model-registry is not installed. Install it with:\n\n"  # fmt: skip
+                "  pip install 'kubeflow[hub]'\n"
+            ) from e
```
jonburdo, line 77: Note, the code currently allows the user to import `from kubeflow.hub import ModelRegistryClient` without error if `model-registry` is missing. It's not until `ModelRegistryClient()` is called that they'll get an error. I believe this makes sense.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,285 @@
...
+        model_source_kind: str | None = None,
+        model_source_class: str | None = None,
+        model_source_group: str | None = None,
+        model_source_id: str | None = None,
+        model_source_name: str | None = None,
+        author: str | None = None,
+        owner: str | None = None,
+        description: str | None = None,
+        metadata: Mapping[str, SupportedTypes] | None = None,
+    ) -> RegisteredModel:
```
jonburdo, line 106: Matches `ModelRegistry.register_model`. I'm not sure how I feel about the required keyword-only args.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,285 @@
...
+        Args:
+            model: The registered model to update. Must have an ID.
+
+        Returns:
+            Updated registered model.
+
+        Raises:
+            model_registry.exceptions.StoreError: If model does not have an ID.
+        """
+        return self._registry.update(model)
```
jonburdo, line 164: This `update_*` methods are effectively all aliases (aside from type-annotations) - maybe worth actually erroring if a user calls the wrong one? For example, we could throw an exception if a user passes a `ModelVersion` to `update_model` instead of to `update_model_version`

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,285 @@
...
+            name: Name of the model.
+            version: Version of the model.
+
+        Returns:
+            Model version or None if not found.
+
+        Raises:
+            model_registry.exceptions.StoreError: If the model does not exist.
+        """
+        return self._registry.get_model_version(name, version)
```
jonburdo: Do we want to make the return `ModelVersion` instead of `ModelVersion | None` and error if there is no matching model version? (Same for `get_model_artifact` if so)

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,144 @@
+"""Tests for ModelRegistryClient."""
+
+from __future__ import annotations
+
+from unittest.mock import Mock
+
+import pytest
+
+
+class TestModelRegistryClientInitialization:
```
jonburdo: I'm happy to take a different approach to testing if preferred. I generally don't love mocking this much, though this is just a thin wrapper

README.md
tarilabs: also if you wanted, I think we can omit specifying the port (where obvious) here like we do for the internal client

kubeflow/hub/api/model_registry_client.py
tarilabs: +1, and for completeness, we (MR WG) have implemented this async_runner "plug-in" so to be able to support end user request that are working within Ray environments

kubeflow/hub/api/model_registry_client.py
tarilabs: +1 thanks for following up on the original discussion in https://github.com/kubeflow/sdk/issues/59#issuecomment-3564188759

.github/workflows/check-pr-title.yaml
```diff
@@ -33,6 +33,7 @@ jobs:
             ci
             docs
             examples
+            hub
```
kramaranya, line 37: Sorry I'm still a bit confused with hub vs model registry, @johnugeorge would you mind elaborating on this, specifically what do we expose to Kubeflow SDK users? 
Shall this be model-registry or mr instead, since we only include only ModelRegistryClient?

pyproject.toml
```diff
@@ -39,15 +39,20 @@ docker = [
 podman = [
   "podman>=5.6.0"
 ]
+hub = [
+  "model-registry>=0.3.0",
+]
```
kramaranya, line 44: I think we should include model-registry by default. wdyt @kubeflow/kubeflow-sdk-team 
The issue is that Model Registry and other components already have their own extras, for example https://github.com/kubeflow/model-registry/blob/main/clients/python/pyproject.toml#L33-L36. We can't do "extras of extras", so if we tried to consolidate them we'd lose that functionality.
I think we should include core Kubeflow components by default (including model-registry) and reserve extras for third-party integrations like OpenTelemetry, MLflow, Hugging Face, etc.

pyproject.toml
```diff
@@ -39,15 +39,20 @@ docker = [
...
 
 [dependency-groups]
 dev = [
+  "kubeflow[hub]",
```
kramaranya, line 48:
```suggestion
  "model-registry>=0.3.0",
```

README.md
kramaranya: Makes sense

README.md
```diff
@@ -109,6 +109,55 @@ optimization_id = OptimizerClient().optimize(
...
+
+**Advanced: Use an existing ModelRegistry instance**
+```python
+from kubeflow.hub import ModelRegistryClient
+from model_registry import ModelRegistry
+
+# Custom ModelRegistry configuration
+registry = ModelRegistry(server_address="...", port=443, author="...", custom_ca="/path/to/ca.pem")
+client = ModelRegistryClient.from_registry(registry)
+```
```
kramaranya: I think we should just expose ModelRegistryClient here, and add advanced functionality to the Kubeflow website once we have it as part of https://github.com/kubeflow/sdk/issues/50

kubeflow/hub/api/model_registry_client.py
kramaranya: Agree, I'd like to keep it simple for start and we can always add more args if user needs. 
Do we need user_token_envvar, custom_ca_envvar, log_level, async_runner or can we drop it for now?

kubeflow/hub/api/model_registry_client.py
kramaranya: What is model_format_version and can we default it?

kubeflow/hub/api/model_registry_client.py
kramaranya: > This `update_*` methods are effectively all aliases (aside from type-annotations) - maybe worth actually erroring if a user calls the wrong one? For example, we could throw an exception if a user passes a `ModelVersion` to `update_model` instead of to `update_model_version`

SGTM

kramaranya (commented):
Thank you for this amazing work, @jonburdo!
Overall looks good, I left my initial comments

.github/workflows/check-pr-title.yaml
jonburdo: I'm thinking `kubeflow.hub` is the appropriate submodule given:
- We plan to rename from "Kubeflow Model Registry" to "Kubeflow Hub" assuming [KEP-907](https://github.com/kubeflow/community/pull/907) goes through.
- The scope of the project is already broader than just model registry and it is conceivable that we could offer other clients in the future.

To highlight the point, we already have a model catalog and some basic experiment tracking. Now we don't plan to have kubeflow sdk clients for these afaik, but we are positioned to be able to extend the project for other AI/ML asset types (mcp, prompt management). We also have some storage helpers and integrations as outlined [here](https://github.com/kubeflow/community/pull/907#issuecomment-3509842213).

Note my assumption here is basically that we want to have roughly one submodule per kubeflow sub-project. We could decide not to do this and then have `kubeflow.model_registry` and add separate submodules for any new clients or functionality we add later.

pyproject.toml
jonburdo: I think the takeaway in the Kubeflow SDK & ML Experience WG call was that we will keep `model-registry` optional for now. I'm thinking that means we leave this as is - lmk if not.

A few other things to note:
- The subset of model registry functions we are exposing here does not rely on the `model-registry` extras if I'm not mistaken
- This doesn't prevent users from doing `pip install 'model-registry[hf,boto3,olot]'` if needed
- Other options could include having `kubeflow[hub]` for `model-registry` and `kubeflow[hub-all]` for `model-registry[hf,boto3,olot]` if we really needed it (but I don't think we do)

pyproject.toml
jonburdo: I think we just want to include anything in the `hub` extra rather than re-listing the packages.

Makefile
```diff
@@ -53,6 +53,7 @@ verify: install-dev  ## install all required tools
 	@uv lock --check
 	@uv run ruff check --show-fixes --output-format=github .
 	@uv run ruff format --check kubeflow
+	@uv run mypy kubeflow/hub --exclude '_test\.py$$'
```
jonburdo: I've added mypy - for now just on `kubeflow/hub`. It would simplify linting if we used the typical pattern of a separate `test` dir in the root of the repo for tests instead of mixing in `*_test.py` files, but I stuck with the existing pattern for now.

README.md
```diff
@@ -109,6 +109,37 @@ optimization_id = OptimizerClient().optimize(
...
+from kubeflow.hub import ModelRegistryClient
+
+client = ModelRegistryClient("https://model-registry.kubeflow.svc.cluster.local", author="Your Name")
+
+# Register a model
+model = client.register_model(
+    name="my-model",
+    uri="s3://bucket/path/to/model",
+    version="v1.0.0",
+)
```
jonburdo, line 145: This is a minimal example of this function. It could make more sense to me to have `uri, name` rather than `name, uri`. This is also what `mlflow` does for `mlflow.register_model("uri", "name")` and feel more natural if we think of this as a copy or store operation (i.e `source, dest` or "register X as Y").

But it's probably more important to keep this consistent with `ModelRegistry`.

I'd also love for `version` to be auto-generated if not provided instead of required. Maybe something for the future.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,288 @@
...
+    def __init__(
+        self,
+        base_url: str,
+        port: int | None = None,
+        *,
+        author: str | None = None,
+        is_secure: bool | None = None,
+        user_token: str | None = None,
+        custom_ca: str | None = None,
+    ):
```
jonburdo, line 47: How does this look? Can we remove `is_secure`? I've changed so that `is_secure` and `port` are properly inferred if they are `None`

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,288 @@
...
+        model_source_kind: str | None = None,
+        model_source_class: str | None = None,
+        model_source_group: str | None = None,
+        model_source_id: str | None = None,
+        model_source_name: str | None = None,
+        author: str | None = None,
+        owner: str | None = None,
+        description: str | None = None,
+        metadata: Mapping[str, SupportedTypes] | None = None,
+    ) -> RegisteredModel:
```
jonburdo, line 106: I've give `model_format_*` args default values and move `version` up. `name, uri, version` are the only required args.

@tarilabs Can we consider any of these args non-critical and remove them? It can be overwhelming to have so many args (i.e. `uri` plus also `storage_key` and `storage_path`). I see we have some info [here](https://github.com/kubeflow/model-registry/blob/dbf79f7c6763ccc5d31341fa38fc8e36767180c9/docs/logical_model.md?plain=1#L47-L63)

It's probably okay as long as we introduce it with simple examples.

jonburdo (commented):
I've made some updates and simplifications since the first iteration:
- simplify `__init__` and `register_model`
- remove `from_registry` classmethod - We can have something that does this in `ModelRegistry` instead of here I think
- return `ModelVersion` instead of `ModelVersion | None` from `get_model_version` - same for other `get_*` functions. Error if missing

README.md
tarilabs: this minimal example however won't be enough to be Deploy-able on KServe tho ...

kubeflow/hub/api/model_registry_client.py
tarilabs: `is_secure` is needed when you want to skip TLS verification, ie you have self-signed certs in a local development environment

kubeflow/hub/api/model_registry_client.py
tarilabs: Let's walk through these step-by-step and indeed refer to the logical model you linked, keeping in mind the role of a/the Model Registry is also to keep the metadata needed in order to go deployment (with KServe)

- `model_format_name`
- `model_format_version`

the former is _required_ by KServe to know which ServingRuntime to be used, and the latter is _typically_ needed when some formats require to specify a version.

- `storage_key`
- `storage_path`
- `service_account_name`

more often than not, you need some credential to access the model (weights) wherever they have been stored, these maps the KServe mechanism to access the necessary "credentials" and is quite abstracted (away from KServe) being standard K8s mechanism to deal with these.

- `model_source_kind`
- `model_source_class`
- `model_source_group`
- `model_source_id`
- `model_source_name`

these were added on request from Kubeflow Pipelines to track origination (provenance) of the Model, see example usage in [this blog post](https://blog.kubeflow.org/fraud-detection-e2e/#4-model-registration). My only regret of that modification, I was not successful in advocating in grouping/consolidating them in a dataclass or the like (instead of the flat-list we have today).

The remainder is business as usual.

I'd summarize as: model_format_* are somehow still necessary if you want the complete e2e flow and cannot really take a default value; the rest assuming the storage is "public" or you don't need provenance can be skipped, or be "updated" by subsequent calls to "update_" methods?

README.md
jonburdo: I've added back the model_format args

tarilabs: thanks, this will help for cohesive e2e flows imho

kubeflow/hub/api/model_registry_client.py
jonburdo: Sounds good. I've defaulted `model_format_*` to `None`, removed `storage_*, service_account_name`, removed `model_source_*`, and kept the others: `name, uri, version, author, owner, description, metadata` (with the first three being required)

tarilabs: SGTM, thanks a lot Jon

jonburdo:
I've addressed initial feedback and believe this is ready for review by @kubeflow/kubeflow-sdk-team

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        return self._registry.register_model(
+            name=name,
+            uri=uri,
+            model_format_name=model_format_name,  # type: ignore[arg-type]
+            model_format_version=model_format_version,  # type: ignore[arg-type]
+            version=version,
+            author=author,
+            owner=owner,
+            description=version_description,
```
jonburdo, line 143: `description` really goes to `ModelVersion` not to `RegisteredModel` so it probably should have been call `version_description` from the beginning.

tarilabs: agree, and it's nice this becomes an internal impl. detail surfacing just what needed for the KF SDK a few lines above

Makefile
andreyvelich: Can we use `ruff` for linting ? Using unified tool will allow us to reduce number of dependencies users have to install.

README.md
```diff
@@ -109,6 +109,40 @@ optimization_id = OptimizerClient().optimize(
...
 
+### Manage models with Model Registry
+
+**Install Model Registry support:**
+```bash
+pip install 'kubeflow[hub]'
+```
+
+```python
+from kubeflow.hub import ModelRegistryClient
```
andreyvelich, line 133: Are we going to have 2 clients in Hub? ModelRegistry and Catalog.

pyproject.toml
andreyvelich: Yes, let's keep the `hub` extras for now as we discussed. Later we can figure out a way how to manage extras of extras.

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,198 @@
...
+    """Test ModelRegistryClient initialization."""
+
+    @pytest.fixture(autouse=True)
+    def skip_if_no_model_registry(self):
+        """Skip these tests if model-registry not installed."""
+        pytest.importorskip("model_registry")
+
+    def test_init_raises_helpful_error_when_not_installed(self, monkeypatch):
```
andreyvelich: Please can you refactor these tests to align with this: https://github.com/kubeflow/sdk/blob/854e6ba627635a51e32d278b4fbd13e9a8752504/kubeflow/trainer/backends/kubernetes/backend_test.py#L635

E.g. every test should align with the function name: `test_register_model(test_case)`  and we create test cases via :
```py
@pytest.mark.parametrize(
    "test_case",
    [
        TestCase()
    ]
```

That will allow contributors to easily understand code and help with changes for Hub SDKs.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
+from __future__ import annotations
+
+from collections.abc import Iterator, Mapping
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
```
andreyvelich, line 20: Why do we need type checking here?

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+    Requires the model-registry package to be installed. Install it with:
+
+        pip install 'kubeflow[hub]'
+
+    """
+
+    def __init__(
+        self,
+        base_url: str,
+        port: int | None = None,
```
andreyvelich, line 41: Please Union instead of `|` for now, since we still support Python 3.9 for our users.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+    ):
+        """Initialize the ModelRegistryClient.
+
+        Args:
+            base_url: Base URL of the model registry server including scheme.
+                     Examples: "https://registry.example.com", "http://localhost"
+                     The scheme is used to infer is_secure and port if not explicitly provided.
+
+        Keyword Args:
+            port: Server port. If not provided, inferred from base_url scheme:
```
andreyvelich, line 56: Could we parse the port from the base_url if provided or you think it is less explicit?
e.g.
```
https://registry.example.com:1234
```

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+                 - https:// defaults to 443
+                 - http:// defaults to 8080
+                 - no scheme defaults to 443
+            author: Name of the author.
+            is_secure: Whether to use a secure connection. If not provided, inferred from base_url:
```
andreyvelich, line 61: Same question, why not check it from the URL directly ?

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+        is_http = base_url.startswith("http://")
+        if is_secure is None:
+            is_secure = not is_http
+        if port is None:
+            port = 8080 if is_http else 443
+
+        self._registry = ModelRegistry(
+            server_address=base_url,
+            port=port,
+            author=author,  # type: ignore[arg-type]
```
andreyvelich, line 88: Since author is part of register_model API, does client still require it: https://github.com/kubeflow/sdk/blob/854e6ba627635a51e32d278b4fbd13e9a8752504/kubeflow/hub/api/model_registry_client.py#L88 ?

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+        connection-specific parameters, `storage_key` and `storage_path` or,
+        simply a `service_account_name`. URI builder utilities are recommended
+        when referring to specialized storage; for example `utils.s3_uri_from`
+        helper when using S3 object storage data connections.
+
+        Args:
+            name: Name of the model.
+            uri: URI of the model.
+
+        Keyword Args:
```
andreyvelich, line 122: @kramaranya Do you know if we can auto-generate Kubeflow SDK docs from such format of docstring: Keyword Args?

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+            user_token=user_token,
+            custom_ca=custom_ca,
+        )
+
+    def register_model(
+        self,
+        name: str,
+        uri: str,
+        *,
+        version: str,
```
andreyvelich, line 99: Since version is mandatory parameter why you put it after "*" as keyword args?

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+            model: The registered model to update. Must have an ID.
+
+        Returns:
+            Updated registered model.
+
+        Raises:
+            TypeError: If model is not a RegisteredModel instance.
+            model_registry.exceptions.StoreError: If model does not have an ID.
+        """
+        from model_registry.types import RegisteredModel
```
andreyvelich, line 160: Can we move all imports to the top of this file?

README.md
tarilabs: We have been [discussing in KF Hub biweekly meeting](https://youtu.be/mG27J_MSiK8?si=yb3oDrzWYS6PKTZm&t=1001) that any new _user facing_ client will be developed in KF SDK.
👉 So when the user need comes for a _user facing_ py client for the Catalog, we'd directly develop this in KF SDK.
Hope this is satisfactory answer for ML WG (this group) and from KSC pov.

Detail note: from a E2E testing point of view, we need python based code generated clients for _all_ the endpoints, which is something that for example ML WG (this group) mentioned is not interested in exposing, etc. So infrastructure code we need for the purpose of E2E, fuzz, and coverage testing of the endpoints, we're continuing maintaining in the [KF/model-registry repo](https://github.com/kubeflow/model-registry). We're open to revisit this setup too.

Makefile
jonburdo: We are using `ruff` for linting, but it doesn't include type-checking, which we really want, so we **also** need mypy (e.g. if we're passing an `int` to `str` arg ruff won't catch that). The same org behind `ruff` is working on a type checker, `ty`, but it's relatively new and would still be a separate tool: https://github.com/astral-sh/ty

kubeflow/hub/api/model_registry_client.py
jonburdo: We want type annotations on the functions for better type-safety. LSPs will use them too (docs/checks in code editors). The `if TYPE_CHECKING` allows us to import things we may not have a runtime. mypy and LSPs will use the imports but if `model-registry` is missing we don't get an import error at this point.

Since `hub` is an optional dependency, we check for `model-registry` upon usage of the client `ModelRegistryClient()` rather than upon import as discussed:
https://github.com/kubeflow/sdk/pull/186#discussion_r2583177514

Otherwise in a new env, this would error:
```
pip install kubeflow
python -c 'import kubeflow.hub'
```

kubeflow/hub/api/model_registry_client.py
jonburdo: I've added `from __future__ import annotations` at the top in order to allow [postponed evaluation of annotations](https://peps.python.org/pep-0563/). This allows things like:
- `|` for union
- missing-at-runtime variables like return type `-> RegisteredModel` instead of ` -> "RegisteredModel"`

I've double checked this works on python 3.9 by ensuring `from kubeflow.hub import ModelRegistryClient` works as-is. Removing the `from __future__ import annotations` would result in:
```
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

kubeflow/hub/api/model_registry_client.py
jonburdo: Good idea. Maybe we can do this as a follow-up?

I'm thinking we should keep the option to use the separate `port` arg, but if it's not used, we should also allow users to pass it via `base_url`, so these are all valid and equivalent:
```python
ModelRegistryClient(base_url="https://example.com")
ModelRegistryClient(base_url="https://example.com", port=443)
ModelRegistryClient(base_url="https://example.com:443")
```
This can also be a change made in the base developer client (`ModelRegistry` from `model-registry` package) instead of here.

README.md
andreyvelich: Great, thank you for sharing recording @tarilabs!

>we need python based code generated clients for all the endpoints, which is something that for example ML WG (this group) mentioned is not interested in exposing, etc.

Yeah, that make sense. I guess, the conclusion was that we expose minimal set of APIs via `kubeflow.hub` package, and gather initial users feedback.
If users need other APIs that we don't expose via `kubeflow.hub`, they can always use it via this parameter:
```
self._registry()
```
Or import the client directly `import model_registry`.
In the future, we can potentially add more APIs to the `kubeflow.hub` directly.


The only question that I have, do users need two separate clients for Catalog and Model Registry APIs inside the `kubeflow.hub` sub package ?
e.g.:
```
from kubeflow.hub import ModelRegistryClient
from kubeflow.hub import CatalogClient
or just
from kubeflow.hub import HubClient
```

tarilabs: > Great, thank you for sharing recording @tarilabs!
> 
> > we need python based code generated clients for all the endpoints, which is something that for example ML WG (this group) mentioned is not interested in exposing, etc.
> 
> Yeah, that make sense. I guess, the conclusion was that we expose minimal set of APIs via `kubeflow.hub` package, and gather initial users feedback. If users need other APIs that we don't expose via `kubeflow.hub`, they can always use it via this parameter:
> 
> ```
> self._registry()
> ```
> 
> Or import the client directly `import model_registry`. In the future, we can potentially add more APIs to the `kubeflow.hub` directly.

Great, sounds we are all aligned, personally glad and great to hear 👍 

As for the latter question, my current assessment is:

```
from kubeflow.hub import ModelRegistryClient
from kubeflow.hub import CatalogClient
```

as they're different patterns, and in turn different use-cases.
However as we'll define it in more concrete terms, we can refine the case more specifically then.
Hope this is understandable and fair.

andreyvelich: Sure, that sounds great, thank you!

kubeflow/hub/api/model_registry_client.py
jonburdo: https://github.com/kubeflow/sdk/pull/186#discussion_r2587958905
> is_secure is needed when you want to skip TLS verification, ie you have self-signed certs in a local development environment

We will default to `is_secure=True` for https and `is_secure=False` for http, but I believe we need to allow users to disable it too (`is_secure=False` with https) for development (shouldn't be done in production).

kubeflow/hub/api/model_registry_client.py
jonburdo: This provides a default value so it doesn't have to be passed to `register_model` every time it's called.

kubeflow/hub/api/model_registry_client.py
jonburdo: It matches what we have in `ModelRegistry` and provides a little more flexibility if we want to re-order args after `name, uri`. I do wonder if we want to make `version` optional in the future and support an autogenerated version string if not provided.

kubeflow/hub/api/model_registry_client.py
jonburdo: This would cause an immediate error upon import for `import kubeflow.hub` if the user has done:
```
pip install kubeflow
```
but not
```
pip install 'kubeflow[hub]'
```
We could opt for that. Or we could do a lazy import. I would suggest leaving it as-is though.

kubeflow/hub/api/model_registry_client_test.py
jonburdo: Sure, I've made this change. This does we are importing `TestCase` from `kubeflow.trainer`. I also have mixed feelings about this, but if it's something we want to standardize in the sdk, then I guess it works.

jonburdo (commented):
Thanks @andreyvelich for the review and all the thoughtful feedback! I believe I've addressed it. The one thing I'm not sure about is the question about docs: https://github.com/kubeflow/sdk/pull/186#discussion_r2676680508

Makefile
andreyvelich: @kramaranya @szaher @astefanutti What are your thoughts on this? Shall we adopt `ty` for type checking?
@HumairAK @mprahl @hbelmiro @nsingla Are you planning to use any typing tool for Pipelines library?

kubeflow/hub/api/model_registry_client_test.py
andreyvelich: Thank you! Yes, eventually, we should have common types/helper function that we can re-use across packages.
This is yet another benefit to use common SDK to reduce code duplication and standardize on the code practices.

kubeflow/hub/api/model_registry_client.py
andreyvelich: Sounds great! We will soon deprecate Python 3.9, so we should eventually migrate all SDK code to this new structure.

kubeflow/hub/api/model_registry_client.py
andreyvelich: I think standardizing on `base_url` would make ModelRegistryClient consistent with SparkClient
https://github.com/kubeflow/sdk/blob/f140fdc9a73c45423e8e7f9a9f14bc71558e027c/kubeflow/spark/api/spark_client.py#L47

@Shekharrajak We might want to use `base_url` instead of `url` in the `connect()` API too.

Having a consistent experience across APIs is likely what users expect and will reduce confusion.


>This can also be a change made in the base developer client (ModelRegistry from model-registry package) instead of here.

The changes in the ModelRegistry client can come later if needed.

kubeflow/hub/api/model_registry_client.py
andreyvelich: If it is only for development why not use `http`? Did you have use-cases when users want to verify `https` locally without secure connection?

kubeflow/hub/api/model_registry_client.py
andreyvelich: I think, this is good idea. I would suggest to make it Optional.

kubeflow/hub/api/model_registry_client.py
andreyvelich: But since we put it under `TYPE_CHECKING` that should be fine, isn't?
https://github.com/kubeflow/sdk/blob/7f41fa7aa5fe7cb38b18c86f3b54d0fede3c07ae/kubeflow/hub/api/model_registry_client.py#L6

.github/workflows/check-pr-title.yaml
kramaranya: I see, thanks @jonburdo!

Makefile
kramaranya: @jonburdo have you tried to compare ty and mypy? does it cover everything you're using mypy for?

kubeflow/hub/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Kubeflow Hub API."""
```
kramaranya: Can you add copyright header please to all new files?

kubeflow/hub/api/model_registry_client.py
kramaranya: I agree with @andreyvelich, this will keep our APIs consistent for users

kubeflow/hub/api/model_registry_client.py
kramaranya: Yes, I think it's configurable. I'll soon open a PR for the website and will verify it

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,265 @@
...
+            raise ValueError(f"Model artifact not found for model {name!r} version {version!r}")
+        return artifact
+
+    def list_models(self) -> Iterator[RegisteredModel]:
+        """Get an iterator for registered models.
+
+        Yields:
+            Registered models.
+        """
+        yield from self._registry.get_registered_models()
```
kramaranya, line 265: We return an Iterator in Model Registry with `list_models`, however, we return lists for both OptimizerClient and TrainerClient in APIs like `list_jobs()`. 
@andreyvelich @jonburdo to provide consistent APIs to our users, shall we consider aligning OptimizerClient and TrainerClient with ModelRegistryClient and return an Iterator, alternatively we can return list in ModelRegistryClient?

andreyvelich: I feel like returning the list might be a good option, with future improvements to return Iterator if users will request it.

@tarilabs @jonburdo Do you see a lot of use-cases when this doesn't work?
I don't expect that users manage more than millions of models in their registries.

kubeflow/hub/api/model_registry_client.py
Shekharrajak: Ack, we need to brainstorm some good way to restrict new development to follow these consistent APIs.

Makefile
hbelmiro: @andreyvelich No concrete plans at the moment. For the components/pipelines themselves, KFP doesn't really enforce or leverage Python type hints.
Adding a type checker for our scripts/CLIs could still be valuable. Though I'm guessing that's not the part you're asking about.

jonburdo: Looking again at `ty`, I think it's a good option. It will be faster and align with our use of `uv` and `ruff`. Replacing mypy with `ty`

kubeflow/hub/__init__.py
jonburdo: Good catch! Added

kubeflow/hub/api/model_registry_client.py
jonburdo: In this function `if not isinstance(model, RegisteredModel):` needs to have `RegisteredModel` at runtime.

For example, if you use `from __future__ import annotations` or python>=3.10 then you can use:
```python
if TYPE_CHECKING:
    ... import NonRuntimeVar

my_var: NonRuntimeVar = ...
```
`mypy` and `ty` will still analyze use of `NonRuntimeVar` here, but you couldn't actually do:
```python
NonRuntimeVar()
# or
isinstance(something, NonRuntimeVar)
```

kubeflow/hub/api/model_registry_client.py
jonburdo: Discussed in the community call today. We definitely want to keep this as a generator.

We use pagination, so eagerly fetching and returning a `list` would make extra API calls, and it is certainly possible to have enough entries for significant overhead.

I'm not sure I'd consider it real inconsistency if some functions in a library return a list and some return a generator - especially if it's done in a logical way based on amount of data or usefulness of evaluating lazily.

kubeflow/hub/api/model_registry_client.py
jonburdo: We discussed in the community call today. To confirm, should I leave the `ModelRegistryClient` arg as `base_url`? Or did we want to converge on `url`?

kubeflow/hub/api/model_registry_client.py
jonburdo: Discussed in community call today. @tarilabs noted that this is important for certain use-cases. Users may use this library for development and have a need to use unsigned certificates.

Fwiw, we do leave this out of our primary quick-start examples in the README

kubeflow/hub/api/model_registry_client.py
jonburdo: Discussed in the community meeting. I believe we agreed to leave it as is for now. Apparently it becomes complicated if we try to provide a default as users won't always agree on if it should be uuid or incrementing v1, v2, ...

kubeflow/hub/api/model_registry_client.py
andreyvelich: Oh, I see, thanks for clarifying!

kubeflow/hub/api/model_registry_client.py
andreyvelich: Let's use base_url everywhere including SparkClient.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -0,0 +1,279 @@
...
+        author: str | None = None,
+        is_secure: bool | None = None,
+        user_token: str | None = None,
+        custom_ca: str | None = None,
+    ):
+        """Initialize the ModelRegistryClient.
+
+        Args:
+            base_url: Base URL of the model registry server including scheme.
+                     Examples: "https://registry.example.com", "http://localhost"
```
andreyvelich, line 52: Please can you add example with port as well? e.g. https://registry.example.com:8888

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,442 @@
...
+    except ImportError as e:
+        assert test_case.expected_status == FAILED
+        assert "pip install 'kubeflow[hub]'" in str(e)
+
+
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="http URL infers port=8080 and is_secure=False",
```
andreyvelich, line 93: Could we add a test case when port is taken from the base_url?
```py
"base_url": "http://localhost:1234",
```

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,442 @@
...
+        TestCase(
+            name="get_model delegates to get_registered_model",
+            expected_status=SUCCESS,
+            config={
+                "name": "test-model",
+            },
+        ),
+    ],
+)
+def test_get_model(test_case, client, mock_registry):
```
andreyvelich, line 199: Could we add tests for `test_get_model_version` and `test_get_model_artifact` ?

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,442 @@
...
+            expected_status=FAILED,
+            config={
+                "wrong_type": "ModelArtifact",
+                "name": "artifact",
+            },
+            expected_error=TypeError,
+        ),
+    ],
+)
+def test_update_model_type_check(test_case, client):
```
andreyvelich: Can you add all of these test cases to the `test_update_model` ? Similar to other tests, we can have one function with test cases for each public API we expose.

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,442 @@
...
+            name="update_model_version raises TypeError for RegisteredModel",
+            expected_status=FAILED,
+            config={
+                "name": "model",
+            },
+            expected_error=TypeError,
+        ),
+    ],
+)
+def test_update_model_version_type_check(test_case, client):
```
andreyvelich: Same suggestion to move them to the `test_update_model_version`

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -0,0 +1,442 @@
...
+            name="update_model_artifact raises TypeError for RegisteredModel",
+            expected_status=FAILED,
+            config={
+                "name": "model",
+            },
+            expected_error=TypeError,
+        ),
+    ],
+)
+def test_update_model_artifact_type_check(test_case, client):
```
andreyvelich: Same here.

README.md
```diff
@@ -109,6 +109,40 @@ optimization_id = OptimizerClient().optimize(
 print(f"OptimizationJob created: {optimization_id}")
 ```
 
+### Manage models with Model Registry
+
```
andreyvelich, line 124: Can we also point users to the Model Registry documentation which says how to install control plane: https://www.kubeflow.org/docs/components/model-registry/installation/#standalone-installation

In the future, we might also want to introduce public ConfigMap check that verifies that Kubeflow Hub control plane is available on the cluster: https://github.com/kubeflow/sdk/issues/221
We can do this after the initial release.

kubeflow/hub/api/model_registry_client.py
jonburdo: I think we want to fix this in the developer (`model-registry`) client. I've created an issue for this: https://github.com/kubeflow/model-registry/issues/2110

Otherwise we'll have url parsing here and then again there (in `ModelRegistry`)

andreyvelich: Sounds good @jonburdo!
Please can you create an issue in `kubeflow/sdk` to also track this update once Model Registry client supports it upstream?

jonburdo: Added here: https://github.com/kubeflow/sdk/issues/227

kramaranya:
@jonburdo feel free to add OWNERS file to https://github.com/jonburdo/sdk/tree/kubeflow-hub-api/kubeflow/hub with Model Registry WG.
cc @andreyvelich

Issue #171, opened 2025-11-21
Add support for dataset and model initializers in container backend


Fiona-Waters:
### What you would like to be added?

Kubeflow Trainer uses [Initializers](https://github.com/kubeflow/trainer/tree/master/pkg/initializers) to pull datasets and models required for TrainJobs. These initialializers pull the required dataset and/or model into a shared volume prior to starting the TrainJob nodes. We should add support for this in the container backend.

### Why is this needed?

This is needed to bring the container backend implementation in line with the kubernetes backend.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

HKanoje:
Hi @Fiona-Waters ! I'm interested in working on this issue. I've analyzed the Kubernetes backend implementation and have a plan to add initializer support to the container backend following the same patterns.

My approach would be to:

- Create separate initializer containers that run before training starts
- Add utility functions for building initializer commands/env vars
- Ensure the implementation is consistent with the Kubernetes backend

Happy to discuss the approach!

HKanoje:
Sure thing! Thanks @Fiona-Waters 

/assign

immortal71:
@Fiona-Waters 
same here ,I’m currently working on other assigned issues, so I won’t be able to take this one right now.
Please feel free to assign it to someone else.

Thanks again for the support!

PR #188, opened 2025-12-04
feat(trainer): add dataset and model initializer support to container backend

HKanoje:

This PR implements dataset and model initializer support in the container backend, bringing it to feature parity with the Kubernetes backend. This addresses issue #171 by enabling users to automatically download and prepare datasets and models before training starts.

## Solution Overview

This implementation adds full initializer support to the container backend by:

1. **Running initializer containers before training** - Initializers execute sequentially (dataset first, then model) before any training containers start
2. **Using shared volumes** - All containers (initializers and training nodes) share the same workspace directory on the host
3. **Proper lifecycle management** - Initializers must complete successfully before training begins, with automatic cleanup on failures
4. **Comprehensive error handling** - Clear error messages and proper resource cleanup when initialization fails

#### `build_initializer_command(initializer, init_type)`
Builds the appropriate container command based on initializer type:
- **HuggingFace**: Uses `kubeflow.storage_initializer.hugging_face` module
- **S3**: Uses `kubeflow.storage_initializer.s3` module  
- **DataCache**: Uses `kubeflow.storage_initializer.datacache` module

#### `build_initializer_env(initializer, init_type)`
Constructs environment variables from initializer configuration:
- Sets `STORAGE_URI` from the initializer config
- Sets `OUTPUT_PATH` to `/workspace/dataset` or `/workspace/model` based on type
- Adds optional fields like `ACCESS_TOKEN`, `ENDPOINT`, `REGION`, etc.
- Handles DataCache-specific variables like `CLUSTER_SIZE`, `METADATA_LOC`

#### `get_initializer_image()`
Returns the initializer container image (`kubeflow/training-operator:latest`).
This can be made configurable via backend config in future iterations.

#### `_run_initializers(job_name, initializer, workdir, network_id)`
Orchestrates the initialization phase:
- Pulls the initializer image if needed (respects `pull_policy`)
- Runs dataset initializer if configured
- Runs model initializer if configured
- Ensures proper sequencing and error propagation

#### `_run_single_initializer(job_name, initializer_config, init_type, image, workdir, network_id)`
Executes a single initializer container:
- Creates container with proper labels for tracking
- Mounts shared volume to `/workspace`
- Monitors container status with configurable timeout (default 10 minutes)
- Waits for successful completion (exit code 0)
- Captures and reports logs on failure
- Cleans up failed containers automatically

#### Updated `train()` method
- Creates network and working directory first
- Runs initializers before generating training script
- Only proceeds to training if initialization succeeds
- Maintains backward compatibility (initializers are optional)

#### Updated `__get_trainjob_from_containers()`
- Correctly counts only training nodes for `num_nodes` (excludes initializers)
- Includes initializer containers in the steps list
- Properly tracks initializer status

#### Updated `get_job_logs()`
- Supports fetching logs from specific initializer steps
- When requesting `node-0` logs (default), only shows training container logs
- Can explicitly request initializer logs with `step="dataset-initializer"` or `step="model-initializer"`

### 3. Comprehensive Test Coverage (`kubeflow/trainer/backends/container/backend_test.py`)

Added 11 new test cases covering:

#### Initialization Success Scenarios
- **HuggingFace dataset initializer** - Tests storage_uri parsing and access_token handling
- **S3 dataset initializer** - Tests endpoint, region, and credential configuration
- **HuggingFace model initializer** - Tests model downloads with ignore_patterns
- **S3 model initializer** - Tests S3-compatible storage with authentication
- **Both dataset and model** - Tests sequential execution of both initializers
- **DataCache initializer** - Tests distributed cache configuration with metadata_loc and cluster size

#### Log Retrieval
- Tests getting logs from dataset-initializer step
- Tests getting logs from model-initializer step  
- Tests that default node logs exclude initializer logs

#### Error Handling
- **Non-zero exit code** - Verifies proper error reporting when initializer fails
- **Timeout handling** - Ensures timeout errors are caught and reported
- **Resource cleanup** - Confirms containers and networks are cleaned up on failure

### Initialization Flow
1. User calls `train()` with `initializer` parameter

2. ContainerBackend creates:
```
   ├── Working directory: ~/.kubeflow/trainer/containers/<job-name>/
   └── Network: <job-name>-net
```

3. If `initializer.dataset` is set:
```
   ├── Create dataset-initializer container
   ├── Mount workdir to /workspace
   ├── Run initialization (downloads to /workspace/dataset)
   └── Wait for completion (exit code 0)
```

4. If `initializer.model` is set:
```
   ├── Create model-initializer container
   ├── Mount workdir to /workspace
   ├── Run initialization (downloads to /workspace/model)
   └── Wait for completion (exit code 0)
```

5. Create training containers:
```
   ├── Mount same workdir to /workspace
   └── Access data at /workspace/dataset and /workspace/model
```

### Volume Layout

**Host:** `~/.kubeflow/trainer/containers/<job-name>/`
```
├── dataset/ (from dataset-initializer, if configured)
│   └── <downloaded dataset files>
├── model/ (from model-initializer, if configured)
│   └── <downloaded model files>
└── outputs/ (accessible to all training nodes)
    └── <training outputs>
```

**Container Mount:** `/workspace/`
```
├── dataset/ (read by training code)
├── model/ (read by training code)
└── outputs/ (written by training code)
```

## Testing Results

All tests pass with no regressions:

- [x] 43/43 container backend tests passed
- [x] 173/173 trainer module tests passed
- [x] make verify (ruff lint + format) passed
- [x] pre-commit hooks passed

## Usage Example
```python
from kubeflow.trainer.types import (
    CustomTrainer,
    Initializer,
    HuggingFaceDatasetInitializer,
    HuggingFaceModelInitializer,
)

# Define initializers
initializer = Initializer(
    dataset=HuggingFaceDatasetInitializer(
        storage_uri="hf://username/my-dataset",
        access_token="hf_xxxxx",
    ),
    model=HuggingFaceModelInitializer(
        storage_uri="hf://username/my-model",
        access_token="hf_xxxxx",
    ),
)

# Define trainer
trainer = CustomTrainer(
    func=train_func,
    num_nodes=2,
)

# Train with initializers
client = TrainerClient()
job_name = client.train(
    trainer=trainer,
    initializer=initializer,  # <-- Now works with container backend!
)

# Data is automatically available at:
# - /workspace/dataset (in training containers)
# - /workspace/model (in training containers)
```

### DIFF
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -540,7 +540,6 @@ class ContainerBackend(RuntimeBackend):
             self._run_single_initializer(
                 job_name=job_name,
                 container_init=dataset_init,
-                init_type="dataset",
                 workdir=workdir,
                 network_id=network_id,
             )
@@ -555,7 +554,6 @@ class ContainerBackend(RuntimeBackend):
             self._run_single_initializer(
                 job_name=job_name,
                 container_init=model_init,
-                init_type="model",
                 workdir=workdir,
                 network_id=network_id,
             )
@@ -565,7 +563,6 @@ class ContainerBackend(RuntimeBackend):
         self,
         job_name: str,
         container_init: container_utils.ContainerInitializer,
-        init_type: str,
         workdir: str,
         network_id: str,
     ):
@@ -574,20 +571,19 @@ class ContainerBackend(RuntimeBackend):
 
         Args:
             job_name: Name of the training job.
-            container_init: ContainerInitializer with image, command, and env.
-            init_type: Type of initializer ("dataset" or "model").
+            container_init: ContainerInitializer with name, image, command, and env.
             workdir: Working directory path on host.
             network_id: Network ID for containers.
 
         Raises:
             RuntimeError: If initializer fails.
         """
-        container_name = f"{job_name}-{init_type}-initializer"
+        container_name = f"{job_name}-{container_init.name}"
 
         # Create labels for tracking
         labels = {
             f"{self.label_prefix}/trainjob-name": job_name,
-            f"{self.label_prefix}/step": f"{init_type}-initializer",
+            f"{self.label_prefix}/step": container_init.name,
             f"{self.label_prefix}/network-id": network_id,
         }
 
@@ -599,7 +595,7 @@ class ContainerBackend(RuntimeBackend):
             }
         }
 
-        logger.debug(f"Starting {init_type} initializer container: {container_name}")
+        logger.debug(f"Starting {container_init.name} container: {container_name}")
 
         # Create and start the initializer container
         # The initializer images use /app as their working directory
@@ -625,7 +621,7 @@ class ContainerBackend(RuntimeBackend):
             )
 
             if exit_code == 0:
-                logger.debug(f"{init_type} initializer completed successfully")
+                logger.debug(f"{container_init.name} completed successfully")
                 # Clean up the successful container
                 self._cleanup_container_resources(container_ids=[container_id], stop_timeout=0)
                 return
@@ -633,14 +629,14 @@ class ContainerBackend(RuntimeBackend):
                 # Get logs for debugging
                 logs = list(self._adapter.container_logs(container_id, follow=False))
                 error_msg = (
-                    f"{init_type} initializer failed with exit code {exit_code}. "
+                    f"{container_init.name} failed with exit code {exit_code}. "
                     f"Logs: {' '.join(logs[-10:]) if logs else 'No logs available'}"
                 )
                 raise RuntimeError(error_msg)
 
         except TimeoutError:
             logger.error(
-                f"{init_type} initializer did not complete within "
+                f"{container_init.name} did not complete within "
                 f"{self.cfg.initializer_timeout} seconds"
             )
             # Clean up the timed-out container
@@ -648,7 +644,7 @@ class ContainerBackend(RuntimeBackend):
             raise
 
         except Exception as e:
-            logger.error(f"Error running {init_type} initializer: {e}")
+            logger.error(f"Error running {container_init.name}: {e}")
             # Clean up the failed container
             self._cleanup_container_resources(container_ids=[container_id], stop_timeout=5)
             raise
--- a/kubeflow/trainer/backends/container/utils.py
+++ b/kubeflow/trainer/backends/container/utils.py
@@ -218,6 +218,7 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
 class ContainerInitializer:
     """Internal type for container initializer configuration."""
 
+    name: str
     image: str
     command: list[str]
     env: dict[str, str]
@@ -280,6 +281,7 @@ def get_dataset_initializer(dataset: types.BaseInitializer, config) -> Container
     env.update(get_optional_initializer_envs(dataset, required_fields={"storage_uri"}))
 
     return ContainerInitializer(
+        name="dataset-initializer",
         image=config.dataset_initializer_image,
         command=["bash", "-c", "python -m pkg.initializers.dataset"],
         env=env,
@@ -313,6 +315,7 @@ def get_model_initializer(model: types.BaseInitializer, config) -> ContainerInit
     env.update(get_optional_initializer_envs(model, required_fields={"storage_uri"}))
 
     return ContainerInitializer(
+        name="model-initializer",
         image=config.model_initializer_image,
         command=["bash", "-c", "python -m pkg.initializers.model"],
         env=env,


kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,102 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+def get_initializer_image() -> str:
+    """
+    Get the container image for initializers.
+
+    Returns:
+        Container image name for initializers.
+    """
+    # Use the training-operator image which contains initializer scripts
+    # This can be made configurable via backend config in the future
+    return "kubeflow/training-operator:latest"
```
Fiona-Waters: Should we make this configurable rather than hardcoding it?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -467,6 +473,156 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+            working_dir=constants.WORKSPACE_PATH,
+        )
+
+        logger.debug(f"Initializer container started: {container_id[:12]}")
+
+        # Wait for the initializer to complete
+        try:
+            import time
+
+            timeout = 600  # 10 minutes timeout for initialization
```
Fiona-Waters: Should this be configurable, or is 10 minutes always going to be enough time?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -467,6 +473,156 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+                elapsed += polling_interval
+
+            raise TimeoutError(f"{init_type} initializer did not complete within {timeout} seconds")
+
+        except Exception as e:
+            logger.error(f"Error running {init_type} initializer: {e}")
+            # Clean up the failed container
+            from contextlib import suppress
+
+            with suppress(Exception):
```
Fiona-Waters: As well as cleaning up when a failure occurs, should we clean up the initializer containers when they have been successful also?

kubeflow/trainer/backends/container/backend.py
```diff
@@ -257,6 +257,26 @@ def train(
...
+                labels={
+                    f"{self.label_prefix}/trainjob-name": trainjob_name,
+                    f"{self.label_prefix}/runtime-name": runtime.name,
+                    f"{self.label_prefix}/workdir": workdir,
+                },
+            )
+            logger.debug(f"Created network: {network_id}")
+
+            # Run initializers if configured
+            if initializer:
```
Fiona-Waters, line 310: If the initializer fails should we clean up the network we have created?

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,102 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+    # Use the training-operator initializer script
+    # The initializer script is expected to be available in the image
+    python_cmd = (
+        "python -m kubeflow.storage_initializer.s3 "
+        if isinstance(initializer, (types.S3DatasetInitializer, types.S3ModelInitializer))
+        else "python -m kubeflow.storage_initializer.hugging_face "
+        if isinstance(
+            initializer, (types.HuggingFaceDatasetInitializer, types.HuggingFaceModelInitializer)
+        )
+        else "python -m kubeflow.storage_initializer.datacache "
```
Fiona-Waters: Here we are setting datacache as the default/fallback, do we want to do this? In the[kubernetes backend](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/utils.py) we offer 2 options and raise a value error if the type is invalid.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -467,6 +473,156 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+        )
+
+        logger.debug(f"Initializer container started: {container_id[:12]}")
+
+        # Wait for the initializer to complete
+        try:
+            import time
+
+            timeout = 600  # 10 minutes timeout for initialization
+            polling_interval = 2
```
Fiona-Waters: Could using wait API be supported?

Fiona-Waters (commented):
Thanks @HKanoje for this PR! I've left some comments, ptal! Thanks

kubeflow/trainer/backends/container/utils.py
HKanoje: Yes, Definitely. I've made this configurable via `ContainerBackendConfig.initializer_image` (default: `kubeflow/training-operator:latest`). Users can now customize it when creating the backend.

kubeflow/trainer/backends/container/backend.py
HKanoje: Added `ContainerBackendConfig.initializer_timeout` (default: 600 seconds / 10 minutes). This gives users flexibility for large datasets/models that may take longer to download.

kubeflow/trainer/backends/container/backend.py
HKanoje: Implemented `wait_for_container()` in the adapter interface and both Docker/Podman adapters. This replaces the polling loop with a single blocking wait call - much more efficient.

kubeflow/trainer/backends/container/backend.py
HKanoje: Absolutely! Added cleanup for successful initializer containers after completion to prevent accumulation. Also added cleanup for timed-out containers.

kubeflow/trainer/backends/container/utils.py
HKanoje: You're right - that was inconsistent with the kubernetes backend. Changed to raise `ValueError` with a clear message listing all supported types instead of defaulting to datacache.

HKanoje:
@kramaranya Done! All commits are now signed.

kubeflow/trainer/backends/container/backend.py
Fiona-Waters: Great! Would it make sense to add a helper function for the cleanup logic to reduce duplication?

HKanoje: Done! Added `_cleanup_container_resources()` helper method in commit 0b7a952 to consolidate the duplicated cleanup logic across exception handlers and delete_job().

kubeflow/trainer/backends/container/types.py
```diff
@@ -65,3 +65,11 @@ class ContainerBackendConfig(BaseModel):
         default_factory=TrainingRuntimeSource,
         description="Configuration for training runtime sources",
     )
+    initializer_image: str = Field(
+        default="kubeflow/training-operator:latest",
```
Fiona-Waters: I tested the initializer functionality with the local Docker/Podman container backend and found some issues relating to the use of this image here that need to be addressed in this PR. This image is a controller image and does not container initializer code. Instead we should use these: kubeflow/dataset-initializer:latest, kubeflow/model-initializer:latest. Functionality should be updated to select whichever one is required for each initialiser.

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,111 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+        ValueError: If the initializer type is not supported.
+    """
+    # Use the training-operator initializer script
+    # The initializer script is expected to be available in the image
+    if isinstance(initializer, (types.S3DatasetInitializer, types.S3ModelInitializer)):
+        python_cmd = "python -m kubeflow.storage_initializer.s3 "
+    elif isinstance(
+        initializer, (types.HuggingFaceDatasetInitializer, types.HuggingFaceModelInitializer)
+    ):
+        python_cmd = "python -m kubeflow.storage_initializer.hugging_face "
```
Fiona-Waters:
```suggestion
        python_cmd = "python -m pkg.initializers.dataset"
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -468,6 +505,153 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+        # Create and start the initializer container
+        container_id = self._adapter.create_and_start_container(
+            image=image,
+            command=command,
+            name=container_name,
+            network_id=network_id,
+            environment=env,
+            labels=labels,
+            volumes=volumes,
+            working_dir=constants.WORKSPACE_PATH,
```
Fiona-Waters: This should be /app - https://github.com/kubeflow/trainer/blob/master/cmd/initializers/dataset/Dockerfile#L3
```suggestion
            working_dir=/app,
```

kubeflow/trainer/backends/container/types.py
Fiona-Waters: Those images come from here FYI https://github.com/kubeflow/trainer/tree/master/cmd/initializers

HKanoje: Thanks for catching this! You're right - I've updated the implementation to use the correct images.

kubeflow/trainer/backends/container/utils.py
HKanoje: Done!

kubeflow/trainer/backends/container/backend.py
HKanoje: Fixed! I've changed the `working_dir` from /workspace to /app to match the Dockerfile convention. Added a comment referencing the Dockerfile for future maintainability.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -196,6 +196,35 @@ def _runtime_type(self) -> str:
...
+            container_ids: List of container IDs to stop and remove.
+            network_id: Network ID to delete.
+            stop_timeout: Timeout in seconds for stopping containers.
+        """
+        from contextlib import suppress
+
+        # Stop and remove containers
+        if container_ids:
+            for container_id in container_ids:
+                with suppress(Exception):
```
andreyvelich: Why do we use suppress here? We don't do that in other part of SDK.

kubeflow/trainer/backends/container/backend_test.py
```diff
@@ -861,3 +886,333 @@ def test_create_adapter_error_message_format():
         error_msg = str(exc_info.value)
         assert "Could not connect" in error_msg
         assert "tried:" in error_msg
+
+
+# Tests for Initializer Support
+@pytest.mark.parametrize(
```
andreyvelich: Can you add these test cases to the `test_train()` API: https://github.com/kubeflow/sdk/blob/6241fae531be0d7dc23a00481bcfe387f486ec58/kubeflow/trainer/backends/container/backend_test.py#L470

kubeflow/trainer/backends/container/backend_test.py
```diff
@@ -861,3 +886,333 @@ def test_create_adapter_error_message_format():
...
+            expected_status=SUCCESS,
+            config={
+                "step": constants.NODE + "-0",
+                "expected_log_count": 1,
+                "should_exclude_initializers": True,
+            },
+        ),
+    ],
+)
+def test_get_logs_with_initializers(container_backend, test_case):
```
andreyvelich: Same question, add it to test_get_job_logs()

kubeflow/trainer/backends/container/backend_test.py
```diff
@@ -861,3 +886,333 @@ def test_create_adapter_error_message_format():
...
+                "initializer": types.Initializer(
+                    model=types.S3ModelInitializer(storage_uri="s3://bucket/model")
+                ),
+                "initializer_timeout": True,
+            },
+            expected_error=TimeoutError,
+        ),
+    ],
+)
+def test_initializer_failures(container_backend, test_case):
```
andreyvelich: This can be added to `test_train()`

kubeflow/trainer/backends/container/types.py
```diff
@@ -65,3 +65,15 @@ class ContainerBackendConfig(BaseModel):
         default_factory=TrainingRuntimeSource,
         description="Configuration for training runtime sources",
     )
+    dataset_initializer_image: str = Field(
+        default="kubeflow/dataset-initializer:latest",
+        description="Container image for dataset initializers",
+    )
+    model_initializer_image: str = Field(
+        default="kubeflow/model-initializer:latest",
```
andreyvelich: Can you use GHCR image by default: https://github.com/kubeflow/trainer/pkgs/container/trainer%2Fdataset-initializer

kubeflow/trainer/backends/container/backend.py
```diff
@@ -468,6 +506,157 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+            logger.debug("Running model initializer")
+            self._run_single_initializer(
+                job_name=job_name,
+                initializer_config=initializer.model,
+                init_type="model",
+                image=model_image,
+                workdir=workdir,
+                network_id=network_id,
+            )
+            logger.debug("Model initializer completed")
```
andreyvelich, line 560: Is there a way to run these two containers in parallel?
How does multi attach volume work in Docker/Podman?

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,121 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+        if initializer.head_cpu:
+            env["HEAD_CPU"] = initializer.head_cpu
+        if initializer.head_mem:
+            env["HEAD_MEM"] = initializer.head_mem
+        if initializer.worker_cpu:
+            env["WORKER_CPU"] = initializer.worker_cpu
+        if initializer.worker_mem:
+            env["WORKER_MEM"] = initializer.worker_mem
+        if initializer.iam_role:
+            env["IAM_ROLE"] = initializer.iam_role
```
andreyvelich: Can we simplify this to something like we do here: https://github.com/kubeflow/sdk/blob/6241fae531be0d7dc23a00481bcfe387f486ec58/kubeflow/trainer/backends/kubernetes/utils.py#L574-L587

Env variables are always have the same name as fields, but just upper cased.

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,121 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
     """
     statuses = [get_container_status(adapter, c["id"]) for c in containers]
     return aggregate_status_from_containers(statuses)
+
+
+def build_initializer_command(initializer: types.BaseInitializer, init_type: str) -> list[str]:
```
andreyvelich: you don't need the init_type variable since you can understand whether it is dataset or model initializer by checking type of `initializer` field.
```suggestion
def build_initializer_command(initializer: types.BaseInitializer) -> list[str]:
```

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,121 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+    Raises:
+        ValueError: If the initializer type is not supported.
+    """
+    # Validate initializer type
+    supported_types = (
+        types.S3DatasetInitializer,
+        types.S3ModelInitializer,
+        types.HuggingFaceDatasetInitializer,
+        types.HuggingFaceModelInitializer,
+        types.DataCacheInitializer,
```
andreyvelich: Let's remove it from now, I am not sure how data cache can be supported in Container backend at the moment.
cc @akshaychitneni 
```suggestion
```

kubeflow/trainer/backends/container/utils.py
```diff
@@ -211,3 +211,121 @@ def aggregate_container_statuses(adapter, containers: list[dict]) -> str:
...
+    # have pkg.initializers.dataset and pkg.initializers.model respectively
+    if init_type == "dataset":
+        python_cmd = "python -m pkg.initializers.dataset"
+    else:  # model
+        python_cmd = "python -m pkg.initializers.model"
+
+    return ["bash", "-c", python_cmd]
+
+
+def build_initializer_env(initializer: types.BaseInitializer, init_type: str) -> dict[str, str]:
```
andreyvelich: Same point
```suggestion
def build_initializer_env(initializer: types.BaseInitializer) -> dict[str, str]:
```

kubeflow/trainer/backends/container/backend.py
```diff
@@ -468,6 +506,157 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+            workdir: Working directory path on host.
+            network_id: Network ID for containers.
+
+        Raises:
+            RuntimeError: If initializer fails to complete successfully.
+        """
+        # Run dataset initializer if configured
+        if initializer.dataset:
+            # Get and pull dataset initializer image
+            dataset_image = container_utils.get_initializer_image(self.cfg, "dataset")
```
andreyvelich: Can you simplify the logic here to be consistent with what we do in Kubernetes backend: https://github.com/kubeflow/sdk/blob/6241fae531be0d7dc23a00481bcfe387f486ec58/kubeflow/trainer/backends/kubernetes/backend.py#L704

Just simple define two utils function in container utils:
```py
container_utils.get_dataset_initializer()
container_utils.get_model_initializer()
```
Which returns internal type that you can use in the `_adapter.create_and_start_container()` API:
```py
@dataclass
class ContainerInitializer:
   image: str
   command: str
   env: dict
```

WDYT @HKanoje @Fiona-Waters ?

HKanoje:
Hi @andreyvelich @Fiona-Waters,

I've addressed all the review comments in the latest commit:

- Updated default images to GHCR (ghcr.io/kubeflow/trainer/dataset-initializer:latest and ghcr.io/kubeflow/trainer/model-initializer:latest)
- Replaced suppress with try-except blocks (added # noqa: SIM105 to satisfy linter)
- Refactored initializer utils with ContainerInitializer dataclass
- Added get_dataset_initializer() and get_model_initializer() functions following kubernetes backend pattern
- Simplified env building using get_optional_initializer_envs() helper
- Removed init_type parameter (now determined from initializer type)
- Removed DataCache support (not supported in container backend)
- Merged initializer tests into test_train() and test_get_job_logs()
- Removed duplicate test functions

kubeflow/trainer/backends/container/backend.py
```diff
@@ -468,6 +510,149 @@ def _get_job_containers(self, name: str) -> list[dict]:
...
+                workdir=workdir,
+                network_id=network_id,
+            )
+            logger.debug("Model initializer completed")
+
+    def _run_single_initializer(
+        self,
+        job_name: str,
+        container_init: container_utils.ContainerInitializer,
+        init_type: str,
```
andreyvelich: Instead of passing `init_type` separately, you can simple add `name` to the ContainerInitializer type which can be:
```
name = dataset-initializer
name = model-initializer
```
Then, just use this name in the `f"{self.label_prefix}/step"`

kubeflow/trainer/backends/container/backend.py
andreyvelich: Any thoughts @kubeflow/kubeflow-sdk-team @Fiona-Waters ?

Fiona-Waters: That's a great idea @andreyvelich  From a little bit of reading... both Docker and Podman support multiple containers mounting the same volume simultaneously. The filesystem (not the container runtime) handles concurrent access. As long as containers write to different paths, there are no conflicts. This would be a nice addition.

andreyvelich: @Fiona-Waters @HKanoje Shall we make this improvement in the followup PR?
@HKanoje Please can you create an issue to track it.

kubeflow/trainer/backends/container/backend.py
andreyvelich: cc @HKanoje

HKanoje: Done! @andreyvelich 

Changes made:

Added `name` field to `ContainerInitializer` dataclass
Set `name="dataset-initializer"` in `get_dataset_initializer()`
Set `name="model-initializer"` in `get_model_initializer()`
Removed `init_type` parameter from `_run_single_initializer()`
Now using `container_init.name` for labels and log messages

PR #216, opened 2025-12-23
feat(trainer): Support optional image for CustomTrainer

andreyvelich:
As we discussed on the last call, we would like to give users ability to optionally override an image in `CustomTrainer()`: https://youtu.be/v44YHEwBRGc

This is useful when users would like to bring their own images and use custom training function to start the TrainJob. 


kramaranya (commented):
Thanks, @andreyvelich
/lgtm
/approve

Issue #196, opened 2025-12-10
Add `get_job_events()` API


andreyvelich:
### What you would like to be added?

In the Training Operator v1 SDK users have ability to check job events via `get_job_logs(verbose=True)` API https://github.com/kubeflow/trainer/blob/release-1.9/sdk/python/kubeflow/training/api/training_client.py#L1279

This provides users with additional clarity about the state of the TrainJob when the logs alone are not.

I propose we introduce a new API: `get_job_events() -> []Event`.

The Event object might looks as follows:

```python
@dataclass
class Event:
    # or we can use involvedObject: pod/node-0-0, trainjob/node
    involvedObjectKind: str
    involvedObjectName: str
    message: str
    reason: str
    eventTime: datetime
```

@kubeflow/kubeflow-sdk-team Any thoughts ?

### Why is this needed?

Give users ability to check TrainJob events.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

osamaahmed17:
@andreyvelich Can you assign this issue to me

osamaahmed17:
I see. I will unassign for this issue, thanks for letting me know @andreyvelich

sksingh2005:
@andreyvelich sir is this issue still open ?? Can i start some work on this issue ??

osamaahmed17:
@andreyvelich , if you are busy, I can also look upon and finish this pull request.

andreyvelich:
Sounds good, thanks for your time @osamaahmed17!
/assign @osamaahmed17

PR #197, opened 2025-12-10
feat(trainer): add get_job_events API to retrieve TrainJob events

adityashirsatrao007:
Hey folks,

This PR adds the `get_job_events()` API that was requested in #196.

### What's changed

I've introduced an `Event` dataclass and a new method on the TrainerClient that lets users fetch Kubernetes events tied to their TrainJobs. This is handy when you need more context about what's happening beyond just the logs - things like scheduling issues, image pull errors, resource constraints, etc.

**New bits:**
- `Event` dataclass with fields for the involved object, message, reason, timestamps, count, and type
- `TrainerClient.get_job_events(name)` that returns a sorted list of events
- Backend implementations:
  - Kubernetes backend does the real work (queries events API, filters by job/pods)
  - Local and container backends just return empty lists since events are a K8s thing

### Example usage

```python
from kubeflow.trainer import TrainerClient

client = TrainerClient()
events = client.get_job_events(name="my-training-job")

for e in events:
    print(f"[{e.event_time}] {e.reason}: {e.message}")
```

### Testing

Added unit tests covering success path and error scenarios (timeout, runtime errors). All existing tests still pass.

Let me know if you'd like any changes!

Fixes #196

PR #220, opened 2026-01-06
feat(trainer): add get_job_events API to retrieve TrainJob events

sksingh2005:
This PR adds the get_job_events() API that was requested in https://github.com/kubeflow/sdk/issues/196.
**What this PR does**:

New bits:

- Event dataclass with fields for the involved object, message, reason, timestamps, count, and type
- TrainerClient.get_job_events(name) that returns a sorted list of events

- Backend implementations:
   - Kubernetes backend does the real work (queries events API, filters by job/pods)
   - Local and container backends just return empty lists since events are a K8s thing

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -411,18 +411,18 @@ class KubernetesBackend(RuntimeBackend):
         logger.debug(f"{constants.TRAINJOB_KIND} {self.namespace}/{name} has been deleted")
 
     def get_job_events(self, name: str) -> list[types.Event]:
+        # Get all pod names related to this TrainJob
+        trainjob = self.get_job(name)
+
+        # Create set of all TrainJob-related resource names
+        trainjob_resources = {name}
+        for step in trainjob.steps:
+            trainjob_resources.add(step.name)
+            if step.pod_name:
+                trainjob_resources.add(step.pod_name)
+
         events = []
         try:
-            # Get all pod names related to this TrainJob
-            trainjob = self.get_job(name)
-
-            # Create set of all TrainJob-related resource names
-            trainjob_resources = {name}
-            for step in trainjob.steps:
-                trainjob_resources.add(step.name)
-                if step.pod_name:
-                    trainjob_resources.add(step.pod_name)
-
             # Retrieve events from the namespace
             event_response = self.core_api.list_namespaced_event(
                 namespace=self.namespace,
@@ -461,13 +461,9 @@ class KubernetesBackend(RuntimeBackend):
             events.sort(key=lambda e: e.event_time)
 
             return events
-        except (multiprocessing.TimeoutError, TimeoutError) as e:
+        except multiprocessing.TimeoutError as e:
             raise TimeoutError(
-                f"Timeout getting events for {constants.TRAINJOB_KIND}: {self.namespace}/{name}"
-            ) from e
-        except Exception as e:
-            raise RuntimeError(
-                f"Failed getting events for {constants.TRAINJOB_KIND}: {self.namespace}/{name}"
+                f"Timeout getting {constants.TRAINJOB_KIND} events: {self.namespace}/{name}"
             ) from e
 
     def __get_runtime_from_cr(
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -414,7 +414,7 @@ def mock_list_namespaced_event(*args, **kwargs):
 
     # Errors occur at call time, not during .get()
     if namespace == TIMEOUT:
-        raise TimeoutError()
+        raise multiprocessing.TimeoutError()
     if namespace == RUNTIME:
         raise RuntimeError()
 
--- a/kubeflow/optimizer/backends/kubernetes/backend.py
+++ b/kubeflow/optimizer/backends/kubernetes/backend.py
@@ -329,16 +329,16 @@ class KubernetesBackend(RuntimeBackend):
         logger.debug(f"{constants.OPTIMIZATION_JOB_KIND} {self.namespace}/{name} has been deleted")
 
     def get_job_events(self, name: str) -> list[Event]:
+        # Get the OptimizationJob to ensure it exists
+        job = self.get_job(name)
+
+        # Create set of all OptimizationJob-related resource names
+        optimization_job_resources = {name}
+        for trial in job.trials:
+            optimization_job_resources.add(trial.name)
+
         events = []
         try:
-            # Get the OptimizationJob to ensure it exists
-            job = self.get_job(name)
-
-            # Create set of all OptimizationJob-related resource names
-            optimization_job_resources = {name}
-            for trial in job.trials:
-                optimization_job_resources.add(trial.name)
-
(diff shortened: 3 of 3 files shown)


kubeflow/trainer/types/types.py
```diff
@@ -275,6 +275,26 @@ class TrainJob:
...
+
+    Attributes:
+        involved_object_kind: The kind of object this event is about (e.g., 'TrainJob', 'Pod').
+        involved_object_name: The name of the object this event is about.
+        message: Human-readable description of the event.
+        reason: Short, machine understandable string describing why this event was generated.
+        event_time: The time at which the event was first recorded.
+    """
+
+    involved_object_kind: str
```
andreyvelich, line 293: Check this: https://github.com/kubeflow/sdk/pull/197#discussion_r2621380954

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -410,6 +410,77 @@ def delete_job(self, name: str):
...
+                return events
+
+            # Filter events related to this TrainJob or its pods
+            for event in event_list.items:
+                if not (event.metadata and event.involved_object and event.first_timestamp):
+                    continue
+
+                # Check if event is related to the TrainJob itself or one of its pods
+                involved_object = event.involved_object
+                is_related = False
```
andreyvelich: You can remove this flag, and just append event if condition is true.
```suggestion
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -410,6 +410,79 @@ def delete_job(self, name: str):
...
+                # Event for the TrainJob object itself or JobSet (same name)
+                if (
+                    involved_object.name == name
+                    and involved_object.kind
+                    in {
+                        constants.TRAINJOB_KIND,
+                        "JobSet",
+                    }
+                    or involved_object.kind == "Pod"
+                    and involved_object.name in trainjob_pod_names
```
andreyvelich: Can you check this: https://github.com/kubeflow/sdk/pull/197#discussion_r2621441772
Just create set of the required resource names, that will simplify your if check.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -410,6 +410,79 @@ def delete_job(self, name: str):
...
+            events.sort(key=lambda e: e.event_time)
+
+            return events
+
+        except multiprocessing.TimeoutError as e:
+            msg = (
+                f"Timeout to get events for {constants.TRAINJOB_KIND}: "
+                f"{self.namespace}/{name}"
+            )
+            raise TimeoutError(msg) from e
```
andreyvelich: Please be consistent with other APIs, and return the error directly:
```suggestion
            raise TimeoutError(
              f"Timeout to get events for {constants.TRAINJOB_KIND}: {self.namespace}/{name}"
              ) from e
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -410,6 +410,79 @@ def delete_job(self, name: str):
...
+        except Exception as e:
+            msg = (
+                f"Failed to get events for {constants.TRAINJOB_KIND}: "
+                f"{self.namespace}/{name}"
+            )
+            raise RuntimeError(msg) from e
```
andreyvelich: Same comment.

kubeflow/trainer/backends/kubernetes/backend.py
sksingh2005: @andreyvelich sir is this same applicable for the one in optimizer folder also ??

andreyvelich: Sure, let's be consistent between those two.

kubeflow/trainer/types/types.py
sksingh2005: @andreyvelich  sir should i use # noqa: N815 comments to suppress ruff's naming convention warnings for those specific fields as i was checking with that and it was causing problem ..

andreyvelich: Oh, sorry, you are right, let's use the snake case like for other objects.
Could you also use Args for docstring with the format similar to this: https://github.com/kubeflow/sdk/blob/2002e057c4ef6dd57cbffe4550db4823fb764762/kubeflow/trainer/types/types.py#L32-L42

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1233,13 +1233,43 @@ def test_delete_job(kubernetes_backend, test_case):
         kubernetes_backend.delete_job(test_case.config.get("name"))
         assert test_case.expected_status == SUCCESS
 
-        kubernetes_backend.custom_api.delete_namespaced_custom_object.assert_called_with(
```
andreyvelich, line 1240: Why did you remove it ?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1233,13 +1233,43 @@ def test_delete_job(kubernetes_backend, test_case):
...
+
+
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="get job events with valid trainjob",
+            expected_status=SUCCESS,
+            config={"name": BASIC_TRAIN_JOB_NAME},
+            expected_output=list,  # Expect a list of events
```
andreyvelich: Please can you try to mock event to list events? That will allow us to correctly verify output of this API and the Event objects.
You can check example here: https://github.com/kubeflow/sdk/blob/2002e057c4ef6dd57cbffe4550db4823fb764762/kubeflow/trainer/backends/kubernetes/backend_test.py#L99

https://github.com/kubeflow/sdk/blob/2002e057c4ef6dd57cbffe4550db4823fb764762/kubeflow/trainer/backends/kubernetes/backend_test.py#L1086

kubeflow/trainer/types/types.py
sksingh2005: Done sir

kubeflow/trainer/backends/kubernetes/backend_test.py
sksingh2005: Sir this was the change from previous pr which i took over, should i revert this ?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1233,13 +1233,43 @@ def test_delete_job(kubernetes_backend, test_case):
...
-        kubernetes_backend.custom_api.delete_namespaced_custom_object.assert_called_with(
-            constants.GROUP,
-            constants.VERSION,
-            test_case.config.get("namespace", DEFAULT_NAMESPACE),
-            constants.TRAINJOB_PLURAL,
-            name=test_case.config.get("name"),
-        )
+    except Exception as e:
+        assert type(e) is test_case.expected_error
+    print("test execution complete")
```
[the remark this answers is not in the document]
andreyvelich: It's fine here.

kubeflow/trainer/backends/kubernetes/backend_test.py
sksingh2005: Is this correct sir should i proceed testing this for verification ?

PR #223, opened 2026-01-10
chore: Enforce strict backend version compatibility checks

### LINKED ISSUE #221: Ensure Control Plane Availability During Kubeflow SDK Client Creation

As we discussed in[ this Slack thread](https://cloud-native.slack.com/archives/C073N7BMLB1/p1764075248466009?thread_ts=1763585362.957169&cid=C073N7BMLB1), we want to expose public ConfigMap in the system namespace that tells which version of Kubeflow control plane components are installed (Trainer, Pipelines, Spark Operator, etc.)

This ConfigMaps should be accessible by any user with `system:authenticated` RBAC, and have these names:

```
kubeflow-trainer-public
kubeflow-pipelines-public
kubeflow-spark-public
kubeflow-hub-public
```

Before user tries to use corresponding clients in Kubeflow SDK, we should verify that the appropriate version of control plane is deployed (e.g. `kubeflow_trainer_api == Trainer version`).

We can start with Kubeflow Trainer, and introduce similar functionality for other clients.

- [x] Kubeflow Trainer
   - https://github.com/kubeflow/trainer/pull/3083
- [ ] Kubeflow Pipelines
   - https://github.com/kubeflow/pipelines/pull/12717
- [ ] Kubeflow Spark Operator
- [ ] Kubeflow Hub

cc @kubeflow/wg-pipeline-leads @kubeflow/kubeflow-trainer-team @kubeflow/kubeflow-sdk-team @juliusvonkohout @kubeflow/wg-data-leads @Shekharrajak 

### Why is this needed?

Verifies that Kubeflow SDK clients can be used.

### Love this feature?

sameerdattav:
Fixes #221.

### Summary
Adds a strict control-plane version compatibility check for the Trainer
**Kubernetes** backend. The check is fail-fast, user-friendly, opt-out, and
covered by unit tests.

### Approach
- Runs from `KubernetesBackend.__init__` after `CoreV1Api` creation
- Uses namespace from `KUBEFLOW_SYSTEM_NAMESPACE` (default: `kubeflow`)
- Reads `kubeflow-trainer-public` ConfigMap (`kubeflow_trainer_api_version`)
- Compares against local version from `importlib.metadata`
- Raises `RuntimeError` on mismatch or missing control plane
- Skippable via `KUBEFLOW_TRAINER_SKIP_VERSION_CHECK`

### Tests
- Added coverage for success, mismatch, missing ConfigMap/data, and skip path
- Updated existing tests to account for verification during backend init
- `pytest` passes locally

Thanks @RavinduWeerakoon for the initial approach and discussion.


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -56,6 +59,86 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+
+    def verify_backend(self) -> None:
+        """Verify that the Trainer control plane matches the local SDK version.
+
+        This check only applies to the Kubernetes backend. It compares the
+        version of the installed ``kubeflow-trainer-api`` package with the
+        version published by the control plane in a public ConfigMap.
+
+        The verification can be skipped entirely by setting the
+        ``KUBEFLOW_TRAINER_SKIP_VERSION_CHECK`` environment variable to any
```
andreyvelich: What is the reason to introduce this env variable?
I would suggest that we always check the version for now.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -56,6 +59,86 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+        if local_version != server_version:
+            raise RuntimeError(
+                "Kubeflow Trainer SDK version mismatch detected: local "
+                f"kubeflow-trainer-api=={local_version}, control plane "
+                f"kubeflow-trainer-api=={server_version}. The control plane "
+                "may be outdated or mismatched with this SDK. Please upgrade "
+                "either the SDK or the control plane so the versions match, "
+                "or set KUBEFLOW_TRAINER_SKIP_VERSION_CHECK=1 to bypass "
+                "this verification (not recommended)."
+            )
```
andreyvelich: For now, let's do `logger.warning()` if version of control plane is incompatible or config map is missing.
We need to release the Trainer v2.2 where we add this ConfigMap.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -56,6 +59,86 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+                "Kubeflow Trainer control-plane version ConfigMap '"
+                f"{config_map_name}' in namespace '{system_namespace}' is "
+                "missing the 'kubeflow_trainer_api_version' data key. "
+                "Ensure the control plane is configured correctly, or set "
+                "KUBEFLOW_TRAINER_SKIP_VERSION_CHECK=1 to skip this check."
+            )
+
+        try:
+            local_version = metadata.version("kubeflow-trainer-api")
+        except Exception as e:
```
andreyvelich: You don't need this check, since this package is always installed.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -56,6 +59,86 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+            raise RuntimeError(
+                "Failed to determine local 'kubeflow-trainer-api' package "
+                "version. Ensure the package is installed correctly, or set "
+                "KUBEFLOW_TRAINER_SKIP_VERSION_CHECK=1 to skip this check."
+            ) from e
+
+        if local_version != server_version:
```
andreyvelich: I guess, we should do this check:
1. For the minor version it should: `server_version == local_version`
2. For the patch version it should: `server_version >= local_version`

Since we don't introduce new features for the patch versions.

We might also want to introduce strict dependency for minor version here 
(e.g. `kubeflow-trainer-api~=2.1.0`): https://github.com/kubeflow/sdk/blob/7797f35dfe8dce1993e968364c17c59a6a422697/pyproject.toml#L31C4-L31C24

What do you think @kubeflow/kubeflow-sdk-team @kubeflow/kubeflow-trainer-team ?

Let's talk about it more at the Kubeflow SDK call: https://docs.google.com/document/d/1jH2WAX2ePxOfI4JuiVK9nPlesDMiyg67xzLwhpR7wTQ/edit?tab=t.0

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -56,6 +59,86 @@ def __init__(self, cfg: KubernetesBackendConfig):
...
+                "control plane is installed and exposes this ConfigMap, or "
+                "set KUBEFLOW_TRAINER_SKIP_VERSION_CHECK=1 to skip this "
+                "check."
+            ) from e
+
+        data = getattr(config_map, "data", None)
+        server_version = None
+
+        if data and isinstance(data, dict):
+            server_version = data.get("kubeflow_trainer_api_version")
```
andreyvelich: If the server_version == dev, you should skip the check.

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: @sameerdattav We briefly talked about it at the latest call: https://youtu.be/zwPVfB8tTuY?t=2254 
Initially, let's just verify that ConfigMap with the appropriate info (e.g. `kubeflow_trainer_version`) exists.
In the future, we can discuss how we can verify the appropriate version in TrainerClient.

PR #193, opened 2025-12-09
fix: Upgrade urllib3 to  v2.6.1

Fiona-Waters:
**What this PR does / why we need it**:
This PR upgrades urllib3 to v2.6.1 to remove these CVEs:
https://www.cve.org/CVERecord?id=CVE-2025-66418
https://www.cve.org/CVERecord?id=CVE-2025-66471

**Which issue(s) this PR fixes** _(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)_:

Fixes #

**Checklist:**

- [ ] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing

PR #226, opened 2026-01-13
feat: added git cliff for generating changelogs

### LINKED ISSUE #99: Adopt external changelog generation tool

jaiakash:
For change log generation we are still using custom python script that is not standard. So I had brief discussion (over tools and even format for generated changelogs) on #99 with @kramaranya and finalized with `git-cliff` for tool and a common GitHub format to use for generating change logs.

This PR introduces the standardized changelog generation setup.
- Section for features, bug fixes, misc etc.
- Section for new contributors.
- Add git-cliff as dependancies and updated make command to use git cliff not old script.

Fixes #99

**Checklist:**

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -63,7 +63,7 @@ trim = true
 footer = ""
 
 [git]
-conventional_commits = false
+conventional_commits = true
 filter_unconventional = true
 split_commits = false
 


Makefile
```diff
@@ -65,11 +65,13 @@ uv-venv:  ## Create uv virtual environment
 
 .PHONY: release
 release: install-dev
-	@if [ -z "$(VERSION)" ]; then echo "Usage: make release VERSION=0.1.0"; exit 1; fi
-	@V_NO_V=$(VERSION); \
+	@if [ -z "$(RELEASE_VERSION)" ]; then echo "Usage: make release PREV_VERSION=0.1 RELEASE_VERSION=0.2"; exit 1; fi
+	@if [ -z "$(PREV_VERSION)" ]; then echo "Usage: make release PREV_VERSION=0.1 RELEASE_VERSION=0.2"; exit 1; fi
```
kramaranya: Do we need to introduce prev version? If so, please update RELEASE.md

Makefile
```diff
@@ -65,11 +65,13 @@ uv-venv:  ## Create uv virtual environment
...
+	@V_NO_V=$(RELEASE_VERSION); \
 	sed -i.bak "s/^__version__ = \".*\"/__version__ = \"$$V_NO_V\"/" kubeflow/__init__.py && \
 	rm -f kubeflow/__init__.py.bak
-	@uv run python scripts/gen-changelog.py --token=$${GITHUB_TOKEN} --version=$(VERSION)
+	@MAJOR_MINOR=$$(echo "$(RELEASE_VERSION)" | cut -d. -f1,2); \
+	uv run git-cliff v$(PREV_VERSION)..HEAD -o CHANGELOG/CHANGELOG-$$MAJOR_MINOR.md
```
kramaranya: How does it use our config in `cliff.toml`? Does it detect it by default? Or shall we pass it in the call? Also it uses `-o` for overwrite, right? We need to add changelog on top of existing one for patch releases

RELEASE.md
```diff
@@ -57,7 +57,15 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
 
 #### 1. Update Version and Changelog
 
-1. Generate version and changelog locally (this will sync dependencies automatically):
+1. Generate changelog locally:
+
+   ```sh
+   # Generate changes from the last tag to HEAD
+   # Replace <previous-tag> with the last released version (e.g., v0.2.1)
+   git-cliff --config cliff.toml <previous-tag>..HEAD --prepend CHANGELOG/CHANGELOG-X.Y.md
```
kramaranya: Why do we need this if `make release` does it for us?

cliff.toml
```diff
@@ -0,0 +1,42 @@
+[changelog]
```
kramaranya, line 5: Shall we add integration with GitHub https://git-cliff.org/docs/integration/github? This will let us add:
- GitHub usernames
- Contributors list (all contributors / first time)
- Pull request links (associated with the commits)

kramaranya (commented):
Thank you @jaiakash 
I left a few questions. Please also verify that https://github.com/kubeflow/sdk/blob/main/.github/workflows/release.yml works with the changes

Makefile
jaiakash: 1. By default its takes from the directory where the command is executed.  We have `cliff.toml` at root of repo.
2. Nope, `-o` is for saving output, we are saving to `CHANGELOG/CHANGELOG-$$MAJOR_MINOR.md` files. https://git-cliff.org/docs/usage/args#options
3. For patch releases, we can use `append` parameter to add on top of existing changelog.

RELEASE.md
jaiakash: I agree, we dont need this now. I will edit the `make release` command itself with append parameter which can be used here.

cliff.toml
jaiakash: We are already doing it, right. Thats how its getting the github username for PRs, new contributor list, PR links, etc. 
By default `git-cliff` automatically takes data from git config. The other options are used when the repo link is different.

> As default, remote upstream URL is automatically retrieved from the Git repository.

Source: https://git-cliff.org/docs/integration/github/#setting-up-the-remote

Only thing that i see we should change is to pass the GitHub token when we run in `release.yaml`.
I will edit and verify both the file and `release.yaml` to confirm that. Let me know your thoughts.

Makefile
andreyvelich: @jaiakash Can you dynamically detect previous version based on `VERSION`?
- For the first patch release, take the previous minor version (e.g. v0.4.0 .. v0.5.0)
- For the following patch releases, take the previous patch version (e.g. v0.4.2 .. v0.4.3)

cliff.toml
```diff
@@ -0,0 +1,42 @@
...
+
+footer = ""
+
+
+[git]
+conventional_commits = true
+filter_unconventional = false
+split_commits = false
+
+tag_pattern = "^v[0-9]+\\.[0-9]+\\.[0-9]+$"
```
andreyvelich: Tags don't have `v` prefix in the SDK.

RELEASE.md
andreyvelich: Yes, let's remove it from the release guidelines since we don't want user to make mistakes with previous versions.

Makefile
```diff
@@ -65,11 +65,13 @@ uv-venv:  ## Create uv virtual environment
...
 .PHONY: release
 release: install-dev
-	@if [ -z "$(VERSION)" ]; then echo "Usage: make release VERSION=0.1.0"; exit 1; fi
-	@V_NO_V=$(VERSION); \
+	@if [ -z "$(RELEASE_VERSION)" ]; then echo "Usage: make release PREV_VERSION=0.1 RELEASE_VERSION=0.2"; exit 1; fi
+	@if [ -z "$(PREV_VERSION)" ]; then echo "Usage: make release PREV_VERSION=0.1 RELEASE_VERSION=0.2"; exit 1; fi
+	@V_NO_V=$(RELEASE_VERSION); \
 	sed -i.bak "s/^__version__ = \".*\"/__version__ = \"$$V_NO_V\"/" kubeflow/__init__.py && \
 	rm -f kubeflow/__init__.py.bak
-	@uv run python scripts/gen-changelog.py --token=$${GITHUB_TOKEN} --version=$(VERSION)
```
andreyvelich, line 72: You can remove this script, right ?
https://github.com/kubeflow/sdk/blob/a03e7e6f65b3dcfc136716d2ca3d0ef9a346628c/scripts/gen-changelog.py

cliff.toml
```diff
@@ -0,0 +1,42 @@
...
+ignore_tags = ".*-(alpha|beta|rc).*"
```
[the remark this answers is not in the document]
andreyvelich: Not really needed.

RELEASE.md
```diff
@@ -57,7 +57,15 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
...
+   ```
+
```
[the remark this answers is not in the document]
andreyvelich: Good point, @jaiakash did you try to see how the update of already existing files might work?
E.g. https://github.com/kubeflow/sdk/blob/main/CHANGELOG/CHANGELOG-0.2.md#021-2025-11-25

cliff.toml
jaiakash: My local fork had v as prefix, fixed that.

Makefile
jaiakash: Sure, I edited it to get it dynamically, to get the prev version from the command itself. Check the following output.
1. For first patch release, it takes previous minor version.
```
$ make release VERSION=0.3

Environment is ready.
Auto-detected previous version: 0.2.1
Generating changelog for range: 0.2.1..0.3.0
Changelog generated at CHANGELOG/CHANGELOG-0.3.md
```

2. For following patch release, it stills take prev version and append both to the file. ***

```
$ make release VERSION=0.2

Environment is ready.
Auto-detected previous version: 0.1.0
Generating changelog for range: 0.1.0..0.2.1
Changelog generated at CHANGELOG/CHANGELOG-0.2.md
```

For CHANGELOG-0.3.md it will have all changs in 0.3.0 tag, (excluding any changes in last tag 0.2.1)
For CHANGELOG-0.2.md it will have all change in 0.2.0 AND 0.2.1 tag.

One way we can solve it to use `--apend` parameter, but what i saw in  https://github.com/kubeflow/sdk/blob/main/CHANGELOG/CHANGELOG-0.2.md that it has changes for both `0.2.0` and `0.2.1`

Can you both please confirm this?

RELEASE.md
jaiakash: I updated it. Users can run `make release` along with GITHUB TOKEN to get the change log locally.

Makefile
jaiakash: Done

jaiakash:
Hi @andreyvelich @kramaranya, fixed things after your review comments. Here is the current flow - 

1. We can generate the change log for first patch release. (It will give all change log for `0.3.x`)
```
$ make release VERSION=0.3                                                                                                                                                                  

uv virtual environment already exists in /home/ubuntu/projects/kubeflow/sdk/.venv.
Using virtual environment at: /home/ubuntu/projects/kubeflow/sdk/.venv
Syncing dependencies with uv...
Resolved 55 packages in 0.65ms
Audited 49 packages in 0.44ms
Environment is ready.
Previous version : 0.2.1
Range end        : 0.3.2
Generating changelog for 0.2.1..0.3.2
Changelog generated at CHANGELOG/CHANGELOG-0.3.md
```

2. For subsequent patches, simply giving the update patch number would be fine, the change log would be appended to the `changelog.md` file
```
$ make release VERSION=0.3.1 
                                                                                                                                                            
uv virtual environment already exists in /home/ubuntu/projects/kubeflow/sdk/.venv.
Using virtual environment at: /home/ubuntu/projects/kubeflow/sdk/.venv
Syncing dependencies with uv...
Resolved 55 packages in 0.64ms
Audited 49 packages in 0.44ms
Environment is ready.
Previous version : 0.2.1
Range end        : 0.3.1
Generating changelog for 0.2.1..0.3.1
Changelog generated at CHANGELOG/CHANGELOG-0.3.md
```


Working Samples - 
Sample Release Notes: [jaiakash/sdk@0.3.0 (release)](https://github.com/jaiakash/sdk/releases/tag/0.3.0)
Sample Run: [jaiakash/sdk/actions/runs/21023011855/job/60441399211](https://github.com/jaiakash/sdk/actions/runs/21023011855/job/60441399211)

Makefile
```diff
@@ -65,11 +65,14 @@ uv-venv:  ## Create uv virtual environment
 
 .PHONY: release
 release: install-dev
-	@if [ -z "$(VERSION)" ]; then echo "Usage: make release VERSION=0.1.0"; exit 1; fi
-	@V_NO_V=$(VERSION); \
-	sed -i.bak "s/^__version__ = \".*\"/__version__ = \"$$V_NO_V\"/" kubeflow/__init__.py && \
+	@if [ -z "$(VERSION)" ]; then \
+		echo "Usage: make release VERSION=0.1.0 or VERSION=0.3"; \
```
andreyvelich: Please can we ensure that users always specify the full version (e.g. 0.1.0, 0.3.0, or 0.3.2)
That will remove ambiguity about how the make release command should run.

cliff.toml
```diff
@@ -0,0 +1,63 @@
...
+{%- if commits | length > 0 %}
+
+### {{ group | replace(from="feat", to="🚀 Features")
+          | replace(from="fix", to="🐛 Bug Fixes")
+          | replace(from="docs", to="📚 Documentation")
+          | replace(from="chore", to="🧹 Chores")
+          | replace(from="refactor", to="♻️ Refactoring")
+          | replace(from="test", to="🧪 Tests")
+          | replace(from="ci", to="⚙️ CI")
+          | replace(from="perf", to="⚡ Performance")
```
andreyvelich: We don't such scopes, let's only keep: chore fix feat revert
https://github.com/kubeflow/sdk/blob/main/.github/workflows/check-pr-title.yaml

cliff.toml
```diff
@@ -0,0 +1,63 @@
...
+{%- endif %}
+
+{%- set groups = commits | group_by(attribute="group") -%}
+
+{%- for group, commits in groups %}
+{%- if commits | length > 0 %}
+
+### {{ group | replace(from="feat", to="🚀 Features")
+          | replace(from="fix", to="🐛 Bug Fixes")
+          | replace(from="docs", to="📚 Documentation")
```
andreyvelich: I think we can remove docs as well, since the PR type can be feat or fix, but the scope can be docs but not the type, e.g.
feat(docs): Create new page
fix(docs): Fix the invalid example

cliff.toml
```diff
@@ -0,0 +1,63 @@
...
+{%- else %}
+## [Unreleased]
+{%- endif %}
+
+{%- set groups = commits | group_by(attribute="group") -%}
+
+{%- for group, commits in groups %}
+{%- if commits | length > 0 %}
+
+### {{ group | replace(from="feat", to="🚀 Features")
```
andreyvelich: With the replace, are we going to keep scopes in the list: https://github.com/kubeflow/sdk/blob/main/.github/workflows/check-pr-title.yaml#L32 ?

Can we group PRs by type, but keep the entire title in the list, like here: https://github.com/kubeflow/sdk/blob/main/CHANGELOG/CHANGELOG-0.2.md#new-features-1

scripts/gen-changelog.py
```diff
@@ -1,139 +1,77 @@
...
 
-    if not commits:
-        print("No commits found in range")
-        return
 
-    categories = {"feat": [], "fix": [], "chore": [], "revert": [], "misc": []}
+def get_prev_version(version: str) -> str:
+    major, minor = version.split(".")[:2]
+    tags = run(
+        "git tag --sort=version:refname | grep -E '^[0-9]+\\.[0-9]+\\.[0-9]+$' || true"
```
andreyvelich: Why do you need `|| true` ?

scripts/gen-changelog.py
```diff
@@ -1,139 +1,77 @@
...
+            run(f"git rev-parse {version}")
+        except subprocess.CalledProcessError:
+            print(f"Error: tag {version} does not exist", file=sys.stderr)
+            sys.exit(1)
+        return version
 
-    os.makedirs(CHANGELOG_DIR, exist_ok=True)
+    # VERSION=0.3 → latest 0.3.x
+    tags = run(
+        f"git tag --sort=version:refname | grep -E '^{re.escape(version)}\\.[0-9]+$' || true"
```
andreyvelich: same question.

Makefile
jaiakash: Sure that makes more sense.
Also do you think we should generate the change log file with full version name? Like `CHANELOG-0.2.1.md` or `CHANELOG-0.3.1.md`
Asking because in the `release.yml` I am handeling to take only passed version changelog, if we have full version file, then no conditional parsing would be needed?
https://github.com/jaiakash/sdk/blob/0049dbcfd3e48318d7c1da389fcb368c4e77fb9b/.github/workflows/release.yml#L187

cliff.toml
jaiakash: Sure keeping these 4 only.

scripts/gen-changelog.py
jaiakash: removing it, i was using it during testing.

Makefile
andreyvelich: No, the file name should contain only major + minor version since we append patch releases to the existing files.
This is same as in Kueue: https://github.com/kubernetes-sigs/kueue/tree/main/CHANGELOG

jaiakash: Sure

cliff.toml
jaiakash: Yeah, updated it to have full PR name.

cliff.toml
```diff
@@ -0,0 +1,59 @@
+[changelog]
+header = """
+# Changelog
+"""
+
+body = """
+
+{%- if version %}
+## [{{ version | trim_start_matches(pat="v") }}] ({{ timestamp | date(format="%Y-%m-%d") }})
```
andreyvelich: Do we need trim_start_matches here since version doesn't start with `v` in the SDK?

cliff.toml
```diff
@@ -0,0 +1,59 @@
...
+          | replace(from="revert", to="⏪ Reverts")
+          | replace(from="chore", to="🔧 Chores")
+          | upper_first }}
+
+{%- for commit in commits | reverse %}
+- {{ commit.message | split(pat="\n") | first | trim }}
+   {%- if commit.remote and commit.remote.username %} (@{{ commit.remote.username }})
+   {%- else %} ({{ commit.author.name }})
+   {%- endif %}
+{%- endfor %}
```
andreyvelich: Can we also include reference to the PRs (e.g. #127), something like:
```
feat(trainer): Add CustomTrainerContainer to create TrainJobs from image (#127 by @andreyvelich)
```

jaiakash: Its there right, The `0.3.2` sample release didnt had because all of them are my direct commits to fork.
Check this: https://github.com/jaiakash/sdk/releases/tag/0.3.0

```feat(trainer): Allow to reference runtime by name (#214) (@andreyvelich)```

File: cliff.toml (hunk 15 above)
Comment (jaiakash) [reply cliff.toml]: Removed

File: cliff.toml (hunk 16 above)
Comment (andreyvelich) [reply cliff.toml]: Great!

Comment (andreyvelich) [reply cliff.toml]: Just a small question, is there a way to keep only 2 parenthesis so it is easier to read? e.g. 
`feat(trainer): Allow to reference runtime by name (#214 by @andreyvelich)`

Comment (jaiakash) [reply cliff.toml]: Yeah @andreyvelich , but we have to parse the commits manually. Below is the template we’re using to extract and format commit messages:

```
{% for commit in commits | reverse %}
{%- set message = commit.message | split(pat="\n") | first | trim -%}
{% if commit.remote and commit.remote.username -%}
   {%- set parts = message | split(pat=" (#") -%}
   {% if parts | length > 1 and parts | last | trim | split(pat=")") | length > 1 -%}
      {%- set pr_part = parts | last | trim -%}
      {%- set pr_number = pr_part | replace(from=")", to="") -%}
      - {{ parts | slice(end=-1) | join(sep=" (#") }} (#{{ pr_number }} by @{{ commit.remote.username }})
   {% else -%}
      - {{ message }} (@{{ commit.remote.username }})
   {% endif -%}
{% else -%}
- {{ message }} ({{ commit.author.name }})
{% endif -%}
{% endfor %}
```

This will generate in given format with PR link followed by name.
`- feat(trainer): support NVIDIA MIG device resources in TrainJob device… (#204 by @LabsJS)`
For direct commit.
`- fix: added dev tools (@jaiakash)`

Let me know if its fine?

Comment (andreyvelich) [reply cliff.toml]: I think, that should be fine. Otherwise, the long titles might be hidden with ... , right ?
i. guess, that simpler version also works, right ?
```
{% for commit in commits | reverse %}
{%- set message = commit.message | split(pat="\n") | first | trim -%}

- {{ message }}
  {%- if commit.remote and commit.remote.pr_number and commit.remote.username -%}
    (#{{ commit.remote.pr_number }} by @{{ commit.remote.username }})
  {%- elif commit.remote and commit.remote.username -%}
    (@{{ commit.remote.username }})
  {%- else -%}
    ({{ commit.author.name }})
  {%- endif -%}

{% endfor %}
```

Comment (jaiakash) [reply cliff.toml]: No actually the example commit itself had ... at end that why it was showing those 3 dots. 👀😂
Even for long titles we are keeping the title as it is, followed by PR number and name.

As for simpler version, all commits are coming in same line.  So lets use the same git-cliff format which i shared above.

Have update the PR with config to give the output in new format.

Eg: 
```
- fix: Upgrade urllib3 to  v2.6.1 (#193 by @Fiona-Waters)
- fix(trainer): expose CustomTrainerContainer for import (#185 by @AndEsterson)
- fix: update permissions for welcome workflow to avoid 403 error (#181 by @aniketpati1121)
- fix(ci): Move permissions to the workflow root (#177 by @kramaranya)
```

Comment (andreyvelich) [reply cliff.toml]: Sure, that sounds good, thanks!

File: cliff.toml [hunk 17]
```diff
@@ -0,0 +1,68 @@
...
+
+### 🎉 New Contributors
+{%- for contributor in first_timers %}
+- @{{ contributor.username }} — welcome and thanks for your first contribution!
+{%- endfor %}
+{%- endif %}
+
+"""
+
+footer = "adsasd"
```
kramaranya: This has to be removed

cliff.toml
```diff
@@ -0,0 +1,68 @@
+[changelog]
+header = """
+# Changelog
+"""
+
+body = """
+
+{%- if version %}
+## [{{ version }}] ({{ timestamp | date(format="%Y-%m-%d") }})
```
kramaranya: Can you update this to keep version clickable, same as https://github.com/kubeflow/sdk/blob/main/CHANGELOG/CHANGELOG-0.2.md?plain=1#L1?

cliff.toml
```diff
@@ -0,0 +1,68 @@
...
+          | upper_first }}
+
+{% for commit in commits | reverse %}
+{%- set message = commit.message | split(pat="\n") | first | trim -%}
+{% if commit.remote and commit.remote.username -%}
+   {%- set parts = message | split(pat=" (#") -%}
+   {% if parts | length > 1 and parts | last | trim | split(pat=")") | length > 1 -%}
+      {%- set pr_part = parts | last | trim -%}
+      {%- set pr_number = pr_part | replace(from=")", to="") -%}
+      - {{ parts | slice(end=-1) | join(sep=" (#") }} (#{{ pr_number }} by @{{ commit.remote.username }})
```
kramaranya: Can you also keep PRs clickable? https://github.com/kubeflow/sdk/blob/main/CHANGELOG/CHANGELOG-0.2.md?plain=1#L5

.github/workflows/release.yml
```diff
@@ -183,16 +183,16 @@ jobs:
           CHANGELOG_FILE="CHANGELOG/CHANGELOG-${MAJOR_MINOR}.md"
           set -euo pipefail
           [[ -f "$CHANGELOG_FILE" ]] || { echo "ERROR: $CHANGELOG_FILE not found" >&2; exit 1; }
-          HEADER_REGEX="^# \\[${VERSION//./\\.}\\]"
-          SECTION=$(sed -n "/$HEADER_REGEX/,\$p" "$CHANGELOG_FILE" | tail -n +2)
```
kramaranya, line 187: I wonder whether this change breaks the current release CI, since now the captured section includes the version header and I think this might fail and not find the next version.
@jaiakash could you please verify the full release workflow with your changes and git cliff on your fork, before we cut a new release? fyi you can create a test pypi project similar to https://pypi.org/project/kubeflow-test/#history and here's an example of the workflow https://github.com/kramaranya/sdk/actions/runs/17782191001

RELEASE.md
```diff
@@ -55,13 +55,14 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
 
 ### Step-by-Step Release Process
 
-#### 1. Update Version and Changelog
+#### 1. Generate Changelog Locally
 
-1. Generate version and changelog locally (this will sync dependencies automatically):
+1. Generate changelog locally:
```
kramaranya: `make release` still updates the version, right? In this case revert these chnages please

RELEASE.md
```diff
@@ -55,13 +55,14 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
...
 
    ```sh
    export GITHUB_TOKEN=<your_github_token>
-   make release VERSION=X.Y.Z
+   make release VERSION=<full version>
```
kramaranya: In this doc we call version as X.Y.X, let's keep it as it is unless it's not relevant anymore?
```suggestion
   make release VERSION=X.Y.Z
```

scripts/gen-changelog.py
```diff
@@ -1,139 +1,72 @@
...
-    parser = argparse.ArgumentParser(description="Generate changelog for Kubeflow SDK")
-    parser.add_argument("--token", required=True, help="GitHub Access Token")
-    parser.add_argument("--version", required=True, help="Target version (e.g. 0.1.0)")
-
-    args = parser.parse_args()
+    prev = None
+    for tag in tags:
+        t_major, t_minor, _ = tag.split(".")
+        if (t_major, t_minor) < (major, minor):
+            prev = tag
```
kramaranya: What if we do patch release for 0.2.3, in this case it might find 0.3.0 as prev, since it only looks for tags where minor < 3, right?

scripts/gen-changelog.py
kramaranya, line 1: Do we even need this script? Can we just use git cliff --unreleased instead?

cliff.toml
```diff
@@ -0,0 +1,68 @@
...
+
+"""
+
+footer = "adsasd"
+
+
+[git]
+conventional_commits = false
+filter_unconventional = false
+split_commits = false
```
kramaranya: Currently script generates `### <!-- 0 -->🚀 Features`, I wonder it it's smth to do with conventional_commits? Can we update it to generate `### 🚀 Features` instead?

kramaranya (changes requested):
Thank you @jaiakash. 
I left a few comments

andreyvelich:
> make release VERSION=0.3

@kramaranya We agreed to pass the full version always, so you should run: `make release VERSION=0.3.0`

kramaranya:
> > @jaiakash have you tried to test it for 0.3.0?
> > This doesn't work for me:
> 
> Yes, please add tag 0.3.0 to your local working copy, can as you can see its says tag 0.3.0 doesnt exits. Here is sample release for me - https://github.com/jaiakash/sdk/releases/tag/0.3.0

@jaiakash why do we need a local tag to generate a changelog? In the release doc the process is that you generate a changelog, and [release workflow](https://github.com/kubeflow/sdk/blob/main/.github/workflows/release.yml) will create a tag for you. We should keep the same process and structure, just instead of maintaining our own script - use git cliff. 

> @kramaranya We agreed to pass the full version always, so you should run: make release VERSION=0.3.0

@andreyvelich yes, sorry the output was for 0.3.0, it still failed because of the tag

scripts/gen-changelog.py
andreyvelich: We can remove this script, but then we have to detect the previous version here before running cliff: https://github.com/kubeflow/sdk/blob/24208828239d3f65adb816bda5bbdae97a21abe2/Makefile#L74
I am fine if you can do that @jaiakash.

kramaranya: We can use `--tag-pattern` from git cliff for that

.github/workflows/release.yml
jaiakash: Sure, i will try the `release` workflow with updated commits.

jaiakash:
Hi @kramaranya @andreyvelich Address last few review comments, I have updated the PR. Here are changes - 

> why do we need a local tag to generate a changelog

Now, no tags for local release, users can run `make release VERSION=X.Y.Z` to generate the change log. `release.yml` will take care of rest in CI.

 -----------------

> Currently script generates ### 🚀 Features, I wonder it it's smth to do with conventional_commits? Can we update it to generate ### 🚀 Features instead?

No, conventional commits flag is needed to be false, to have non conventional commits in change log 
```
- fork: release (Akash Jaiswal)
- reveiew: fixes issue and fallback for new contributor page (Akash Jaiswal)
- Kubeflow SDK Official Release 0.2.1 ([#180](https://github.com/kubeflow/sdk/pull/180) by @kramaranya)
```
And yes i removed the index comments, ### 🚀 Features to 🚀 Features

 -----------------

> make release still updates the version, right? In this case revert these changes please

Yes done

 -----------------

> PR and Version Clickable

Yes, done

 -----------------

> Could you please verify the full release workflow with your changes and git cliff on your fork, before we cut a new release?

Yes here is the sample flow - https://github.com/jaiakash/sdk/actions/runs/21090885897/job/60661900497
Sample Release - https://github.com/jaiakash/sdk/releases/tag/0.4.0
Package on PyPi - https://pypi.org/project/kubeflow-sdk-akash/0.4.0/ 

 -----------------

### On the question of the script, 2 reason

#### Major Reason
There is an issue for new contributors list when we try to generate change log from git range and we try to generate the change log.
See: https://github.com/orhun/git-cliff/issues/1347 (Specially the logs which i shared there)

When we pass a commit range to `git-cliff`, the entire context is limited to that range. As a result, the new contributors list is calculated only within that release range.
For example, all contributors appearing in the 0.3.0 release would be treated as new contributors, even if they had already contributed to previous releases.

As a fallback, we currently rely only on --unreleased. Because of this limitation, the version has to be updated manually via the script.

#### Minor Reason

For patch releases, we are using the same changelog file, CHANGELOG-0.3.md, for all 0.3.x versions such as 0.3.0, 0.3.1, etc.
To support this workflow, we use the `--prepend` parameter so that each new release entry is added to the same change log file instead of creating separate files. I can have it in `makefile` itself, but that would complicate the code in `makefile`
https://git-cliff.org/docs/usage/examples

RELEASE.md
```diff
@@ -61,7 +61,8 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
 
    ```sh
    export GITHUB_TOKEN=<your_github_token>
```
andreyvelich, line 63: The GITHUB_TOKEN can be removed right?

RELEASE.md
```diff
@@ -61,7 +61,8 @@ The Kubeflow SDK uses an automated release process with GitHub Actions:
...
-   make release VERSION=X.Y.Z
+   make release VERSION=<X.Y.Z>
```
andreyvelich, line 64: We might need to say that users have to checkout to the release branch for patch releases before running `make release` cmd (e.g. `git checkout release-0.2`)

cliff.toml
```diff
@@ -0,0 +1,72 @@
...
+"""
+
+trim = true
+
+footer = ""
+
+[git]
+conventional_commits = false
+filter_unconventional = false
+split_commits = false
```
kramaranya: Can we set `filter_unconventional = true`? Currently we generate:

### 💼 Other

- reveiew: fixes issue and fallback for new contributor page (Akash Jaiswal)
- Kubeflow SDK Official Release 0.2.1 ([#180](https://github.com/kubeflow/sdk/pull/180) by @kramaranya)

Since all PRs must follow conventional commit format (feat/fix/chore/revert) and Kubeflow SDK release should not be in the chnagelog, I think we should remove it. wdyt @jaiakash @andreyvelich

andreyvelich: Yes, that is fine for me.

RELEASE.md
jaiakash: It wont be needed for CI since we do only once, but for local testing we might need due to rate limiting.
I suggest we should keep it like how it was before.

> [GitHub REST API](https://docs.github.com/en/rest) is being used to retrieve data from GitHub and it has the rate limit of **60 requests per hour** for unauthenticated users.
Although this is enough for a couple of runs of git-cliff, it is suggested that you create an access token to increase the request limit.

Source: https://git-cliff.org/docs/integration/github/#authentication

andreyvelich: Sounds good.

cliff.toml
jaiakash: Sure i will update it then.

RELEASE.md
andreyvelich: @jaiakash Have you checked this comment?

jaiakash: Wait, why do users have to checkout release branch?

Since we are running `git cliff --unreleased`, even for patch releases, They have to checkout `main` branch generate the patch change log for eg ( `0.3.1` from `0.3.0`).
Only thing is that, that PR will be merged to `release-0.3` branch instead of new release branch.

Let me know if there is any gap on my understanding.

cliff.toml
```diff
@@ -0,0 +1,72 @@
...
+
+{% raw %}\n{% endraw %}
+"""
+
+trim = true
+
+footer = ""
+
+[git]
+conventional_commits = false
```
kramaranya: This still generates chnagelog with Kubeflow SDK release PR, which should not be there. 
I think this should fix it:
```suggestion
conventional_commits = true
```
@jaiakash could you verify it and is there anything else we need to change to only include conventional commits with feat, fix, revert and chore prefixes?

jaiakash: Understood that part, but do we want to DROP any other commits _entirely_ from change log which dont follow the above 4 prefixes? 
@kramaranya 

Then we have to make the convetional_commits to `true` as you suggested.

kramaranya: Yes, because Kubeflow SDK release is the only commit that doesn't follow the rule with the prefixes. And this commit should not be in the changelog

PR #228, opened 2026-01-16
fix(trainer): Fix parsing for TrainJob events

andreyvelich:
We should convert response from listing events via:
```py
models.IoK8sApiCoreV1EventList.from_dict(event_response.to_dict())
```

That will return an error as follows:
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for IoK8sApiCoreV1Event
involvedObject
  Input should be a valid dictionary or instance of IoK8sApiCoreV1ObjectReference [type=model_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.12/v/model_type
```
Looking at v1 code, we also didn't do this: https://github.com/kubeflow/trainer/blob/release-1.9/sdk/python/kubeflow/training/api/training_client.py#L1387

We still need to do this for Pods, since container resources should be correctly parsed to the `IoK8sApiCoreV1ResourceRequirements`.

/assign @kubeflow/kubeflow-sdk-team @akshaychitneni @sksingh2005

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -427,19 +427,13 @@ class KubernetesBackend(RuntimeBackend):
         events = []
         try:
             # Retrieve events from the namespace
-            event_response = self.core_api.list_namespaced_event(
+            event_response: models.IoK8sApiCoreV1EventList = self.core_api.list_namespaced_event(
                 namespace=self.namespace,
                 async_req=True,
             ).get(common_constants.DEFAULT_TIMEOUT)
 
-            # Convert to event list
-            event_list = models.IoK8sApiCoreV1EventList.from_dict(event_response.to_dict())
-
-            if not event_list:
-                return events
-
             # Filter events related to this TrainJob or its pods
-            for event in event_list.items:
+            for event in event_response.items:
                 if not (event.metadata and event.involved_object and event.first_timestamp):
                     continue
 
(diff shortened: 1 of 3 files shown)


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -427,19 +428,15 @@ def get_job_events(self, name: str) -> list[types.Event]:
...
                 namespace=self.namespace,
                 async_req=True,
             ).get(common_constants.DEFAULT_TIMEOUT)
 
-            # Convert to event list
-            event_list = models.IoK8sApiCoreV1EventList.from_dict(event_response.to_dict())
-
-            if not event_list:
-                return events
+            print(event_response)
```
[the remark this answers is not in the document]
andreyvelich: No needed.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -428,28 +428,28 @@ def mock_list_namespaced_event(*args, **kwargs):
                     name="test-event-1",
                     namespace=DEFAULT_NAMESPACE,
                 ),
-                involved_object=models.IoK8sApiCoreV1ObjectReference(
+                involvedObject=models.IoK8sApiCoreV1ObjectReference(
```
Shekharrajak, line 431: this changes were not needed ?

andreyvelich: Actually, it used CamelCase by default for models.
You can check example here: https://github.com/kubeflow/sdk/blob/d9c8e7ebc2602999c480e4f2b58935560935afab/kubeflow/trainer/backends/kubernetes/backend_test.py#L543-L555

jaiakash (commented):
LGTM

Issue #227, opened 2026-01-16
Parse port from base url in ModelRegistryClient


jonburdo:
### What you would like to be added?

Support parsing port from `base_url` in `ModelRegistryClient`. This should be added to the internal client from the `model-registry` library as described in this issue: https://github.com/kubeflow/model-registry/issues/2110

Then we will have it for free here for newer versions of `model-registry` (`kubeflow[hub]`). We should try to avoid doing the implementation in this repo as that would mean we doing url parsing in `ModelRegistryClient` and then again in `ModelRegistry` - this would be complicated and more bug-prone.

The criteria for this issue can be:
- [ ] handle `model-registry` version differences
  - preferably just update the minimum version of `model-registry` to a version that has this feature one the issue above is done
  - alternatively include warnings in code and docs about older version of model-registry missing this feature
- [ ] add examples to README/docs of ways to handle port
  ```python
  ModelRegistryClient("https://example.org", port=456)  # port kwarg
  ModelRegistryClient("https://example.org:456")        # base_url
  ModelRegistryClient("https://example.org")            # default (`443` for https, `8080` for http)
  ```
- [ ] document in `ModelRegistryClient.__init__` docstring
- [ ] add test cases for this

### Why is this needed?

If I have `model_registry_url = "https://example.org:456"` (port in the url), then currently I have to separate this out to do:
```python
ModelRegistryClient(base_url="https://example.org", port=456)
```
Ideally I would like to be able to simply do:
```python
ModelRegistryClient(base_url=model_registry_url)
```

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

andreyvelich:
@osamaahmed17 We might need to wait until we fix it in the Model Registry client upstream, but feel free to help with this issue first: https://github.com/kubeflow/model-registry/issues/2110

osamaahmed17:
@andreyvelich , sure. I will look upon this issue at the moment [2110](https://github.com/kubeflow/model-registry/issues/2110)

jonburdo:
This is complete since this is solved in model-registry and we've updated the minimum version here

PR #243, opened 2026-01-26
feat: Added examples to the documentation demonstrating different ways to handle ports

osamaahmed17:
Added examples to the documentation demonstrating different ways to handle ports
feat: #227

### DIFF
--- a/kubeflow/hub/api/model_registry_client_test.py
+++ b/kubeflow/hub/api/model_registry_client_test.py
@@ -97,8 +97,12 @@ def test_init_import_error(test_case, monkeypatch):
                 "author": "test",
             },
             expected_output={
+                "server_address": "http://localhost",
                 "port": 8080,
+                "author": "test",
                 "is_secure": False,
+                "user_token": None,
+                "custom_ca": None,
             },
         ),
         TestCase(
@@ -109,8 +113,12 @@ def test_init_import_error(test_case, monkeypatch):
                 "author": "test",
             },
             expected_output={
+                "server_address": "https://registry.example.com",
                 "port": 443,
+                "author": "test",
                 "is_secure": True,
+                "user_token": None,
+                "custom_ca": None,
             },
         ),
         TestCase(
@@ -122,8 +130,12 @@ def test_init_import_error(test_case, monkeypatch):
                 "author": "test-author",
             },
             expected_output={
+                "server_address": "http://localhost",
                 "port": 9080,
+                "author": "test-author",
                 "is_secure": False,
+                "user_token": None,
+                "custom_ca": None,
             },
         ),
         TestCase(
@@ -134,8 +146,12 @@ def test_init_import_error(test_case, monkeypatch):
                 "author": "test",
             },
             expected_output={
-                "port": 456,
+                "server_address": "https://example.org:456",
+                "port": 443,
+                "author": "test",
                 "is_secure": True,
+                "user_token": None,
+                "custom_ca": None,
             },
         ),
         TestCase(
@@ -146,8 +162,12 @@ def test_init_import_error(test_case, monkeypatch):
                 "author": "test",
             },
             expected_output={
-                "port": 456,
+                "server_address": "http://example.org:456",
+                "port": 8080,
+                "author": "test",
                 "is_secure": False,
+                "user_token": None,
+                "custom_ca": None,
             },
         ),
     ],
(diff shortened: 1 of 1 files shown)


osamaahmed17:
@andreyvelich and @jonburdo, kindly review the code and let me know if anything is missing.
All tasks are completed only one remains, waiting for the new model registry version.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -67,6 +67,10 @@ def __init__(
 
         Raises:
             ImportError: If model-registry is not installed.
+        Examples:
+            ModelRegistryClient("https://example.org", port=456)  # port kwarg
+            ModelRegistryClient("https://example.org:456")        # base_url
+            ModelRegistryClient("https://example.org")            # default (`443` for https, `8080` for http)
```
jonburdo:
```suggestion
            ModelRegistryClient("https://example.org", port=456)  # port kwarg
            ModelRegistryClient("https://example.org:456")        # base_url (including port)
            ModelRegistryClient("https://example.org")            # default port (`443` for https, `8080` for http)
```
Just a suggestion

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -126,6 +126,30 @@ def mock_import(name, *args, **kwargs):
...
+            name="https URL with explicit port parses port",
+            expected_status=SUCCESS,
+            config={
+                "base_url": "https://example.org:456",
+                "author": "test",
+            },
+            expected_output={
+                "port": 456,
+                "is_secure": True,
+            },
```
jonburdo, line 155: The `port` kwarg will be the default since we just pass through `base_url`, which is why we get this failure: https://github.com/kubeflow/sdk/actions/runs/21444907485/job/61757845124?pr=243#step:5:320

At a minimum I think the `TestCase` needs to be adjusted, and in the test body we should also check that the `base_url` is passed as an arg to `mock_registry_class`.

Maybe the `test_init` test should actually be adjusted to do a single:
```python
        assert mock_registry_class.call_args == test_case.expected_output
```
Instead of:
```
        call_kwargs = mock_registry_class.call_args[1]
        assert call_kwargs["port"] == test_case.expected_output["port"]
        assert call_kwargs["is_secure"] is test_case.expected_output["is_secure"]
```
This would require adjusting each `TestCase`

README.md
```diff
@@ -120,6 +120,16 @@ optimization_id = OptimizerClient().optimize(
...
 ```
 
+### Interact with Model Registry
+
+```python
+from kubeflow.hub import ModelRegistryClient
+
+ModelRegistryClient("https://example.org", port=456)  # port kwarg
+ModelRegistryClient("https://example.org:456")        # base_url
+ModelRegistryClient("https://example.org")            # default (`443` for https, `8080` for http)
```
jonburdo: I know I mentioned README/docs in the issue, but I'm not sure we need a new section in the README just for this. The docstring might be enough. If we do want this, I would maybe just show this as additional ways to create a `ModelRegistryClient` at the bottom of the section below (`### Manage models with Model Registry`) so the quickstart (`pip install` / import / use client) is first.

Similar to my comment on the docstring, the code comments here should be more specific about referring to the port if we keep this

osamaahmed17: Thanks a lot @jonburdo for helping me out.

README.md
```diff
@@ -156,6 +156,14 @@ for version in client.list_model_versions("my-model"):
     print(f"Version: {version.name}")
 ```
 
+You can also initialize the client using different port configurations:
```
jonburdo, line 159: Perfect! Succinct and clear :)

kubeflow/hub/api/model_registry_client_test.py
```diff
@@ -144,9 +188,7 @@ def test_init(test_case, monkeypatch):
 
         assert test_case.expected_status == SUCCESS
         mock_registry_class.assert_called_once()
-        call_kwargs = mock_registry_class.call_args[1]
-        assert call_kwargs["port"] == test_case.expected_output["port"]
-        assert call_kwargs["is_secure"] is test_case.expected_output["is_secure"]
+        assert mock_registry_class.call_args[1] == test_case.expected_output
```
jonburdo, line 191: Thanks, this makes the test more clear and more thorough

jonburdo:
Ah, we will also need to wait for the min python version of this project to be updated to python 3.10 since `model-registry==0.3.6` supports 3.10 and later

osamaahmed17:
@jonburdo, how do we merge the code, is there a direct command that I can do for future cases?

PR #249, opened 2026-01-29
fix(trainer): simplify local-exec command building process

### LINKED ISSUE #93: simply local-exec backend command building process
At the moment local-exec uses a single job script template and substitute variables inside that template. K8S backend uses a completey different approach to handle the command building. 

We need to simply the command building process for the local-exec backend.

khushiiagrawal:

Different from the previous approach of using complex Bash script templates, this PR simplifies the `local-exec` command building process by introducing a generated Python runner script (`runner.py`). This improves maintainability, debugging, and cross-platform compatibility.

Specific changes include:
- **Refactored Command Generation**: Replaced `LOCAL_EXEC_JOB_TEMPLATE` (Bash) with `RUNNER_TEMPLATE` (Python) in `constants.py`.
- **Updated Utils**: Modified `utils.py` to generate the `runner.py` script which handles virtual environment creation, dependency installation, and training command execution using Python's `subprocess`.
- **Bug Fix**: Corrected `LocalProcessBackend.__get_job_status` in `backend.py` to properly return `TRAINJOB_COMPLETE` instead of defaulting to `TRAINJOB_CREATED`, ensuring `wait_for_job_status` functions correctly.
- **Testing**: Added `utils_test.py` to verify the command generation logic.

Fixes #93

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -267,7 +267,7 @@ class LocalProcessBackend(RuntimeBackend):
         elif constants.TRAINJOB_CREATED in statuses:
             status = constants.TRAINJOB_CREATED
         else:
-            status = constants.TRAINJOB_CREATED
+            status = constants.TRAINJOB_COMPLETE
 
         return status
 
--- a/kubeflow/trainer/backends/localprocess/constants.py
+++ b/kubeflow/trainer/backends/localprocess/constants.py
@@ -41,48 +41,69 @@ local_runtimes = [
 ]
 
 
+
+TORCH_COMMAND = "torchrun"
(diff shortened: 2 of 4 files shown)


khushiiagrawal:
hii @kramaranya  @astefanutti  

I noticed that the  uv.lock file in the main branch seems to be corrupted. When I try to run the tests locally to verify my changes, uv lock --check fails with a syntax error (missing field version at line 921).

I haven't included the lockfile regeneration in this commit yet, but as a result, the tests are currently failing locally (and likely in CI if it checks the lockfile).

Could you please confirm if I should regenerate and commit the fixed  uv.lock  in this PR ? This seems necessary to get the tests passing.
Thanks!

Issue #263, opened 2026-02-04
Big letter in python script will change to small letter in Trainjob yaml, finally cause npu resource cannot allocate.


wangyakun:
### What happened?

I started a trainer job uses npu by python sdk, the key code related to this issue is here:

```
job_id = TrainerClient().train(
    runtime=TrainerClient().get_runtime(name='custom-minimize-npu'),
    trainer=CustomTrainerContainer(
        # image='36.134.128.101.nip.io:31104/aict/yolov9:20240802',
        image='36.134.128.101.nip.io:31104/aict-2025/llamafactory:0.9.4-npu-a2-tensorboard-bnb',
        num_nodes=1,
        resources_per_node={
            "cpu": 2,
            "memory": "16Gi",
            "huawei.com/Ascend910": "1",
            # "gpu": 1, # Comment this line if you don't have GPUs.
        },
    ),
...
```

<img width="1775" height="1306" alt="Image" src="https://github.com/user-attachments/assets/9a123d88-04fd-40aa-998a-343d14c66ec5" />

Then I run this script, and check the detail of trainjob, I found the letter 'A' is changed to 'a':
(trainjob yaml:)
```
...
kind: TrainJob
spec:
  trainer:
    resourcesPerNode:
      limits:
        cpu: "2"
        huawei.com/ascend910: "1"
        memory: 16Gi
      requests:
        cpu: "2"
        huawei.com/ascend910: "1"
        memory: 16Gi
...
```

<img width="2534" height="1236" alt="Image" src="https://github.com/user-attachments/assets/82b494e1-0082-4987-a534-a4a3eab83cac" />

This will eventually cause the pod to fail to be allocated resources normally:

<img width="2481" height="261" alt="Image" src="https://github.com/user-attachments/assets/90b780d8-f119-4d36-98c0-f364e6cf6839" />

### related code

This code may need to be modified to fix this issue:

sdk\kubeflow\trainer\backends\kubernetes\utils.py

<img width="952" height="478" alt="Image" src="https://github.com/user-attachments/assets/8f770c81-2293-4cb5-8346-46afb90aa4c9" />

### What did you expect to happen?

I expect the letter 'A' don't change to 'a', so npu resource will be properly allocated.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
result:
Client Version: v1.34.3+k3s1
Kustomize Version: v5.7.1
Server Version: v1.34.3+k3s1
---
Kubeflow Trainer version:
```bash
[...]

PR #264, opened 2026-02-04
fix: preserve case for extended resource keys

danish9039:
`fix: preserve case for extended resource keys` 
- fix https://github.com/kubeflow/sdk/issues/263

This PR fixes a bug in the Kubeflow Trainer SDK where resource keys were being unconditionally lowercased. While this is acceptable for standard Kubernetes resources (like `cpu` and `memory`), it causes resource allocation failures for **Extended Resources** (e.g., `huawei.com/Ascend910`) which are case-sensitive.

### The Problem
Device plugins (like the Huawei Ascend plugin) register resources with specific casing. When a user specifies `huawei.com/Ascend910` in their [resources_per_node](cci:1://file:///home/danish/opensource/kubeflow/sdk/kubeflow/trainer/backends/kubernetes/utils.py:222:0-260:20), the SDK previously converted this to `huawei.com/ascend910`. Since the Kubernetes control plane is looking for the exact string, it would fail to find the resource, leaving the `TrainJob` pods in a `Pending` state.

### Proposed Changes
Modified [sdk/kubeflow/trainer/backends/kubernetes/utils.py](cci:7://file:///home/danish/opensource/kubeflow/sdk/kubeflow/trainer/backends/kubernetes/utils.py:0:0-0:0) to implement a selective lowercasing strategy:
1.  **Whitelist normalization**: Only lowercase keys that are standard Kubernetes resources (`cpu`, `memory`, `ephemeral-storage`) or known aliases (`gpu`, `mig-*`).
2.  **Case preservation**: For all other keys (extended resources), maintain the original casing provided by the user.

### Resource Transformation Logic

```mermaid
graph TD
    User["User Input (Python dict)"] -->|"e.g., {'huawei.com/Ascend910': '1'}"| SDK["SDK Logic (get_resources_per_node)"]
    
    subgraph Buggy_Logic ["Before Fix (Indiscriminate Lowercasing)"]
    SDK -->|"k.lower()"| LowerCase["All keys lowercased"]
    LowerCase -->|Result| FailedYAML["TrainJob YAML: huawei.com/ascend910"]
    end
    
    subgraph Fixed_Logic ["After Fix (Selective Normalization)"]
    SDK -->|"Check if standard?"| Branch{Standard / Alias?}
    Branch -->|Yes| Lower["Lowercase (cpu, memory, etc.)"]
    Branch -->|No| Preserve["Preserve Case (huawei.com/Ascend910)"]
    Lower -->|Result| GoodYAML["TrainJob YAML: Correct keys"]
    Preserve -->|Result| GoodYAML
    end

    FailedYAML -->|"Scheduling Error"| Fail["NPU Not Allocated (Mismatch)"]
    GoodYAML -->|"Scheduling Success"| Success["NPU Allocated Successfully"]

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -219,7 +219,6 @@ def get_trainjob_node_step(
     return step
 
 
-# TODO (andreyvelich): Discuss if we want to support V1ResourceRequirements resources as input.
 def get_resources_per_node(
     resources_per_node: dict,
 ) -> models.IoK8sApiCoreV1ResourceRequirements:
@@ -228,19 +227,15 @@ def get_resources_per_node(
     """
 
     # Convert only standard resource keys and aliases to lowercase.
+    # Extended resources (e.g., "example.com/Custom-NPU") preserve their original case.
+    standard_resources = {constants.CPU_LABEL, "memory", "gpu", "storage", "ephemeral-storage"}
+    resource_aliases = {"gpu": "nvidia.com/gpu", "storage": "ephemeral-storage"}
+
     resources = {}
     for k, v in resources_per_node.items():
-        if k.lower() in {
-            constants.CPU_LABEL,
-            "memory",
-            "ephemeral-storage",
-            "gpu",
-        } or k.lower().startswith("mig-"):
-            resources[k.lower()] = models.IoK8sApimachineryPkgApiResourceQuantity(v)
-        else:
-            resources[k] = models.IoK8sApimachineryPkgApiResourceQuantity(v)
-    if "gpu" in resources:
-        resources["nvidia.com/gpu"] = resources.pop("gpu")
+        key = k.lower() if k.lower() in standard_resources or k.lower().startswith("mig-") else k
+        key = resource_aliases.get(key, key)
+        resources[key] = models.IoK8sApimachineryPkgApiResourceQuantity(v)
 
     # Optional alias for MIG: "mig-<profile>" -> "nvidia.com/mig-<profile>"
     # Example: "mig-1g.5gb" -> "nvidia.com/mig-1g.5gb"
@@ -256,11 +251,10 @@ def get_resources_per_node(
             f"GPU (nvidia.com/gpu) and MIG ({mig_keys[0]}) cannot be requested together"
         )
 
-    resources = models.IoK8sApiCoreV1ResourceRequirements(
+    return models.IoK8sApiCoreV1ResourceRequirements(
         requests=resources,
         limits=resources,
     )
-    return resources
 
 
 def get_script_for_python_packages(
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: kubeflow/trainer/backends/kubernetes/utils_test.py [hunk 1]
```diff
@@ -117,6 +117,55 @@ def test_get_container_devices(test_case: TestCase):
...
+            config={
+                "resources_per_node": {
+                    "huawei.com/Ascend910": 1,
+                    "CPU": 2,
+                    "Memory": "16Gi",
+                }
+            },
+            expected_output=models.IoK8sApiCoreV1ResourceRequirements(
+                limits={
+                    "huawei.com/Ascend910": models.IoK8sApimachineryPkgApiResourceQuantity(1),
```
andreyvelich: @tenzen-y previously suggested that we don't use specific vendor names in our tests, so you can just set it as:
```suggestion
                    "example.com/Capitalized": models.IoK8sApimachineryPkgApiResourceQuantity(1),
```

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -117,6 +117,55 @@ def test_get_container_devices(test_case: TestCase):
...
+                    "memory": models.IoK8sApimachineryPkgApiResourceQuantity("16Gi"),
+                },
+            ),
+        ),
+        TestCase(
+            name="diverse resource types and mixed case standard keys",
+            expected_status=SUCCESS,
+            config={
+                "resources_per_node": {
+                    "aws.amazon.com/neuron": 1,
```
andreyvelich: same here
```suggestion
                    "example.com/test": 1,
```

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -227,11 +227,18 @@ def get_resources_per_node(
...
-    }
+    # Convert only standard resource keys and aliases to lowercase.
+    resources = {}
+    for k, v in resources_per_node.items():
+        if k.lower() in {
+            constants.CPU_LABEL,
+            "memory",
+            "ephemeral-storage",
+            "gpu",
+        } or k.lower().startswith("mig-"):
```
andreyvelich: @kubeflow/kubeflow-sdk-team @wangyakun Does this check look good to you?

andreyvelich (commented):
Thanks for the fix @danish9039!

kubeflow/trainer/backends/kubernetes/utils_test.py
danish9039: > @tenzen-y previously suggested that we don't use specific vendor names in our tests, so you can just set it as:

Got it !

kubeflow/trainer/backends/kubernetes/utils.py
astefanutti: I think that's ok, we can iterate if needed.

szaher: should we rename `ephemeral-storage` to just `storage` for simplicity? also, assuming it will only be ephemeral storage and other storage types can be passed via options?

andreyvelich: sure, @danish9039 let's start with `storage` first.

PR #266, opened 2026-02-05
chore: add trivy cve scan and fix workflow

Fiona-Waters:

This PR adds a workflow file that uses trivy for CVE scanning. The workflow will also open a PR on the repo with a fix for any CVEs with CVSS of 7.0 or higher that have fixes available. It runs nightly.

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/trivy-cve-scan.yaml
+++ b/.github/workflows/trivy-cve-scan.yaml
@@ -44,7 +44,7 @@ jobs:
         uses: github/codeql-action/upload-sarif@v4
         with:
           sarif_file: 'trivy-results.sarif'
-
+          category: kubeflow-sdk-trivy-scanner
       - name: Process CVEs and Apply Fixes
         id: fixer
         run: |
@@ -97,4 +97,3 @@ jobs:
           delete-branch: true
           labels: |
             "area/security"
-            automated-fix


.github/workflows/trivy-cve-scan.yaml
```diff
@@ -0,0 +1,91 @@
...
+            ${{ steps.fixer.outputs.fix_details }}
+
+            *(Note: Items above are formatted as `Package==Version | URL`)*
+
+            **Verification:** Updated via `uv lock --upgrade-package`. Please review the diff and ensure PR checks pass before merging.
+          branch: security-nightly-updates-${{ github.run_id }}
+          delete-branch: true
+          labels: |
+            security
+            automated-fix
```
andreyvelich: Let's apply /area security label that I created: https://github.com/kubeflow/sdk/issues?q=state%3Aopen%20label%3Aarea%2Fsecurity

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Run Trivy FS Scan
+        uses: aquasecurity/trivy-action@0.33.1
+        with:
+          scan-type: 'fs'
+          format: 'json'
+          output: 'trivy-results.json'
+          severity: 'HIGH,CRITICAL'
+          ignore-unfixed: true
+
+      - name: Process CVEs and Apply Fixes
```
andreyvelich, line 48: Shall we also upload unresolved CVEs that Trivy detect to GitHub?
https://github.com/kubeflow/trainer/blob/master/.github/workflows/trivy-scan.yaml#L24

Fiona-Waters: We can. We would need to run the trivy scan twice, once with json output so we can parse it to apply fixes and once with a SARIF file output to use in github.

Fiona-Waters: Updated for this. ptal, thanks

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -0,0 +1,100 @@
...
+            | Package & Version | Advisory Link |
+            | :--- | :--- |
+            ${{ steps.fixer.outputs.fix_details }}
+
+            **Verification:** Updated via `uv lock --upgrade-package`.
+          branch: security-nightly-updates-${{ github.run_id }}
+          delete-branch: true
+          labels: |
+            "area/security"
+            automated-fix
```
andreyvelich: We don't have this label in `kubeflow/sdk`.
```suggestion
```

PR #267, opened 2026-02-06
fix: Improve CVE workflow

Fiona-Waters:

Following on from the trivy-cve-scan workflow that was added yesterday I have updated it with the following:

 - It has come to my attention that purely updating the uv.lock is not sufficient as anyone updating the file in another PR can revert this change. In order to counter that I have added a verification step that checks if a "natural" lockfile upgrade is blocked by parent dependencies. If blocked, the bot now automatically injects a tool.uv.override-dependencies into pyproject.toml to ensure the fix cannot be accidentally reverted by manual uv lock commands.

Last nights run found the urllib CVE, which I had already fixed - but another PR had reverted the change.

 - Implemented add-paths filtering in the PR creation step. This ensures only uv.lock and pyproject.toml are committed and nothing else. This was previously missing causing the workflow to fail.

Please review @andreyvelich @kramaranya 

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/cleanup-overrides.yaml
+++ b/.github/workflows/cleanup-overrides.yaml
@@ -84,7 +84,7 @@ jobs:
             fi
 
             # Check the resulting version
-            NATURAL_VER=$(uv tree --package "$PACKAGE" | grep -F "$PACKAGE v" | grep -oP 'v\K[0-9.]+' | head -n 1)
+            NATURAL_VER=$(uv tree --package "$PACKAGE" | uv run python3 .github/scripts/extract_version.py "$PACKAGE")
 
             if [ -z "$NATURAL_VER" ]; then
               echo "Warning: Could not determine natural version, keeping override"
@@ -97,19 +97,21 @@ jobs:
             echo "Natural upgrade resulted in version: $NATURAL_VER"
 
             # Compare versions: if natural version >= override version, we can remove override
-            if [ "$NATURAL_VER" = "$OVERRIDE_VER" ] || \
-               [ "$(printf '%s\n' "$NATURAL_VER" "$OVERRIDE_VER" | sort -V | tail -n1)" = "$NATURAL_VER" ]; then
-              echo "SUCCESS: Natural upgrade now works! Version $NATURAL_VER >= $OVERRIDE_VER"
-              echo "  Removing override for $TARGET"
-              echo "- $TARGET (upgraded to $NATURAL_VER)" >> removed.txt
-              # Keep the modified files (override removed)
-              rm -f pyproject.toml.backup uv.lock.backup
-            else
+            # compare_versions.py exits 0 if first arg < second arg, so we need to check if NATURAL_VER < OVERRIDE_VER
+            if uv run python3 .github/scripts/compare_versions.py "$NATURAL_VER" "$OVERRIDE_VER"; then
+              # NATURAL_VER < OVERRIDE_VER, so we need to keep the override
               echo "SKIP: Natural upgrade insufficient. Version $NATURAL_VER < $OVERRIDE_VER"
               echo "  Keeping override for $TARGET"
               mv pyproject.toml.backup pyproject.toml
               mv uv.lock.backup uv.lock
               echo "- $TARGET (natural upgrade only reached $NATURAL_VER, need $OVERRIDE_VER)" >> kept.txt
+            else
+              # NATURAL_VER >= OVERRIDE_VER, we can remove the override
+              echo "SUCCESS: Natural upgrade now works! Version $NATURAL_VER >= $OVERRIDE_VER"
+              echo "  Removing override for $TARGET"
+              echo "- $TARGET (upgraded to $NATURAL_VER)" >> removed.txt
+              # Keep the modified files (override removed)
+              rm -f pyproject.toml.backup uv.lock.backup
             fi
           done <<< "$OVERRIDES"
 
@@ -182,4 +184,4 @@ jobs:
           branch: cleanup-overrides-${{ github.run_id }}
           delete-branch: true
           labels: |
-            "area/dependencies"
+            "area/security"
--- a/.github/workflows/trivy-cve-scan.yaml
+++ b/.github/workflows/trivy-cve-scan.yaml
@@ -81,7 +81,7 @@ jobs:
 
             # Verify if the upgrade actually worked in the lockfile
             # Extract full version including pre-release, post-release, dev, etc.
-            CURRENT_VER=$(uv tree --package "$PACKAGE" | grep -F "$PACKAGE v" | sed -n 's/.*v\([^ ]*\).*/\1/p' | head -n 1)
+            CURRENT_VER=$(uv tree --package "$PACKAGE" | uv run python3 .github/scripts/extract_version.py "$PACKAGE")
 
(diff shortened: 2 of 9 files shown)


Fiona-Waters:
@andreyvelich please take a look when you can. Thanks

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -68,11 +68,33 @@ jobs:
...
+
+            if [[ "$CURRENT_VER" < "$VERSION" ]]; then
+              echo "Natural upgrade blocked by parents. Adding Hard Pin to pyproject.toml overrides..."
+
+              if ! grep -q "\[tool.uv\]" pyproject.toml; then
+                echo -e "\n[tool.uv]\noverride-dependencies = [\"$TARGET\"]" >> pyproject.toml
+              elif ! grep -q "override-dependencies =" pyproject.toml; then
+                sed -i "/\[tool.uv\]/a override-dependencies = [\"$TARGET\"]" pyproject.toml
+              else
+                sed -i "/override-dependencies = \[/a \    \"$TARGET\"," pyproject.toml
```
szaher: How do we plan to manage these overrides? 
we should monitor them closely as they may hide real conflicts in the libraries that uses them.

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -68,11 +68,33 @@ jobs:
...
-          # Apply fixes via uv
+          # Apply fixes with "Natural Upgrade then Hard Pin" logic
           echo "$FIX_DATA" | while read -r line; do
+            PACKAGE=$(echo "$line" | cut -d'=' -f1 | xargs)
+            VERSION=$(echo "$line" | cut -d'|' -f1 | cut -d'=' -f3 | xargs)
             TARGET=$(echo "$line" | cut -d'|' -f1 | xargs)
-            echo "Applying fix: uv lock --upgrade-package $TARGET"
+
+            echo "Attempting natural upgrade for: $TARGET"
             uv lock --upgrade-package "$TARGET"
```
kramaranya, line 80: I wonder if we should upgrade the package name only first, then check if it reached the fixed version?

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -68,11 +68,33 @@ jobs:
...
+
+            # Verify if the upgrade actually worked in the lockfile
+            CURRENT_VER=$(uv tree --package "$PACKAGE" | grep -oP "$PACKAGE v\K[0-9.]+" | head -n 1)
+
+            if [[ "$CURRENT_VER" < "$VERSION" ]]; then
+              echo "Natural upgrade blocked by parents. Adding Hard Pin to pyproject.toml overrides..."
```
kramaranya: Would bash compare the strings correctly? For example, `"0.10.0" < "0.2.0"`?

kramaranya (commented):
Thanks @Fiona-Waters! I left a few comments

.github/workflows/trivy-cve-scan.yaml
Fiona-Waters: It must reach the fixed version so I think it makes sense just to go straight to it. I'm not sure what value there would be in doing it differently?

.github/workflows/trivy-cve-scan.yaml
Fiona-Waters: Good catch! Will update to use Sort -V instead

.github/workflows/trivy-cve-scan.yaml
Fiona-Waters: I think we should update the workflow to add dated comments when adding overrides so it is clear where they came from. 
We can also create a new workflow for periodic override cleanup checking if any parent packages have been updated allowing for natural upgrades. We could run this monthly. wdyt?

Fiona-Waters: Added a workflow here https://github.com/kubeflow/sdk/pull/267/changes/72eb2b2a0fd4b36c54ccd8d83719889a3a95a7c4

.github/workflows/trivy-cve-scan.yaml
```diff
@@ -68,11 +68,58 @@ jobs:
...
+                echo "override-dependencies = [\"$TARGET\"]" >> pyproject.toml
+              elif ! grep -q "override-dependencies =" pyproject.toml; then
+                # Add header comment if this is the first override
+                sed -i "/\[tool.uv\]/a # Security overrides - Review periodically and remove if parent constraints allow natural upgrade" pyproject.toml
+                sed -i "/\[tool.uv\]/a $COMMENT" pyproject.toml
+                sed -i "/\[tool.uv\]/a override-dependencies = [\"$TARGET\"]" pyproject.toml
+              else
+                # Add comment before appending to existing override-dependencies
+                sed -i "/override-dependencies = \[/i $COMMENT" pyproject.toml
+                sed -i "/override-dependencies = \[/a \    \"$TARGET\"," pyproject.toml
```
kramaranya: I'm concerned about multiple overrides, as it might break after the first entry? WDYT about instead always store it in multi-line format and rewrite the whole list?

Fiona-Waters: I'm updating that as we speak :) thanks!

kramaranya: reading my mind :) Thanks Fiona!

Fiona-Waters: I've updated to use python rather than bash, its more robust for what we need.

.github/workflows/cleanup-overrides.yaml
```diff
@@ -0,0 +1,187 @@
+name: Cleanup Dependency Overrides
+
+on:
+  schedule:
+    - cron: '0 3 1 * *'  # Monthly on the 1st at 3 AM UTC
+  workflow_dispatch:
+
+jobs:
+  cleanup-overrides:
```
andreyvelich, line 9: Can we run this Job in the same workflow: trivy-cve-scan?

.github/workflows/cleanup-overrides.yaml
```diff
@@ -0,0 +1,187 @@
...
+            These overrides are still needed because parent dependency constraints prevent natural upgrades to the required security versions.
+
+            ---
+            **Next Steps:**
+            - Review the changes and verify tests pass
+            - Consider investigating why kept overrides are still blocked (check parent dependency versions)
+          branch: cleanup-overrides-${{ github.run_id }}
+          delete-branch: true
+          labels: |
+            "area/dependencies"
```
andreyvelich:
```suggestion
            "area/security"
```

.github/workflows/cleanup-overrides.yaml
Fiona-Waters: We could but IMO keeping them separate might be more efficient and would help create clearer PRs that focus on either adding new security fixes or cleaning up old overrides. I have set the cleanup-overrides workflow to run monthly since testing each override is resource-intensive and parent dependency constraints typically don't change often enough to need daily cleanup. wdyt?

andreyvelich: sgtm

andreyvelich (commented):
/lgtm
/approve

PR #269, opened 2026-02-06
fix: improve logging around packages_to_install

### LINKED ISSUE #268: Improve pip install logging for runtime package installation
Improvement to logging on pip install <packages_to_install>

### Why is this needed?
When installing extra Python packages in the Kubernetes backend (get_script_for_python_packages), we currently run:
pip install --user ... || pip install ...
This causes the first pip install --user failure to be logged, even if the second non---user install succeeds, leading to confusing error messages in otherwise successful runs.

### Solution
Wrap both pip install attempts in explicit shell logic that:
Captures pip output into /tmp/pip_install.log for both attempts.
Treats the installation as successful if either attempt (with or without --user) succeeds, and prints only a short message:
Successfully installed Python packages: <packages>.
Prints an error message and the captured log only if both attempts fail:
ERROR: Failed to install Python packages: <packages> followed by the pip output.

### Love this feature?

briangallagher:
Improve pip install logging for runtime package installation #268

Fixes #268 

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -15,6 +15,7 @@
 from dataclasses import fields
 import inspect
 import os
+import shlex
 import textwrap
 from typing import Any, Callable, Optional, Union
 from urllib.parse import urlparse
@@ -259,15 +260,21 @@ def get_resources_per_node(
 def get_script_for_python_packages(
     packages_to_install: list[str],
     pip_index_urls: list[str],
+    install_log_file: str = "pip_install.log",
 ) -> str:
     """
     Get init script to install Python packages from the given pip index URLs.
     """
-    packages_str = " ".join(packages_to_install)
+    # Quote package names and URLs with shlex.quote() to prevent shell injection;
+    # each value becomes a single safe shell token when expanded inside the bash script.
+    packages_str = " ".join(shlex.quote(pkg) for pkg in packages_to_install)
 
     # first url will be the index-url.
-    options = [f"--index-url {pip_index_urls[0]}"]
-    options.extend(f"--extra-index-url {extra_index_url}" for extra_index_url in pip_index_urls[1:])
+    options = [f"--index-url {shlex.quote(pip_index_urls[0])}"]
+    options.extend(
+        f"--extra-index-url {shlex.quote(extra_index_url)}"
+        for extra_index_url in pip_index_urls[1:]
+    )
     options_str = " ".join(options)
 
     header_script = textwrap.dedent(
@@ -286,7 +293,7 @@ def get_script_for_python_packages(
         f"""
         PACKAGES="{packages_str}"
         PIP_OPTS="{options_str}"
-        LOG_FILE=/tmp/pip_install.log
+        LOG_FILE="{install_log_file}"
         rm -f "$LOG_FILE"
 
         if PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
@@ -355,6 +362,9 @@ def get_command_using_train_func(
     # The default file location for OpenMPI is: /home/mpiuser/<FILE_NAME>.py
     if is_mpi:
         func_file = os.path.join(constants.DEFAULT_MPI_USER_HOME, func_file)
+        install_log_file = os.path.join(constants.DEFAULT_MPI_USER_HOME, "pip_install.log")
+    else:
+        install_log_file = "pip_install.log"
 
     # Install Python packages if that is required.
     install_packages = ""
@@ -362,6 +372,7 @@ def get_command_using_train_func(
         install_packages = get_script_for_python_packages(
             packages_to_install,
             pip_index_urls,
+            install_log_file=install_log_file,
         )
 
     # Add function code to the Trainer command.
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -228,6 +228,7 @@ def get_custom_trainer(
     install_script = utils.get_script_for_python_packages(
         packages_to_install=packages_to_install,
         pip_index_urls=pip_index_urls,
+        install_log_file="pip_install.log",
     )
 
     # Append the embedded training function script that matches EXEC_FUNC_SCRIPT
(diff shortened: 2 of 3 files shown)


kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -278,18 +279,28 @@ def get_script_for_python_packages(
...
+
+        if PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+            --no-warn-script-location $PIP_OPTS --user $PACKAGES >"$LOG_FILE" 2>&1; then
+            echo "Successfully installed Python packages: $PACKAGES"
+        elif PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+            --no-warn-script-location $PIP_OPTS $PACKAGES >"$LOG_FILE" 2>&1; then
+            echo "Successfully installed Python packages: $PACKAGES"
+        else
+            echo "ERROR: Failed to install Python packages: $PACKAGES" >&2
+            cat "$LOG_FILE" >&2
```
kramaranya, line 307: shall we have exit 1 here?

kramaranya, line 307: Looks like we overwrite the first attempt's output with the second. Can we append the second output?

briangallagher: I think we should, updated.

briangallagher: Good catch, updated now

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -268,6 +268,7 @@ def get_script_for_python_packages(
     # first url will be the index-url.
     options = [f"--index-url {pip_index_urls[0]}"]
     options.extend(f"--extra-index-url {extra_index_url}" for extra_index_url in pip_index_urls[1:])
+    options_str = " ".join(options)
```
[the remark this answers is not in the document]
andreyvelich: Good catch, tho.
@kubeflow/kubeflow-sdk-team @kubeflow/wg-pipeline-leads Do we know if we have the same CVE in KFP Client upstream?

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -278,18 +279,29 @@ def get_script_for_python_packages(
...
+    script_for_python_packages = header_script + textwrap.dedent(
+        f"""
+        PACKAGES="{packages_str}"
+        PIP_OPTS="{options_str}"
+        LOG_FILE=/tmp/pip_install.log
+        rm -f "$LOG_FILE"
+
+        if PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+            --no-warn-script-location $PIP_OPTS --user $PACKAGES >"$LOG_FILE" 2>&1; then
+            echo "Successfully installed Python packages: $PACKAGES"
```
andreyvelich, line 301: Maybe we should say?

```suggestion
            echo "Successfully installed the user' Python packages: $PACKAGES"
```

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -278,18 +279,29 @@ def get_script_for_python_packages(
...
+        elif PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+            --no-warn-script-location $PIP_OPTS $PACKAGES >>"$LOG_FILE" 2>&1; then
+            echo "Successfully installed Python packages: $PACKAGES"
```
andreyvelich, line 304:
```suggestion
            echo "Successfully installed the system's Python packages: $PACKAGES"
```

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -278,18 +279,29 @@ def get_script_for_python_packages(
...
-            packages_str,
-        )
+    # First try per-user installation, then fall back to system-wide installation.
+    # Pip output is captured to a log file and only printed when both attempts fail;
+    # on success we emit a single concise confirmation line.
+    script_for_python_packages = header_script + textwrap.dedent(
+        f"""
+        PACKAGES="{packages_str}"
+        PIP_OPTS="{options_str}"
+        LOG_FILE=/tmp/pip_install.log
```
andreyvelich: What if /tmp directory doesn't exist, shall we just use ?
Do we know if KFP client write pip logs to a file?
cc @kubeflow/wg-pipeline-leads

briangallagher: @andreyvelich Your comment cut off there but I think you might be suggesting to use the current working directory? I've update to that now, so it writes this file similar to how the training file is wrote.

kubeflow/trainer/backends/kubernetes/utils.py
briangallagher: done

kubeflow/trainer/backends/kubernetes/utils.py
briangallagher: done

kubeflow/trainer/backends/kubernetes/utils.py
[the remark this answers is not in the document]
briangallagher: Added shlex to prevent shell injection

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -278,18 +279,29 @@ def get_script_for_python_packages(
...
-            " ".join(options),
-            packages_str,
-        )
+    # First try per-user installation, then fall back to system-wide installation.
+    # Pip output is captured to a log file and only printed when both attempts fail;
+    # on success we emit a single concise confirmation line.
+    script_for_python_packages = header_script + textwrap.dedent(
+        f"""
+        PACKAGES="{packages_str}"
+        PIP_OPTS="{options_str}"
```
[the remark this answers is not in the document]
briangallagher: Added shlex to prevent shell injection

kubeflow/trainer/backends/kubernetes/utils.py
andreyvelich: Yeah, sorry, let's use the current directory, in-case `/tmp` doesn't exist.

astefanutti (commented):
Thanks @briangallagher!

/lgtm
/approve

PR #280, opened 2026-02-10
feat: Adds a GitHub Actions workflow to check kubeflow/hub/OWNERS.

### LINKED ISSUE #275: Add OWNERS validation for kubeflow/hub
approvers for kubeflow hub should also be in `kubeflow-hub-team` in `kubeflow/internal-acls` 

See https://github.com/kubeflow/model-registry/pull/2182#pullrequestreview-3774598224

We have a checker in `model-registry` for this: https://github.com/kubeflow/model-registry/blob/main/.github/workflows/check-owners.yml

A good first step would be adding that checker to this repo and adjusting it to look at `kubeflow/hub/OWNERS`. 

Perhaps it should also warn if out of sync with the base `OWNERS` in model-registry?

muhammadjunaid8047:
This PR ensures that approvers for Kubeflow Hub are properly synchronized with the kubeflow-hub-team in kubeflow/internal-acls 
Fixes #275

### DIFF
--- a/.github/workflows/check-owners.yaml
+++ b/.github/workflows/check-owners.yaml
@@ -3,7 +3,7 @@ name: "Check OWNERS"
 on:
   pull_request:
     paths:
-      - 'OWNERS'
+      - 'kubeflow/hub/OWNERS'
   workflow_dispatch:
 
 permissions:


muhammadjunaid8047:
Hi @jonburdo,
Kindly review this PR.

.github/workflows/check-owners.yaml
```diff
@@ -0,0 +1,77 @@
+name: "Check OWNERS"
+
+on:
+  pull_request:
+    paths:
+      - 'OWNERS'
```
jonburdo:
```suggestion
      - 'kubeflow/hub/OWNERS'
```

.github/workflows/check-owners.yaml
```diff
@@ -0,0 +1,77 @@
...
+            echo "::warning file=kubeflow/hub/OWNERS,line=1::Approvers not in kubeflow-hub-team: $missing"
+          else
+            echo "✅ All approvers are in kubeflow-hub-team!"
+          fi
+
+          # -------------------------------------------------------------------
+          # Check if out of sync with model-registry/OWNERS (Base OWNERS)
+          # -------------------------------------------------------------------
+
+          echo "Checking sync with model-registry OWNERS..."
```
jonburdo, line 55: This added block makes sense to me (I did suggest it in the issue), but I would like to get a second opinion.

I also wonder if we should keep it simple and:
- make the OWNERS files identical (currently they are slightly different I believe)
- simplify this check to simply require that the file content is identical (avoids the sorting and comparing)

This could maybe look something like:
```bash
local_owners_path="kubeflow/hub/OWNERS"
remote_owners_url="https://raw.githubusercontent.com/kubeflow/model-registry/main/OWNERS"

if ! diff -u "$local_owners_path" <(curl -s "$remote_owners_url"); then
    echo "❌ Error: OWNERS files out of sync for kubeflow hub!"
    exit 1
else
    echo "✅ OWNERS files in sync for kubeflow hub!"
fi
```

jonburdo (commented):
Thanks @muhammadjunaid8047 for working on this!

I see that this is the same as the model-registry workflow but with;
- `OWNERS` replaced with `kubeflow/hub/OWNERS`
- an extra block to compare this kubeflow/hub/OWNERS with model-registry OWNERS

Seems like the right approach to me. Just a couple minor comments

For other reviewers, here is the current diff between model-registry's workflow file and this one: [check-owners.yaml diff](https://github.com/kubeflow/sdk/compare/2598943e4adc08e9b6e27773a44bcb14fdb4de9a..3140543e2788e37406e481789838de3fc5f7ee8c)

.github/workflows/check-owners.yaml
muhammadjunaid8047: Yes, this approach makes more sense to me.

PR #282, opened 2026-02-11
chore!: upgrade to Python 3.10

jonburdo:
This upgrades the minimum Python version for the project from 3.9 to 3.10. Python 3.9 is past end-of-life and dependencies will likely require a supported version soon.

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections.abc import Callable
 from dataclasses import fields
 import inspect
 import os
 import textwrap
-from typing import Any, Callable, Optional, Union
+from typing import Any, Optional, Union
 from urllib.parse import urlparse
 
 from kubeflow_trainer_api import models
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "kubeflow"
 dynamic = ["version"]
-requires-python = ">=3.9"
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 name = "kubeflow"
 dynamic = ["version"]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
```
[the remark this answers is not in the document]
kramaranya: +1

szaher (approved):
/lgtm

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -12,11 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections.abc import Callable
 from dataclasses import fields
 import inspect
 import os
 import textwrap
-from typing import Any, Callable, Optional, Union
+from typing import Any, Optional, Union
```
andreyvelich, line 20: As part of migration to Python 3.10, do we want to replace Union with `|` to be consistent with MR client: https://github.com/jonburdo/sdk/blob/5f164624652227a33bdb83638e17a7536b934382/kubeflow/hub/api/model_registry_client.py#L100?

jonburdo: Yes, I think so, and `Optional[x]` to `x | None`. I can do this, but maybe in a quick follow-up just to keep this diff simpler?

andreyvelich: Sure, sounds good.

PR #288, opened 2026-02-11
chore: upgrade code style for python3.10

jonburdo:

Primarily updates code style after upgrading the min Python version to 3.10 to change;
- `Union[T, U]` &rarr; `T | U`
- `Optional[T]` &rarr; `T | None`

as suggested here: https://github.com/kubeflow/sdk/pull/282#discussion_r2794067017

Also includes a second commit to handle such union types in `convert_value` utility with a test.

**Checklist:**

### DIFF
--- a/kubeflow/optimizer/backends/kubernetes/utils.py
+++ b/kubeflow/optimizer/backends/kubernetes/utils.py
@@ -13,7 +13,8 @@
 # limitations under the License.
 
 from dataclasses import fields
-from typing import Any, Optional, get_args, get_origin
+from types import NoneType, UnionType
+from typing import Any, Union, get_args, get_origin
 
 from kubeflow_katib_api import models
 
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -123,8 +123,6 @@ select = [
 
 ignore = [
   "B006",  # mutable-argument-default
-  "UP007", # Use X | Y instead of Union[X, Y] (requires Python 3.10+)
-  "UP045", # Use X | None instead of Optional[X] (requires Python 3.10+)
```
jonburdo, line 134: For the first commit I just removed these and ran:
```
uv run ruff check --fix
uv run ruff format
```

kubeflow/optimizer/backends/kubernetes/utils.py
```diff
@@ -32,11 +33,24 @@
...
+    """
     origin = get_origin(target_type)
-    args = get_args(target_type)
 
-    if origin is Optional:
-        target_type = args[0]
+    # `T | None` produces UnionType instead of Union before Python 3.14.
+    # `Optional[T]` always produces Union
+    if origin is Union or origin is UnionType:
+        args = get_args(target_type)
```
jonburdo, line 50: `get_origin` isn't perfectly consistent across python versions
```
$ python3.13 - <<'EOF'
from typing import get_origin, Optional
print(get_origin(int | None))
print(get_origin(Optional[int]))
EOF

<class 'types.UnionType'>
typing.Union


$ python3.14 - <<'EOF'
from typing import get_origin, Optional
print(get_origin(int | None))
print(get_origin(Optional[int]))
EOF

<class 'typing.Union'>
<class 'typing.Union'>
```

PR #289, opened 2026-02-11
chore: bump minimum model-registry version to 0.3.6

jonburdo:
Bump minimum `model-registry` version under `hub` extra to `0.3.6`.

**Checklist:**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,7 @@ podman = [
   "podman>=5.6.0"
 ]
 hub = [
-  "model-registry>=0.3.0",
+  "model-registry>=0.3.6",
 ]
 
 [dependency-groups]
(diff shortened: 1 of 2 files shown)


jonburdo:
/cc @tarilabs 

Are we okay with doing this?

kramaranya:
> @kramaranya not sure about the README.md as that signifies which Model Registry stream this client is compatible with (i.e. 0.3)
> 
> https://github.com/kubeflow/sdk/blob/36e22827955d636610da0124d63e06e5505786e6/README.md?plain=1#L189
> 
> 🤔 or have I looked in the wrong place?

Ah yeah, you're right!

Thanks @jonburdo @tarilabs!

/lgtm
/approve
/hold feel free to unhold when ready

PR #307, opened 2026-02-18
fix: Fix runtime lookup fallback and test local SDK in E2E

XploY04:

Fixes #

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -108,38 +108,30 @@ class KubernetesBackend(RuntimeBackend):
             async_req=True,
         )
 
-        # Helper to fetch and convert runtime list
-        def fetch_runtime_list(thread, kind, model_class):
-            try:
-                return model_class.from_dict(thread.get(common_constants.DEFAULT_TIMEOUT)), None
-            except multiprocessing.TimeoutError as e:
-                return None, TimeoutError(f"Timeout to list {kind}s")
-            except Exception as e:
-                return None, RuntimeError(f"Failed to list {kind}s")
-
-        cluster_runtimes, cluster_error = fetch_runtime_list(
-            cluster_thread,
-            constants.CLUSTER_TRAINING_RUNTIME_KIND,
-            models.TrainerV1alpha1ClusterTrainingRuntimeList,
-        )
-        namespace_runtimes, namespace_error = fetch_runtime_list(
-            namespace_thread,
-            constants.TRAINING_RUNTIME_KIND,
-            models.TrainerV1alpha1TrainingRuntimeList,
-        )
-
-        if namespace_error and cluster_error:
-            if isinstance(namespace_error, TimeoutError) and isinstance(cluster_error, TimeoutError):
-                raise namespace_error
-            raise RuntimeError(
-                f"Failed to list runtimes. Namespace error: {namespace_error}. "
-                f"Cluster error: {cluster_error}"
+        # Fetch namespace-scoped TrainingRuntimes.
+        namespace_runtimes = None
+        try:
+            namespace_runtimes = models.TrainerV1alpha1TrainingRuntimeList.from_dict(
+                namespace_thread.get(common_constants.DEFAULT_TIMEOUT)
             )
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(f"Timeout to list {constants.TRAINING_RUNTIME_KIND}s") from e
+        except client.ApiException as e:
+            if e.status != 404:
+                raise RuntimeError(f"Failed to list {constants.TRAINING_RUNTIME_KIND}s") from e
+        except Exception as e:
+            raise RuntimeError(f"Failed to list {constants.TRAINING_RUNTIME_KIND}s") from e
 
-        if namespace_error:
-            logger.warning(f"Failed to list namespaced runtimes, skipping: {namespace_error}")
-        if cluster_error:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-e2e.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           pip install papermill==2.6.0 jupyter==1.1.1 ipykernel==6.29.5
 
           echo "Install Kubeflow SDK with Docker support for local Notebooks"
```
jaiakash: nit: add comment that sdk is being installed from the checkout code only. Somethings like 
"Installing Kubeflow SDK from source, with Docker support "

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -186,11 +187,8 @@ def get_runtime(self, name: str) -> types.Runtime:
             raise TimeoutError(
                 f"Timeout to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
             ) from e
-        except Exception as e:
-            raise RuntimeError(
-                f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
-            ) from e
-
+        except Exception:
```
jaiakash: We do need error/warning for this, right!
Why `pass` ?

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -203,14 +201,13 @@ def get_runtime(self, name: str) -> types.Runtime:
...
-
         except multiprocessing.TimeoutError as e:
             raise TimeoutError(
-                f"Timeout to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+                f"Timeout to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {name}"
             ) from e
         except Exception as e:
             raise RuntimeError(
-                f"Failed to get {constants.CLUSTER_TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+                f"Failed to get Runtime: {name} (checked both namespaced and cluster scope)"
```
jaiakash, line 222: Also, can you review and check if we can update the tests at https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend_test.py

kubeflow/trainer/backends/kubernetes/backend.py
XploY04: Yes, I initially thought that it should directly fall to the cluster_runtimes, but it can have other errors as well, added that now.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -186,9 +199,14 @@ def get_runtime(self, name: str) -> types.Runtime:
...
             ) from e
+        except client.ApiException as e:
+            if e.status != 404:
+                raise RuntimeError(
+                    f"Failed to get namespaced {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+                ) from e
         except Exception as e:
             raise RuntimeError(
-                f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
+                f"Failed to get namespaced {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
```
andreyvelich: you don't need to put namespaced here since it is already in the Kind.

```suggestion
                    f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
                ) from e
except Exception as e:
raise RuntimeError(
                f"Failed to get {constants.TRAINING_RUNTIME_KIND}: {self.namespace}/{name}"
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -111,23 +111,36 @@ def list_runtimes(self) -> list[types.Runtime]:
...
         def fetch_runtime_list(thread, kind, model_class):
             try:
-                return model_class.from_dict(thread.get(common_constants.DEFAULT_TIMEOUT))
+                return model_class.from_dict(thread.get(common_constants.DEFAULT_TIMEOUT)), None
             except multiprocessing.TimeoutError as e:
-                raise TimeoutError(f"Timeout to list {kind}s") from e
+                return None, TimeoutError(f"Timeout to list {kind}s")
             except Exception as e:
-                raise RuntimeError(f"Failed to list {kind}s") from e
+                return None, RuntimeError(f"Failed to list {kind}s")
```
andreyvelich: Since we need to process CTR and TR differently, it might be easier to remove this function, and just do similar things we run for `get_runtime()`.
E.g. run `list_namespaced_custom_object` twice with the different Exception checks (for TrainingRuntime you should check `e.status != 404`

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -111,23 +111,36 @@ def list_runtimes(self) -> list[types.Runtime]:
...
+                raise namespace_error
+            raise RuntimeError(
+                f"Failed to list runtimes. Namespace error: {namespace_error}. "
+                f"Cluster error: {cluster_error}"
+            )
+
+        if namespace_error:
+            logger.warning(f"Failed to list namespaced runtimes, skipping: {namespace_error}")
+        if cluster_error:
+            logger.warning(f"Failed to list cluster runtimes, skipping: {cluster_error}")
```
andreyvelich: No need this for now.

```suggestion
```

Issue #125, opened 2025-10-27
Integrate the Kubeflow Pipelines with Kubeflow SDK


andreyvelich:
### What you would like to be added?

[The Kubeflow SDK goal](https://docs.google.com/document/d/1rX7ELAHRb_lvh0Y7BK1HBYAbA0zi9enB0F_358ZC58w/edit?tab=t.0#heading=h.e0573r7wwkgl) is to provide unified interface to interact with Kubeflow tools, and we would like to create `PipelinesClient()` to give users ability to create ML pipelines, as we discussed at [the last Pipelines WG](https://docs.google.com/document/d/1cHAdK1FoGEbuQ-Rl6adBDL5W2YpDiUbnMLIwmoXBoAU/edit?tab=t.0).

Given that migration of `kfp` into `kubeflow` SDK might take time, @mprahl and @HumairAK proposed that we install `kfp` in the Kubeflow SDK packages, and create create a thin wrapper on top of the [KFP Client](https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/client/client.py#L101):

```python
class PipelinesClient(kfp.Client):
  ....
```

Additionally, we still want to install Kubeflow SDK as part of `kfp` if users define it in their components: https://github.com/kubeflow/pipelines/issues/12027. @Prateekbala has already opened PR here: https://github.com/kubeflow/pipelines/pull/12368

I believe, that might help us to drive the integration forward and achieve the long-term goal of unified Python SDK to manage AI workloads on Kubernetes.

cc @kubeflow/wg-pipeline-leads @kubeflow/kubeflow-sdk-team @franciscojavierarceo @kubeflow/kubeflow-trainer-team 

### Why is this needed?

Kubeflow SDK should support Kubeflow Pipelines.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

andreyvelich:
We haven't started work on this yet.
@Anjankumar-Developer Are you still interested to move this forward?

kramaranya:
@MStokluska will work on the integration from our side!
/unassign @Anjankumar-Developer @kramaranya

andreyvelich:
This is still open.
KEP is merged here: https://github.com/kubeflow/sdk/pull/343
/reopen

andreyvelich:
We can close this since `PipelinesClient()` is now part of Kubeflow SDK 🎉 
/close

PR #308, opened 2026-02-19
chore: Trainer: Specialized Trainers

szaher:

Fixes #

**Checklist:**


docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+   interface and a suite of framework-specific implementations (`TorchTrainer`, `MPITrainer`,
+   `JAXTrainer`, etc.) that automatically discover and validate the correct
+   `ClusterTrainingRuntime` using the `trainer.kubeflow.org/framework` label. This fills the
+   "missing middle" between the overly generic `CustomTrainer` and the overly narrow
+   `BuiltinTrainer`.
+
+2. **`RuntimeConfig` dataclass** — A dedicated configuration object that cleanly separates
+   per-job runtime environment settings (packages, pip config, environment variables) from
+   training logic and scaling parameters. This replaces the current pattern where
+   `CustomTrainer` conflates runtime concerns with trainer concerns.
```
kramaranya, line 94: Would this require runtime/controller changes?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+   `CustomTrainerContainer`, `BuiltinTrainer`, and `TrainerClient.train()` APIs.
+
+### Non-Goals
+
+1. **Controller/CRD changes.** This proposal is SDK-only. No changes to the Kubeflow
+   Trainer controller, `TrainJob` CRD, or `ClusterTrainingRuntime` CRD are required.
+2. **New runtime labels or conventions.** We rely on the existing
+   `trainer.kubeflow.org/framework` label already required on all runtimes.
+3. **Deprecating `CustomTrainer` or `BuiltinTrainer`.** Both remain supported.
+   Specialized trainers are an additional option, not a replacement.
```
kramaranya: Is the plan to eventually deprecate those or do we want to always maintain both options?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+            ValueError: If the runtime's framework is not in supported_frameworks.
+        """
+        # Runtime.trainer.framework holds the value of
+        # the trainer.kubeflow.org/framework label.
+        if runtime.trainer.framework not in self.supported_frameworks:
+            raise ValueError(
+                f"{type(self).__name__} supports frameworks "
+                f"{self.supported_frameworks}, but runtime '{runtime.name}' "
+                f"has framework '{runtime.trainer.framework}'"
+            )
```
kramaranya, line 344: We also would need to validate `runtime.trainer.trainer_type` too

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    def get_train_func_args(self) -> Optional[dict]:
+        return self.func_args
+
+    def get_framework_args(self) -> dict:
+        args = {}
+        if self.max_restarts is not None:
+            args["max-restarts"] = str(self.max_restarts)
+        if self.monitor_interval is not None:
+            args["monitor-interval"] = str(self.monitor_interval)
+        return args
```
kramaranya, line 441: where these new args go in the TrainJob spec?

kramaranya (commented):
Thanks @szaher!
Looks great to me, and it should be a great improvement to the user experience in Kubeflow SDK!

/assign @andreyvelich @astefanutti @briangallagher @Fiona-Waters @MStokluska

MStokluska:
Thanks @szaher - it looks really good to me!

astefanutti (commented):
Thanks @szaher!

/lgtm

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+-->
+
+|                |                                                              |
+| -------------- | ------------------------------------------------------------ |
+| **Authors**    | @szaher                                                      |
+| **Status**     | Draft                                                        |
+| **Created**    | 2026-02-11                                                   |
+| **Reviewers**  |                                                              |
+| **Supersedes** | N/A                                                          |
+| **Relevant Issues** | https://github.com/kubeflow/sdk/issues/285              |
```
andreyvelich, line 15: Instead of this, can we consider to add kep.yaml in a format that we use for k8s. Check: https://github.com/kubernetes-sigs/jobset/blob/main/keps/463-ElasticJobsets/kep.yaml

Also, you can add implementation history as here: https://github.com/kubeflow/trainer/blob/master/docs/proposals/2170-kubeflow-trainer-v2/README.md#implementation-history

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+
+---
+
+## Overview
+
+This proposal introduces two backward-compatible enhancements to the Kubeflow SDK
+(`kubeflow/sdk`) trainer subsystem:
+
+1. **Specialized, framework-aware trainer abstractions** — A new `BaseTrainer` abstract
+   interface and a suite of framework-specific implementations (`TorchTrainer`, `MPITrainer`,
```
andreyvelich: Why MPI and Torch are in the same category?
MPI is just a technology we use for distributed workloads.
As you can see, we create dedicated runtime like DeepSpeed Distributed and MLX Distributed which leverage MPI: https://github.com/kubeflow/trainer/tree/master/manifests/base/runtimes

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+   `JAXTrainer`, etc.) that automatically discover and validate the correct
+   `ClusterTrainingRuntime` using the `trainer.kubeflow.org/framework` label. This fills the
+   "missing middle" between the overly generic `CustomTrainer` and the overly narrow
+   `BuiltinTrainer`.
```
andreyvelich: Do we require to do any changes to BuiltinTrainer after this? IIRC, we said that this will be the foundation for Builtins as well.
cc @Sapthagiri777 @Electronic-Waste @khushiiagrawal

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+  (packages, pip URLs, env vars) into a single flat dataclass. It provides no
+  framework-specific validation or argument handling.
+
+- **`BuiltinTrainer`** is restricted to a single use case (`TorchTuneConfig`) and does
+  not accept user-defined training functions.
+
+For the majority of distributed training workloads — "run this PyTorch DDP function on
+N nodes" or "run this MPI script across a cluster" — neither abstraction fits well.
+Users must either use the low-level `CustomTrainer` with manual runtime wiring, or
+fall back to raw YAML.
```
andreyvelich, line 122: Would love to hear feedback from @vsoch on this proposal.
Do you know if we have any interest from HPC users to leverage our Kubeflow Python SDK for TrainJob submission to manage HPC tasks on k8s?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+| **MLOps Engineer** | Must help data scientists find the correct runtime name for their framework | Framework validation catches mismatches at submission time |
+| **Platform Admin / DevOps** | Cannot enforce that users pick the correct runtime for their framework | Trainers validate `trainer.kubeflow.org/framework` labels on runtimes |
+
+### Goals
+
+1. Define a `BaseTrainer` abstract interface that all trainer implementations satisfy,
+   enabling the SDK and backends to handle any trainer polymorphically.
+2. Implement Tier 1 framework-specific trainers (`TorchTrainer`, `MPITrainer`,
+   `JAXTrainer`, `XGBoostTrainer`) that auto-discover runtimes by the
+   `trainer.kubeflow.org/framework` label and validate runtime compatibility.
```
andreyvelich: Who will define which frameworks we will support?
For example, we also have DeepSpeed and MLX framework: https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/deepspeed_distributed.yaml#L6

@astefanutti @tenzen-y Any thoughts on exposing `framework` in the Runtime API directly, or it is fine to use labels for now?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+3. Provide a clear extension point (Tier 2) for community-contributed, application-level
+   trainers (e.g., `TransformersTrainer`, `DeepSpeedTrainer`).
```
andreyvelich: Do we expect to create dedicated ClusterTrainingRuntime for Trainers we support upstream?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+
+- Mixes runtime-environment settings (`packages_to_install`, `pip_index_urls`, `env`)
+  with scaling/resource settings (`num_nodes`, `resources_per_node`) and training logic
+  (`func`, `func_args`).
+- No framework awareness. A user can pass a PyTorch training function with an MPI
+  runtime and the SDK will not catch the mismatch until the controller rejects the job
+  or, worse, it fails at execution time.
+- `func_args` is an untyped `dict` that conflates user hyperparameters with framework
+  arguments (e.g., `rdzv_endpoint`, `nnodes`) that the Trainer controller already
+  injects via environment variables.
```
andreyvelich, line 221: What do you mean by that? These args just pass to the training function, not container args

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+        func_args: Arguments passed to the training function. Should contain
+            only user hyperparameters — framework arguments like rdzv_endpoint
+            and nnodes are injected by the Kubeflow Trainer controller.
+        num_nodes: Number of nodes for distributed training.
+        resources_per_node: Resource requirements per node (cpu, memory, gpu).
+        image: Optional custom container image.
+        max_restarts: Maximum number of worker group restarts before failing.
+            Maps to torchrun --max-restarts.
+        monitor_interval: Interval in seconds for the elastic agent to monitor
+            workers. Maps to torchrun --monitor-interval.
```
andreyvelich, line 426: How would this work with TrainJob API? Some of those `torchrun` capabilities needs to be supported in the orchestration layer first.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    """
+
+    supported_frameworks: ClassVar[list[str]] = ["torch"]
```
andreyvelich: I think, we should have definition what is Framework.
Do we expect that users can run DeepSpeed jobs with TorchTrainer?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    """Trainer for HuggingFace Transformers training.
+
+    Wraps HuggingFace's Trainer API and maps to a PyTorch runtime.
+    """
+
+    supported_frameworks: ClassVar[list[str]] = ["torch"]
+
+    model_name: str
+    training_args: dict
+    dataset: str
```
andreyvelich: Why HuggingFace Trainer has dataset and model_name?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+
+    index_urls: list[str] = field(
+        default_factory=lambda: list(constants.DEFAULT_PIP_INDEX_URLS)
+    )
+    quiet: bool = True
+    install_as_user: bool = False
+
+
+@dataclass
+class RuntimeConfig:
```
andreyvelich, line 672: Would the name be conflicting with our TrainingRuntime and ClusterTrainingRuntime, or it should be fine?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    """Per-job runtime environment configuration.
+
+    Separates runtime-environment concerns (what packages to install, what
+    environment variables to set) from training-loop and scaling concerns.
+
+    This is passed to `TrainerClient.train()` and applies regardless of
+    the trainer type used.
```
andreyvelich: Shall we introduce this config to CustomTrainer as well?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+
+    packages: Optional[list[str]] = None
+    pip_config: Optional[PipConfig] = None
+    env: Optional[dict[str, str]] = None
+```
+
+**Design decisions:**
+
+- Uses `@dataclass` (not Pydantic `BaseModel`) to be consistent with the rest of the
+  SDK codebase.
```
andreyvelich, line 704: IIRC, we discussed before to use dataclass only for internal types, and use Pydantic BaseModel for user-facing APIs.
Do we still want to do that change @kubeflow/kubeflow-sdk-team @Fiona-Waters @tariq-hasan @Shekharrajak ?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+- `PipConfig` is a separate dataclass rather than inline fields, because pip
+  configuration is a distinct concern with its own options.
+- `packages` replaces `packages_to_install` for brevity.
```
andreyvelich: That will make us in-consistent with KFP.
Do we see any issues since @MStokluska is working on PipelinesClient?

https://github.com/kubeflow/sdk/issues/125
cc @kubeflow/wg-pipeline-leads

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+        initializer: Optional["Initializer"] = None,
+        trainer: Optional[
+            Union[
+                "CustomTrainer",
+                "CustomTrainerContainer",
+                "BuiltinTrainer",
+                "BaseTrainer",        # NEW: accepts any specialized trainer
+            ]
+        ] = None,
+        runtime_config: Optional["RuntimeConfig"] = None,  # NEW
```
andreyvelich, line 744: Shall the runtime_config be part of Trainers?
I would imagine a use-case when users want to set custom env for the initializers as well.
cc @akshaychitneni

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+- `validate_runtime()` has a default implementation so subclasses get validation for
+  free but can extend it.
+- The interface uses simple return types (`Optional[Callable]`, `dict`, `Optional[int]`)
+  rather than framework-specific types, keeping the base class framework-agnostic.
+- Methods use `get_*` naming to clearly indicate they are accessors, not setters.
+
+### B. Tier 1: Framework-Specific Trainers
+
+Each Tier 1 trainer maps 1:1 to a framework identified by the
+`trainer.kubeflow.org/framework` label value.
```
andreyvelich: What happen if users deploy two ClusterTrainingRuntime with the same framework label?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+### Runtime Validation
+
+Validation happens at two levels:
+
+1. **Framework label check** (in `BaseTrainer.validate_runtime()`): Ensures the
+   runtime's `trainer.kubeflow.org/framework` label value is in the trainer's
+   `supported_frameworks` list.
+
+2. **Framework-specific checks** (in subclass overrides): For example, `MPITrainer`
+   could verify that the runtime's MPI policy source is configured correctly.
```
andreyvelich: I think we should talk more on the validation. Usually, we offload this to the control plane (e.g. webhook).

If that is only framework compatibility, we can directly fetch the desired runtime using the label.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+### B. Tier 1: Framework-Specific Trainers
+
+Each Tier 1 trainer maps 1:1 to a framework identified by the
+`trainer.kubeflow.org/framework` label value.
+
+#### TorchTrainer
+
+```python
+@dataclass
+class TorchTrainer(BaseTrainer):
```
andreyvelich: @kubeflow/kubeflow-sdk-team @Fiona-Waters @abhijeet-dhumal @akshaychitneni Did we ever explore what capabilities Ray Train implements in [their trainers, e.g. TorchTrainer.](https://github.com/ray-project/ray/blob/master/python/ray/train/v2/torch/torch_trainer.py#L19)
I know that they do some changes to the dataset and model to attach it to the Distributor.

---
Additionally (as a future extension), it would be interesting to validate users code in a way that it is correctly configured for distributed training with Torch or any other framework. For example, we can use AI Agents to analyze users' code before submission and suggest changes/enhancements, since we have a lot of context around distributed configuration on k8s.

cc @VassilisVassiliadis @bigsur0

andreyvelich (commented):
Thanks @szaher! I left a few comments, sorry for the delay.

docs/proposals/285-specialized-trainers/README.md
astefanutti: The supported frameworks are defined on the trainer side by the training runtimes. Also my understand is we want it to be extensible so users can bring their own runtimes / frameworks / trainers.

I think using the `trainer.kubeflow.org/framework` is OK. I remember we discussed whether it could be in the TrainingRuntime API when we introduced it: https://github.com/kubeflow/sdk/pull/31#issuecomment-3117456148

Now that we are aligning the SDK to rely on that information, it may warrant promoting it to a proper spec field. 

Accessorily, field indexing is now supported for custom resources: https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/#custom-resources-fields

docs/proposals/285-specialized-trainers/README.md
astefanutti: That case is covered in a latter section specifying the SDK fails fast and the user has to pass the name explicitely.

docs/proposals/285-specialized-trainers/README.md
szaher: no, it should required any backend changes but it should align closely with it.

docs/proposals/285-specialized-trainers/README.md
szaher: this is `non-goal` not a goal of the kep

docs/proposals/285-specialized-trainers/README.md
szaher: is that supported by a backend label or annotation?

docs/proposals/285-specialized-trainers/README.md
szaher: those aren't backend args it's framework args. it gets passed via the entrypoint to the script that runs in the pods

docs/proposals/285-specialized-trainers/README.md
szaher: That is similar to `TorchTrainer` it's a sort of custom trainer where we pass your code as to the dedicate runtime as is.

docs/proposals/285-specialized-trainers/README.md
krishdef7: This connects directly to the TRL/RLHF integration discussed in #2839, a TRLTrainer (or more generally, a post-training specialized trainer) would sit in Tier 2 here, built on top of BaseTrainer. The BuiltinTrainer today has the hardcoded isinstance(trainer.config, TorchTuneConfig) check, a dynamic registry pattern (as proposed in #2839) would let TRL and other frameworks plug in without modifying that dispatch logic. Happy to draft what that integration point would look like if useful.

docs/proposals/285-specialized-trainers/README.md
szaher: the proposal here groups all pip configurations under a specific class which passed to the train function. the pipelines is using decorator to capture that so in their case it might be easier to keep all parameters flat for ease of user as I think most of the `dsl.component` might have a safe default value.


```python

@dsl.component(base_image="quay.io/my-org/training-image:latest", 
packages_to_install=[packages for customizing pipelines task runtime])
def submit_training(model_path: str, num_epochs: int = 3, nproc_per_node: int = 1):
    client = TrainerClient()
    client.train(..., pip_config=PipConfig(extra packages, enable verbose pip, extra index), 
    runtime_config=RuntimeConfig(extra packages that go on the training runtime.))
    client.wait_for_job_status("train-job")

```

docs/proposals/285-specialized-trainers/README.md
szaher: we can move it to trainers if needed. do we need to also, move the pip config?

docs/proposals/285-specialized-trainers/README.md
szaher: sure, I think we need to validate job before submitting too and backend can validate everything in tact before running the job. SDK validation could just raise warnings.

docs/proposals/285-specialized-trainers/README.md
szaher: ray trainers have different philosophy, The trainer gets everything (runtime config, scale config, job config, job script(s), worker specific config) then calls trainer.fit() so there is no client initialization similar to kubeflow (we need that as we don't have a pre-deployed cluster)
we can discuss it if we want

docs/proposals/285-specialized-trainers/README.md
andreyvelich: I am fine to move these fields under `RuntimeConfig`, but I suggest to keep parameter name as `packages_to_install`.

docs/proposals/285-specialized-trainers/README.md
andreyvelich: But we don't support MPI Runtime upstream, we only deploy DeepSpeedRuntime and MLXRuntime which are MPI-based, so I am not sure if we should have dedicated `MPITrainer`

docs/proposals/285-specialized-trainers/README.md
andreyvelich: But users can still use TorchTrainer if they provide the runtime name, right?
We just validate that Runtime has the correct framework label?

docs/proposals/285-specialized-trainers/README.md
andreyvelich: That sounds great! We can later refactor this to a dedicated Runtime API field if that is needed.

>Also my understand is we want it to be extensible so users can bring their own runtimes / frameworks / trainers.

Yeah, that makes sense, and users can bring their own Trainers via Tier 1 and Tier 2 extensions as @szaher mentioned in the KEP.

docs/proposals/285-specialized-trainers/README.md
andreyvelich: Do we need to have separate PipConfig type under RuntimeConfig? 

Another feature request I got from users that sometimes they want to install custom tools on top of base image (e.g. using `apt install <package_name>`).
We don't support it today, and users require to re-build Docker image manually.

docs/proposals/285-specialized-trainers/README.md
szaher: done

docs/proposals/285-specialized-trainers/README.md
szaher: Can we have DeepSpeed Runtime based on torch plugin and another runtime based on MPI?

docs/proposals/285-specialized-trainers/README.md
szaher: I think we don't need the builtin Trainer but I am just keeping it for backward compatibility unless we don't care I can update the kep to remove it.

docs/proposals/285-specialized-trainers/README.md
szaher: sure, we can keep it as it is.

docs/proposals/285-specialized-trainers/README.md
szaher: I think this one might need a separate KEP. We can use something like [olot](https://pypi.org/project/olot/) 
@astefanutti WDYT?

docs/proposals/285-specialized-trainers/README.md
tariq-hasan: I suppose the question may be how we handle config-driven trainers for post-training LLM fine-tuning that currently fall under the scope of `BuiltInTrainers`.

If the `BaseTrainer` hierarchy is purely for function-based trainers then do we segregate config-driven trainers entirely outside of the `BaseTrainer` scope?

Or if the `BaseTrainer` hierarchy is meant to be a catch-all for everything then do we classify these "dynamic LLM trainers" as Tier-2 trainers, and if so, how do we differentiate them from function-based Tier 2 trainers such as the proposed `TransformersTrainer`?

I am assuming either way we may need to retain an interface for config-driven trainers. But the appropriate placement for them may help to determine the scope of https://github.com/kubeflow/trainer/issues/2839 within the broader scope for this KEP.

docs/proposals/285-specialized-trainers/README.md
astefanutti: Yes it would probably make sense workspace snapshotting #48 be part of the dynamic runtime configuration.

Would dynamic runtime configuration make sense for "built-in" / tier-2 trainers? Would we be able to guarantee their runtimes can be configured dynamically?

If not, `runtime_config` might have to be part of Trainers conceptually, and maybe Initializers eventually.

docs/proposals/285-specialized-trainers/README.md
krishdef7: Many post-training frameworks (TRL, Unsloth, Axolotl, etc.) are effectively **config-driven trainers**, where the training entrypoint is a framework trainer object (e.g. `SFTTrainer`) rather than a user-defined `train()` function.

While experimenting with a small prototype around the backend registry idea from #2839, one pattern that seemed to work well was resolving the backend based on the **trainer config type**, e.g.:

```python
@register_backend(TRLConfig)
class TRLBackend(RuntimeBackend):
    ...
```

This keeps the core `BaseTrainer` abstraction simple while allowing config-driven frameworks to plug in dynamically without expanding SDK dispatch logic.

Conceptually it would allow both styles to coexist:

* function-based trainers (`TorchTrainer(train_fn=...)`)
* config-driven trainers (`TorchTrainer(config=TRLConfig(...))`)

with the backend registry selecting the appropriate runtime adapter.

Curious if this aligns with the intended direction for the dynamic trainer framework.

docs/proposals/285-specialized-trainers/README.md
andreyvelich: DeepSpeed can be bootstrapped via `mpirun` or `torchrun`: https://www.deepspeed.ai/getting-started/
It depends how users want to configure it, but today we serve only MPI-based runtime for DeepSpeed in Trainer: https://github.com/kubeflow/trainer/blob/master/manifests/base/runtimes/deepspeed_distributed.yaml

Shall we start with dedicated DeepSpeedTrainer which will be MPI-based for now?
@astefanutti Any thoughts?

Maybe @kuizhiqing or @tenzen-y has more information when users want to use `mpirun` or `torchrun` to run DeepSpeed workloads?

docs/proposals/285-specialized-trainers/README.md
andreyvelich: @tariq-hasan is right. @szaher I am trying to understand how are we going to refactor the BuiltinTrainer interface once we implement the BaseTrainer? And how can we dynamically register new LLM fine-tuning framework backends ?

docs/proposals/285-specialized-trainers/README.md
andreyvelich: Yeah, I think this use-case should be covered by custom trainers when users have more control over training script. But I agree, we can discuss it in the followup proposals.

docs/proposals/285-specialized-trainers/README.md
astefanutti: Yes it validates the runtime has the correct framework label, just that the user has to provide it by name explicitly.

docs/proposals/285-specialized-trainers/README.md
astefanutti: Right, DeepSpeed can be run with `torchrun`, ideally the DeepSpeed Trainer would be "smart" enough to support both MPI and Torch runtime, and dynamically detect and adapt to the one that's available.

andreyvelich: I agree, we can always provide default option which can be overridden by the user.
However, I still think that DeepSpeed should be a Tier 1 trainer. Any thoughts @szaher?

docs/proposals/285-specialized-trainers/README.md
andreyvelich: I see, it's explained here: https://github.com/szaher/sdk/blob/1562655703734bf6fcf0da1eb5d1169be1005905/docs/proposals/285-specialized-trainers/README.md#runtime-auto-discovery

In that case, I’m wondering what additional capabilities specialized Trainers provide to users. Is the main benefit that users don’t need to worry about whether the Runtime is compatible with their code? Or are we planning to extend Tier 1 trainers to enable additional customization specifically for PyTorch code?

This also relates to a question I raised earlier: https://github.com/kubeflow/sdk/pull/308#discussion_r2891906231

docs/proposals/285-specialized-trainers/README.md
szaher: @andreyvelich should I rename the proposal to have to main abstract classes `ConfigTrainer` and `FuncTrainer` ?
then other trainers will inherit from those two?

andreyvelich: I am not sure, we need to discuss it the main differences between those trainers.

NarayanaSabari: @andreyvelich @szaher @tariq-hasan - I've redesigned [KEP-2839](https://github.com/kubeflow/trainer/pull/3263) to directly address this question.

The updated proposal introduces `ConfigTrainer` as a subclass of `BaseTrainer` specifically for **config-driven trainers** - where the framework's own CLI is the entrypoint (`trl sft ...`, `tune run ...`), not a user-supplied Python function.

This gives the hierarchy two branches under `BaseTrainer`:

```
                    BaseTrainer (ABC)                ← KEP-285
                         │
         ┌───────────────┼───────────────┐
         │               │               │
   TorchTrainer     JAXTrainer     ConfigTrainer      ← KEP-2839
   (func-based)     (func-based)   (config-driven)
                                        │
                         ┌──────────────┼──────────────┐
                         │              │              │
                  TorchTuneTrainer  TRLTrainer    (future: Unsloth,
                                                  LlamaFactory)
```

Key design points:
- `ConfigTrainer.get_train_func()` returns `None` - there is no user function
- `ConfigTrainer` adds `command` (ClassVar) and `to_args()` for CLI argument generation
- `TorchTuneConfig` becomes `TorchTuneTrainer(ConfigTrainer)` with a backward-compatible alias
- `BuiltinTrainer.config` type widens from `TorchTuneConfig` to `ConfigTrainer`
- Runtime auto-discovery works via the same `trainer.kubeflow.org/framework` label

Full details in the updated KEP: https://github.com/kubeflow/trainer/pull/3263

NarayanaSabari: **Update:** After discussion with @tariq-hasan, I've revised the KEP-2839 design. Instead of placing `ConfigTrainer` under `BaseTrainer`, **`LLMTrainer` is now a separate ABC** — parallel to `BaseTrainer`, not beneath it.

**Why the change:** Function-based trainers and config-driven trainers are architecturally distinct. Forcing them under one ABC creates dead methods (`get_train_func() → None`) and LSP violations. Separate ABCs let each hierarchy evolve independently.

**The user experience stays flat** — both ABCs are accepted through the same `trainer=` parameter:

```python
# Function-based (KEP-285)
client.train(trainer=TorchTrainer(func=my_train_fn, num_nodes=4))

# Config-driven (KEP-2839) — same parameter, same pattern
client.train(trainer=TRLTrainer(trainer_type=SFT, learning_rate=2e-5))
```

Updated hierarchy:

```
   TrainerClient.train(trainer=...)
   ─────────────────────────────────
   Accepts ANY of these:

   BaseTrainer (ABC)          LLMTrainer (ABC)
   (KEP-285, func-based)     (KEP-2839, config-driven)
      │                         │
      ├── TorchTrainer          ├── TorchTuneTrainer
      ├── JAXTrainer            └── TRLTrainer
      └── XGBoostTrainer
```

Shared design points unchanged:
- Both use `trainer.kubeflow.org/framework` for dispatch
- Both support `supported_frameworks` + `validate_runtime()` for auto-discovery
- No dependency between the two ABCs — KEP-285 can land independently

Updated KEP: https://github.com/kubeflow/trainer/pull/3263

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    Args:
+        func: The training function.
+        func_args: Arguments passed to the training function.
+        num_nodes: Number of nodes for distributed training.
+        resources_per_node: Resource requirements per node.
+        image: Optional custom container image.
+        num_proc_per_node: Number of MPI processes per node.
+    """
+
+    supported_frameworks: ClassVar[list[str]] = ["mpi"]
```
kramaranya: ** 🤖 AI Review **

`MPITrainer` declares `supported_frameworks=["mpi"]`, but the upstream Kubeflow Trainer ships zero ClusterTrainingRuntimes with `trainer.kubeflow.org/framework: mpi`. The actual MPI-based runtimes carry their own framework labels: `deepspeed-distributed` uses `framework: deepspeed`, `mlx-distributed` uses `framework: mlx`.

Auto-discovery for `MPITrainer` will always produce zero matches and raise an error on every cluster. This also contradicts Non-Goal #2 ("No new runtime labels or conventions").

MPI is a communication transport, not a training framework — the upstream correctly treats it that way. Options:
- (a) Remove `MPITrainer` from Tier 1
- (b) Propose a new upstream `mpi-distributed` ClusterTrainingRuntime (contradicts non-goals)
- (c) Redesign auto-discovery to match on MLPolicy type rather than framework label

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+    Class Attributes:
+        supported_frameworks: Framework identifiers this trainer supports.
+            Must match values of the `trainer.kubeflow.org/framework` label
+            on ClusterTrainingRuntime resources.
+    """
+
+    supported_frameworks: ClassVar[list[str]]
+
+    @abstractmethod
+    def get_train_func(self) -> Optional[Callable]:
```
kramaranya: ** 🤖 AI Review **

The name `BaseTrainer` suggests it is the root for all trainer abstractions, but `get_train_func() -> Optional[Callable]` as an abstract method makes it structurally function-centric. Config-driven trainers (TorchTune, TRL, Unsloth) that use framework CLIs as entrypoints would have to implement this method to always return `None` — an LSP smell.

The PR discussion already surfaced this: KEP-2839 proposes a parallel `LLMTrainer` ABC, and the author floated `ConfigTrainer`/`FuncTrainer`.

Suggestion: Either (a) rename to `FuncTrainer` to accurately scope it, or (b) explicitly document in this KEP that `BaseTrainer` is for function-based trainers only and state the relationship with KEP-2839 as a companion design. This avoids creating two competing "base" abstractions by accident.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+   runtime's `trainer.kubeflow.org/framework` label value is in the trainer's
+   `supported_frameworks` list.
+
+2. **Framework-specific checks** (in subclass overrides): For example, `MPITrainer`
+   could verify that the runtime's MPI policy source is configured correctly.
+
+Validation errors are raised as `ValueError` at submission time, *before* the
+`TrainJob` CR is created in the cluster.
+
+### Framework Argument Separation
```
kramaranya, line 934: ** 🤖 AI Review **

This section defines `get_framework_args() -> dict` for "framework CLI args not injected by controller" but never specifies where these land in the TrainJob spec. The author mentioned "it gets passed via the entrypoint" in PR discussion, but the spec-level mapping is the most technically critical integration point and it's undocumented.

Questions the KEP should answer:
1. Do framework args append to `.spec.trainer.args`? Merge into `.command`? Become env vars?
2. How do they interact with controller-injected args from MLPolicy? (e.g., `torchrun` already has `--max-restarts` in some runtime configurations)
3. What happens on conflict between SDK-provided and controller-injected args?

A concrete example showing `TorchTrainer(max_restarts=3)` rendered into a TrainJob YAML would resolve this.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+        return self.func
+
+    def get_train_func_args(self) -> Optional[dict]:
+        return self.func_args
+
+    def get_framework_args(self) -> dict:
+        return {}
+```
+
+### C. Tier 2: Application-Level Trainers
```
kramaranya: ** 🤖 AI Review **

DeepSpeed is deferred to Tier 2 here, yet the upstream Trainer ships `deepspeed-distributed` as a first-class ClusterTrainingRuntime with its own framework label, Dockerfile, and runtime image. @andreyvelich already suggested it should be Tier 1.

Additionally, @astefanutti noted DeepSpeed can be bootstrapped via either `mpirun` or `torchrun`, which means it doesn't cleanly subclass either `TorchTrainer` or `MPITrainer`. This is actually a good argument for `DeepSpeedTrainer` as its own Tier 1 trainer with `supported_frameworks=["deepspeed"]` matching the actual upstream label, independent of the launcher mechanism.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+## Migration and Backward Compatibility
+
+| Aspect | Impact |
+|---|---|
+| `CustomTrainer` | **No change.** Remains fully functional. `packages_to_install`, `pip_index_urls`, and `env` fields are retained. |
+| `CustomTrainerContainer` | **No change.** |
+| `BuiltinTrainer` | **No change.** |
+| `TrainerClient.train()` | **Additive only.** New `runtime_config` parameter is optional with default `None`. The `trainer` parameter type union is extended to include `BaseTrainer`. |
+| `TrainJobTemplate` | **No change in this proposal.** Future work can extend it to support `BaseTrainer` subclasses. |
+| `RuntimeConfig` vs `CustomTrainer` fields | When both `RuntimeConfig` and `CustomTrainer` fields are provided, `RuntimeConfig` takes precedence. This is documented but does not break existing code since `RuntimeConfig` defaults to `None`. |
```
kramaranya, line 1136: ** 🤖 AI Review **

The migration table says "RuntimeConfig takes precedence over CustomTrainer fields" but the overlap is substantial and "takes precedence" is ambiguous.

Concrete scenario: user passes `CustomTrainer(packages_to_install=["torch"], env={"FOO": "1"})` with `runtime_config=RuntimeConfig(packages=["numpy"], env={"BAR": "2"})`.

- Is it full replacement? (only numpy installed, only BAR set)
- Is it merge? (torch+numpy installed, FOO+BAR both set)
- If merge, what happens on `env` key collision?
- Are warnings emitted when both paths provide the same config?

Also: `BaseTrainer` subclasses inherit neither set of fields, so where does package installation config live for a `TorchTrainer` user? Only via the separate `RuntimeConfig`? This should be explicit.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+|---|---|---|
+| `get_train_func_args()` | User hyperparameters | `{"learning_rate": 1e-4, "epochs": 10}` |
+| `get_framework_args()` | Framework-specific CLI args not injected by the controller | `{"max-restarts": "3", "monitor-interval": "5"}` |
+
+Arguments that the Kubeflow Trainer controller already injects (e.g., `rdzv_endpoint`,
+`nnodes`, `nproc_per_node`, `node_rank`) are **excluded** from `get_framework_args()`.
+The specialized trainer documentation explicitly lists which arguments it manages vs.
+which the controller manages.
+
+### Backend Integration
```
kramaranya, line 950: ** 🤖 AI Review **

The proposal says "Each backend checks `isinstance(trainer, BaseTrainer)`" but `ContainerBackend` and `LocalProcessBackend` currently only support `CustomTrainer` and raise `ValueError` for anything else. The `RuntimeBackend.train()` abstract method signature must also change to accept `BaseTrainer` in the union type.

The KEP should include a backend compatibility matrix:
- Which backends support which trainer types?
- What error messages do users see for unsupported combinations (e.g., `JAXTrainer` on `LocalProcessBackend`)?
- Does `BaseTrainer` degrade gracefully to CustomTrainer-equivalent behavior on non-Kubernetes backends?

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+        raise ValueError(
+            f"Multiple runtimes found for frameworks "
+            f"{trainer.supported_frameworks}: {[r.name for r in matching]}. "
+            f"Please specify the runtime explicitly."
+        )
+
+    return matching[0]
+```
+
+### Runtime Validation
```
kramaranya, line 871: ** 🤖 AI Review **

`validate_runtime()` checks framework labels in the SDK client, but as @andreyvelich noted, validation is canonically the control plane's responsibility via webhooks. SDK-side validation creates a second source of truth that can go stale: if the controller adds new valid framework combinations, SDK users on older versions get false rejections.

Suggestion: Clarify that SDK validation is **advisory fail-fast UX** (not authoritative), document the version-coupling risk, and consider making it a warning rather than a hard exception — or provide a `skip_validation` escape hatch.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+5. Maintain 100% backward compatibility with the existing `CustomTrainer`,
+   `CustomTrainerContainer`, `BuiltinTrainer`, and `TrainerClient.train()` APIs.
+
+### Non-Goals
+
+1. **Controller/CRD changes.** This proposal is SDK-only. No changes to the Kubeflow
+   Trainer controller, `TrainJob` CRD, or `ClusterTrainingRuntime` CRD are required.
+2. **New runtime labels or conventions.** We rely on the existing
+   `trainer.kubeflow.org/framework` label already required on all runtimes.
+3. **Deprecating `CustomTrainer` or `BuiltinTrainer`.** Both remain supported.
```
kramaranya, line 177: ** 🤖 AI Review **

This non-goal says BuiltinTrainer is not deprecated, but the author acknowledged in PR discussion "I think we don't need the builtin Trainer." The proposal creates three unrelated abstractions (`CustomTrainer`, `BuiltinTrainer`, `BaseTrainer` subclasses) sharing no common ancestor.

Suggestion: Add a forward-looking statement clarifying intent. Even a sentence like "BuiltinTrainer will be evaluated for migration to BaseTrainer in a future KEP" would resolve the ambiguity and help KEP-2839 align its design. Without this, users and downstream KEPs cannot tell whether `BaseTrainer` is the future foundation or a parallel track.

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+### 2. Use Pydantic `BaseModel` instead of `@dataclass`
+
+Use Pydantic for automatic validation, serialization, and schema generation.
+
+**Rejected because:**
+- The existing SDK codebase uses `@dataclass` exclusively. Introducing Pydantic
+  would add a dependency and create an inconsistency in the codebase.
+- Pydantic validation can be replicated with `__post_init__` where needed.
+
+### 3. Put RuntimeConfig inside BaseTrainer instead of as a separate parameter
```
kramaranya, line 1309: ** 🤖 AI Review **

@andreyvelich raised whether `runtime_config` should be part of the Trainer rather than a separate `train()` parameter, especially for initializer env customization. This alternative was rejected ("orthogonal concern") but the initializer use case isn't addressed.

If `RuntimeConfig` is orthogonal to training logic, why is it scoped to `train()` and not to `TrainerClient` or the backend? Conversely, if it configures the trainer's execution environment, why is it separate from the trainer?

This should be listed as an explicit open question with trade-offs documented, particularly: does `RuntimeConfig` also apply to initializer containers?

kramaranya (commented):
** 🤖 AI Review **

docs/proposals/285-specialized-trainers/README.md
```diff
@@ -0,0 +1,1000 @@
...
+class BaseTrainer(ABC):
+    """Abstract base class for all specialized trainer implementations.
+
+    Class Attributes:
+        supported_frameworks: Framework identifiers this trainer supports.
+            Must match values of the `trainer.kubeflow.org/framework` label
+            on ClusterTrainingRuntime resources.
+    """
+
+    supported_frameworks: ClassVar[list[str]]
```
kramaranya: ** 🤖 AI Review **

`supported_frameworks: ClassVar[list[str]]` allows a trainer to declare compatibility with multiple frameworks (e.g., a future `DeepSpeedTrainer` with `["deepspeed", "torch"]` since DeepSpeed can use torchrun). But auto-discovery filters by a single framework label per runtime.

If a trainer lists multiple frameworks and multiple matching runtimes are found across different frameworks, the disambiguation is undefined. Consider either constraining to a single framework per trainer class (`ClassVar[str]`) or documenting the multi-framework auto-discovery priority.

docs/proposals/285-specialized-trainers/README.md
szaher: this means the `func_args` parameter doesn't stop users from adding `torch` parameters like `rdzv_endpoint` and `nnodes` 

users who usually train manually or using different frameworks might calculate those parameters manually and add them to their func_args which is not required and must be prohibited because it conflicts with the kuberflow trainer

docs/proposals/285-specialized-trainers/README.md
szaher: we should add a mechanism to automatically filter out parameters that conflict with the backend. This could be done by either sdk or trainer.
WDYT?

docs/proposals/285-specialized-trainers/README.md
szaher: agree. we need a clear and concise definition of what is a `Framework`.
Regarding DeepSpeed I have added a dedicated `DeepSpeedTrainer` https://github.com/szaher/sdk/blob/56e2ed7ab2e70cd2a41017068c5c7c9e4c2e69ca/docs/proposals/285-specialized-trainers/README.md

docs/proposals/285-specialized-trainers/README.md
szaher: I think the name makes more sense as we have `TrainingRuntime` and a configuration for it to enable the user to customize or pass some parameters to it via `RuntimeConfig`

PR #313, opened 2026-02-22
feat: Run dataset and model initializers in parallel

### LINKED ISSUE #290: feat(container): Run dataset and model initializers in parallel

Support for running dataset and model initializer containers in parallel instead of sequentially in the container backend.

### Why is this needed?

Currently, when both dataset and model initializers are configured, they run one after another. Since Docker and Podman support multiple containers mounting the same volume simultaneously, running them in parallel would reduce the total initialization time - especially beneficial when downloading large datasets and models.

Reference: #188

### Love this feature?

priyank766:
Currently, the [ContainerBackend](cci:2://file:///c:/Users/Priyank/Documents/CODE/gsoc-26/kubeflow-sdk/kubeflow/trainer/backends/container/backend.py:67:0-869:59) (Docker/Podman) runs dataset and model initializers sequentially. For Large Language Model (LLM) fine-tuning or heavy data workloads, downloading both a massive dataset and a base model separately can add significant overhead to the job startup time.

This PR refactors the [_run_initializers](cci:1://file:///c:/Users/Priyank/Documents/CODE/gsoc-26/kubeflow-sdk/kubeflow/trainer/backends/container/backend.py:513:4-577:73) logic to use `concurrent.futures.ThreadPoolExecutor`. If both a dataset and a model are configured, they are now initialized in parallel threads. This reduces the total initialization time to the duration of the longest single download, rather than the sum of both.

**Key technical changes:**
- Refactored [kubeflow/trainer/backends/container/backend.py](cci:7://file:///c:/Users/Priyank/Documents/CODE/gsoc-26/kubeflow-sdk/kubeflow/trainer/backends/container/backend.py:0:0-0:0) to use a thread pool for parallel initializer dispatch.
- Added `import concurrent.futures` to the backend.
- Ensured thread safety and proper error propagation (if one initializer fails, the main thread correctly identifies the failure and cleans up).
- Verified that shared volume mounts work correctly during concurrent writes from separate containers.

Fixes #290

**Checklist:**

- [x] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing (Internal performance improvement, no change to public API surface)
- [x] Unit tests pass: `uv run pytest kubeflow/trainer/backends/container/backend_test.py`
- [x] Code adheres to [Ruff formatting/linting standards](https://github.com/kubeflow/sdk/blob/main/CONTRIBUTING.md#coding-style) verified via `make verify`.

---

**Verification Results:**
Ran the specific backend tests to confirm the threading logic works as expected without race conditions:
```bash
uv run pytest kubeflow/trainer/backends/container/backend_test.py
# Result: 19 passed

### DIFF
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -543,11 +543,11 @@ class ContainerBackend(RuntimeBackend):
 
         # Use ThreadPoolExecutor to run configured initializers in parallel
         futures = {}
-        with concurrent.futures.ThreadPoolExecutor(max_workers=len(initializer_configs) or 1) as executor:
+        with concurrent.futures.ThreadPoolExecutor(
+            max_workers=len(initializer_configs) or 1
+        ) as executor:
             for name, init in initializer_configs:
-                container_utils.maybe_pull_image(
-                    self._adapter, init.image, self.cfg.pull_policy
-                )
+                container_utils.maybe_pull_image(self._adapter, init.image, self.cfg.pull_policy)
                 logger.debug("Queueing %s initializer", name)
                 future = executor.submit(
                     self._run_single_initializer,

### CODE REVIEW

File: kubeflow/trainer/backends/container/backend.py [hunk 1]
```diff
@@ -529,35 +530,52 @@ def _run_initializers(
...
+                        workdir=workdir,
+                        network_id=network_id,
+                    )
+                )
+
+            # Wait for all initializers to complete and raise exceptions if any failed
+            for future in concurrent.futures.as_completed(futures):
+                future.result()
+
+        logger.debug("All configured initializers completed in parallel")
```
Fiona-Waters: Could this be updated to specify the initializers that completed, just for clarity.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -529,35 +530,52 @@ def _run_initializers(
...
+                        self._run_single_initializer,
+                        job_name=job_name,
+                        container_init=dataset_init,
+                        workdir=workdir,
+                        network_id=network_id,
+                    )
+                )
+
+            # Dispatch model initializer if configured
+            if initializer.model:
```
Fiona-Waters: There is some duplication here with the model and dataset initializers - could this be updated to use a loop perhaps? It would be good to make it easier to add other initializers in the future also.

kubeflow/trainer/backends/container/backend.py
```diff
@@ -529,35 +530,40 @@ def _run_initializers(
...
+                    job_name=job_name,
+                    container_init=init,
+                    workdir=workdir,
+                    network_id=network_id,
+                )
+                futures[future] = name
+
+            # Wait for all initializers to complete and raise exceptions if any failed
+            for future in concurrent.futures.as_completed(futures):
+                future.result()
```
kramaranya, line 563: I don't think we need threads to get parallel downloads, we can just start both containers first as sequential API calls and wait for them

priyank766: Thanks for the suggestion! Just to make sure I understand the approach correctly — would the idea be to call create_and_start_container() for each initializer sequentially, and then loop through wait_for_container() for each? That way Docker handles the actual parallelism while we simplify the Python side.

One thing I wanted to check — _run_single_initializer also handles cleanup and error logging per container. Would you prefer that logic stays per-container in the wait loop, or should it be restructured differently?

Happy to refactor either way, just want to make sure it aligns with what you have in mind.
what should we finalize??@kramaranya & @Fiona-Waters

Fiona-Waters: We could do either, there are pros and cons. Since we just have 2 initializers right now then what @kramaranya suggests should work fine.

priyank766: @kramaranya any specific thing you have in mind?

andreyvelich: @Fiona-Waters @kramaranya @priyank766 Any pending items to move forward with this PR?

priyank766: @kramaranya suggestions needed for the this maybe you can also help @andreyvelich what should be the approach 
As for the issue current files changes are Completed and also tested by @Fiona-Waters 
Right now we only have two intialisers but for future there could be more added so 

> Thanks for the suggestion! Just to make sure I understand the approach correctly — would the idea be to call create_and_start_container() for each initializer sequentially, and then loop through wait_for_container() for each? That way Docker handles the actual parallelism while we simplify the Python side.
> 
> One thing I wanted to check — _run_single_initializer also handles cleanup and error logging per container. Would you prefer that logic stays per-container in the wait loop, or should it be restructured differently?
> 
> Happy to refactor either way, just want to make sure it aligns with what you have in mind. what should we finalize??@kramaranya & @Fiona-Waters

Fiona-Waters: There's nothing pending from my side. The current implementation works. Anya's suggestion could also work but may need to be updated in the future if there are lots more initializers added, IMHO.

Issue #315, opened 2026-02-22
LocalProcessBackend.__get_job_status never returns Complete


kevo-1:
### What happened?

**What happened?**

`LocalProcessBackend.__get_job_status` has a fallthrough issue in its status resolution logic. When all steps are `Complete`, none of the conditions match and the function returns `Created` instead of `Complete`.

**Steps to Reproduce:**
> Found via code review — no execution required. The bug is in `localprocess/backend.py` in `__get_job_status`. 

**Root Cause:**

The `__get_job_status` method checks for `FAILED`, `RUNNING`, and `CREATED` statuses, but has no condition handling the case where all steps are `Complete`. The `else` branch incorrectly returns `TRAINJOB_CREATED` as a fallback.

```python
  def __get_job_status(self, job: LocalBackendJobs) -> str:
      statuses = [_step.job.status for _step in job.steps]
      # if status is running or failed will take precedence over completed
      if constants.TRAINJOB_FAILED in statuses:
          status = constants.TRAINJOB_FAILED
      elif constants.TRAINJOB_RUNNING in statuses:
          status = constants.TRAINJOB_RUNNING
      elif constants.TRAINJOB_CREATED in statuses:
          status = constants.TRAINJOB_CREATED
      else:
          status = constants.TRAINJOB_CREATED

      return status
```

### What did you expect to happen?

`wait_for_job_status()` should return successfully once all steps report `Complete`, without timing out.
> This does not affect production workloads, but impacts local development workflows by causing every successful job to hang for 600 seconds before timing out.

### Impacted by this bug?

Give it a 👍 We prioritize the issues with most 👍

priyank766:
I have tried to solve the TrainJob Complete issue in this PR
#340 if anyone one assigned can review it that would be great

Issue Fixes : #338, #315

@Fiona-Waters @szaher @kramaranya

priyank766:
@Fiona-Waters @kramaranya  I Think this issue should be closed now right  ?
PR : https://github.com/kubeflow/sdk/pull/340

andreyvelich:
Thanks @priyank766!
/close

PR #316, opened 2026-02-22
fix: LocalProcessBackend.__get_job_status never returns Complete

kevo-1:
**Summary:**
Fixes a fallthrough issue in `LocalProcessBackend.__get_job_status` where the status resolution logic had no condition for when all steps are `Complete`, causing the `else` branch to incorrectly return `TRAINJOB_CREATED` as a fallback.

**Problem:**
When all steps report `TRAINJOB_COMPLETE`, none of the `if/elif` branches matched, so `wait_for_job_status()` would never receive the `Complete` status and would always time out after 600 seconds on local development runs.

**Fix:**
Added an explicit `elif` check before the `else` fallback:
```python
elif all(s == constants.TRAINJOB_COMPLETE for s in statuses):
    status = constants.TRAINJOB_COMPLETE
```

**Testing:**
- [x] Existing tests pass

**Fixes**: #315

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -266,10 +266,8 @@ class LocalProcessBackend(RuntimeBackend):
             status = constants.TRAINJOB_RUNNING
         elif constants.TRAINJOB_CREATED in statuses:
             status = constants.TRAINJOB_CREATED
-        elif all(s == constants.TRAINJOB_COMPLETE for s in statuses):
-            status = constants.TRAINJOB_COMPLETE
         else:
-            status = constants.TRAINJOB_CREATED
+            status = constants.TRAINJOB_COMPLETE
 
         return status
 


andreyvelich (commented):
Thanks for this fix @kevo-1!
Please can you add unit tests to verify that Job has the correct statuses after running `get_job()`: https://github.com/kevo-1/sdk/blob/2427e0b30a6c79e82b9489478aa1923eee583d46/kubeflow/trainer/backends/localprocess/backend_test.py#L399

kubeflow/trainer/backends/localprocess/backend.py
```diff
@@ -266,6 +266,8 @@ def __get_job_status(self, job: LocalBackendJobs) -> str:
             status = constants.TRAINJOB_RUNNING
         elif constants.TRAINJOB_CREATED in statuses:
             status = constants.TRAINJOB_CREATED
+        elif all(s == constants.TRAINJOB_COMPLETE for s in statuses):
```
ruskaruma: This check isn't needed. If we already ruled out Failed, Running, and Created above, the only thing left is Complete. You can just change the else to return TRAINJOB_COMPLETE directly. As is, the else is now unreachable and still
   returns Created.

kevo-1: Noted, thank you.

PR #322, opened 2026-02-25
fix(optimizer): prevent input mutation in optimize()

### LINKED ISSUE #323: OptimizerClient: optimize() mutates caller's input objects
### What happened?

While reading through the optimizer module, I noticed that `KubernetesBackend.optimize()` in [`backends/kubernetes/backend.py`](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/backends/kubernetes/backend.py) mutates the caller's input objects in-place:

- Line 97: sets `trial_template.trainer.func_args = {}` on the caller's object
- Line 100: `param_spec.name = param_name` overwrites the caller's search space objects
- Line 110: injects Katib template strings like `${trialParameters.lr}` into `func_args`

If a user reuses the same `trial_template` or `search_space` for a second `optimize()` call, `func_args` contains Katib template strings instead of the original values, and `param_spec.name` carries over from the first call.

The trainer module handles this correctly — it uses `copy.deepcopy()` before modification ([trainer/backends/kubernetes/backend.py line 230](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L230)).

`optimize()` should deep copy `trial_template` and `search_space` before modifying them, so the caller's objects remain unchanged. This matches the trainer module's approach.

### Environment

Main branch. Verified via code review and local reproduction.

### Impacted by this bug?

ruskaruma:

`KubernetesBackend.optimize()` mutates the caller's `trial_template` and `search_space` objects in-place during Katib experiment construction. Line 97, 100, and 110 overwrite `func_args`, `param_spec.name`, and inject Katib template strings directly into the caller's objects. If a user reuses these for a second `optimize()` call, the inputs carry stale state from the first call.

This adds `copy.deepcopy()` for both inputs at the start of `optimize()`, matching the pattern used in the trainer module ([backend.py line 230](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L230)). Also adds unit tests verifying caller objects remain unchanged.

Fixes #323

**Checklist:**

### DIFF
--- a/kubeflow/optimizer/backends/kubernetes/backend_test.py
+++ b/kubeflow/optimizer/backends/kubernetes/backend_test.py
@@ -77,7 +77,7 @@ def optimizer_backend():
         ),
     ],
 )
-def test_optimize_does_not_mutate_inputs(optimizer_backend, test_case):
+def test_optimize(optimizer_backend, test_case):
     """Test that optimize() does not mutate the caller's input objects."""
     print("Executing test:", test_case.name)
 


kubeflow/optimizer/backends/kubernetes/backend.py
```diff
@@ -82,6 +83,10 @@ def optimize(
         if not search_space:
             raise ValueError("Search space must be set.")
 
+        # Deep copy inputs to avoid mutating the caller's objects.
+        trial_template = copy.deepcopy(trial_template)
+        search_space = copy.deepcopy(search_space)
+
```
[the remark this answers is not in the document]
ruskaruma: Updated. Linked to the correct issue now.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -0,0 +1,118 @@
...
+            config={
+                "search_space": {
+                    "lr": Search.uniform(min=0.001, max=0.1),
+                    "epochs": Search.choice([10, 20, 30]),
+                },
+            },
+        ),
+    ],
+)
+def test_optimize_does_not_mutate_inputs(optimizer_backend, test_case):
```
andreyvelich: Please can you name this test `test_optimize()` according to this guidance: https://github.com/ruskaruma/sdk/blob/41062d3b3220ff51eee4c3ad6dc3622a7b0d31aa/AGENTS.md#3-testing-requirements

andreyvelich (commented):
thanks @ruskaruma, that makes sense, just small nit.

kubeflow/optimizer/backends/kubernetes/backend_test.py
ruskaruma: done!

PR #325, opened 2026-02-25
fix(optimizer): add missing get_job_events() to RuntimeBackend base c…

### LINKED ISSUE #321: OptimizerClient: get_job_events() missing from RuntimeBackend base class
### What happened?

While reading through the optimizer module, I noticed that `get_job_events()` is called by `OptimizerClient` ([[optimizer_client.py#L248](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/api/optimizer_client.py#L248)](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/api/optimizer_client.py#L248)) and implemented in `KubernetesBackend` ([[backend.py#L337](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/backends/kubernetes/backend.py#L337)](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/backends/kubernetes/backend.py#L337)), but it's not declared in the `RuntimeBackend` base class ([[base.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/backends/base.py)](https://github.com/kubeflow/sdk/blob/main/kubeflow/optimizer/backends/base.py)).

The base class has the other 7 methods (`optimize`, `list_jobs`, `get_job`, `get_job_logs`, `get_best_results`, `wait_for_job_status`, `delete_job`) but `get_job_events` is missing. So if someone writes a new backend, they won't know they need to implement it until they hit an `AttributeError` at runtime.

I checked the trainer module and it handles this specific:  `trainer/backends/base.py` has `get_job_events()` with a default `return []` ([[line 70](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/base.py#L70)](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/base.py#L70)), which makes sense since not every backend will support events.

`get_job_events()` should be declared in the optimizer `RuntimeBackend` base class, ideally with a default `return []` to match the trainer pattern.

### Environment

Main branch. This is a code gap that I could see here and no cluster is needed to verify.

ruskaruma:

`get_job_events()` is called by `OptimizerClient` (optimizer_client.py#L248) and implemented in `KubernetesBackend` (backend.py#L337), but it's not declared in the `RuntimeBackend` base class. The base class defines the other 7 methods but omits this one, so anyone implementing a custom backend won't know they need it until they hit an `AttributeError` at runtime.

This adds `get_job_events()` with a default `return []` to the optimizer `RuntimeBackend`, matching the trainer module's pattern (trainer/backends/base.py line 70).

Fixes #321

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -798,7 +798,7 @@ class ContainerBackend(RuntimeBackend):
                 yield f"Error getting logs: {e}\n"
 
     def get_job_events(self, name: str) -> list[types.Event]:
-        return []
+        raise NotImplementedError()
 
     def wait_for_job_status(
         self,
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -209,6 +209,8 @@ class LocalProcessBackend(RuntimeBackend):
             # (adjust args if stream_logs has different signature)
             yield from _step.job.logs(follow=follow)
 
+    def get_job_events(self, name: str) -> list[types.Event]:
+        raise NotImplementedError()
 
     def wait_for_job_status(
         self,


kubeflow/optimizer/backends/base.py
```diff
@@ -72,6 +72,9 @@ def wait_for_job_status(
     ) -> OptimizationJob:
         raise NotImplementedError()
 
+    def get_job_events(self, name: str) -> list[Event]:
```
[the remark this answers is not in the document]
andreyvelich: This is not needed.

kubeflow/optimizer/backends/base.py
```diff
@@ -72,6 +72,9 @@ def wait_for_job_status(
...
+        return []
```
andreyvelich: Good catch, please. can you also update it to the following:

```suggestion
        raise NotImplementedError()
```

kubeflow/optimizer/backends/base.py
andreyvelich, line 76: Add annotation since it is abstract method.
```
@abc.abstractmethod
```

kubeflow/optimizer/backends/base.py
```diff
@@ -72,6 +72,17 @@ def wait_for_job_status(
...
 
+    def get_job_events(self, name: str) -> list[Event]:
+        """Return events associated with an optimization job.
+
+        Args:
+            name: Name of the optimization job to retrieve events for.
+
+        Returns:
+            A list of Event objects describing the job lifecycle.
+        """
```
andreyvelich: You can remove this docstring, and please also update it Trainer: https://github.com/ruskaruma/sdk/blob/23cb0e4aaaff9366bd667803c0e5f3d0f32dad9f/kubeflow/trainer/backends/base.py#L69-L80

kubeflow/trainer/backends/base.py
```diff
@@ -68,14 +68,6 @@ def get_job_logs(
...
     def get_job_events(self, name: str) -> list[types.Event]:
-        """Get events for a TrainJob.
-
-        Args:
-            name: Name of the TrainJob.
-
-        Returns:
-            A list of Event objects associated with the TrainJob.
-        """
         return []
```
andreyvelich: Can you also replace it with  NotImplementedError() and add `@abc.abstractmethod`

kubeflow/trainer/backends/container/backend.py
```diff
@@ -797,6 +797,9 @@ def get_job_logs(
                 logger.warning(f"Failed to get logs for {container['name']}: {e}")
                 yield f"Error getting logs: {e}\n"
 
+    def get_job_events(self, name: str) -> list[types.Event]:
+        return []
```
andreyvelich: This is no needed, since we should raise NotImplementedError exception for this backend.

kubeflow/trainer/backends/localprocess/backend.py
```diff
@@ -209,6 +209,9 @@ def get_job_logs(
             # (adjust args if stream_logs has different signature)
             yield from _step.job.logs(follow=follow)
 
+    def get_job_events(self, name: str) -> list[types.Event]:
+        return []
```
andreyvelich:
```suggestion
```

kubeflow/trainer/backends/container/backend.py
andreyvelich: @ruskaruma Please address this, also check why E2Es and CI are failing.

ruskaruma: on it

ruskaruma: The CI failed because removing get_job_events from localprocess/backend.py left it as an unimplemented @abc.abstractmethod, which causes Python to raise TypeError when instantiating LocalProcessBackend. I've added get_job_events  back to both localprocess and container backends with raise NotImplementedError() instead of return [].

PR #328, opened 2026-02-25
fix(trainer): handle falsy values in get_args_from_peft_config

### LINKED ISSUE #327: bug(trainer): get_args_from_peft_config drops falsy values (e.g. lora_dropout=0.0)
## What happened?
In `get_args_from_peft_config`, falsy values such as:
- `lora_dropout=0.0`
- `apply_lora_to_mlp=False`

are not included in the generated args.

## Expected behavior
These values should be preserved in the generated arguments, since they are valid user configurations.

Example:
Input:
    `LoraConfig(lora_dropout=0.0)`

Expected output:
    `["model.lora_dropout=0.0", ...]`

## Actual behavior
The values are silently skipped and not included in the args.

## Root cause
The condition:
    `if value:`

treats valid falsy values (`0.0`, `False`) as absent.

## Proposed fix
Replace:
    `if value:`

with:
    `if value is not None:`

## Additional context
- This change ensures explicit user configurations are respected.
- Added unit tests covering these cases.

krishdef7:
Fixes #327

## Summary
Replace `if value:` with `if value is not None:` to ensure falsy values
like `0.0` and `False` are not silently dropped.

## Changes
- Updated condition in `get_args_from_peft_config`
- Added unit tests for:
  - `lora_dropout=0.0`
  - `apply_lora_to_mlp=False`

## Impact
Ensures explicit user configurations are respected and not ignored.

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils_test.py
+++ b/kubeflow/trainer/backends/kubernetes/utils_test.py
@@ -741,4 +741,3 @@ def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
         assert test_case.expected_status == FAILED
         assert type(e) is test_case.expected_error
     print("test execution complete")
-    


kubeflow/trainer/backends/kubernetes/utils.py
ruskaruma, line 563: same bug: train_on_input=False gets silently dropped
  if dataset_preprocess_config.train_on_input:  # should be `is not None`

Would you be open to including this fix + a test case for it in this PR? Happy to help if needed.

krishdef7: @ruskaruma Done — extended the fix to get_args_from_dataset_preprocess_config as well, with tests covering both train_on_input=False and train_on_input=True. All 32 tests passing locally.

astefanutti:
@krishdef7 thanks!

/ok-to-test
/lgtm

PR #336, opened 2026-02-27
feat(trainer): Support NPU labels in TrainJob device

sujalshah-bit:
Add support for NPU resource labels in container resource limits,
resolving the existing TODO to support additional accelerator types
(e.g., NPUs).

Fixes NONE

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -41,7 +41,6 @@ def get_container_devices(
     # TODO (andreyvelich): We should discuss how to get container device type.
     # Potentially, we can use the trainer.kubeflow.org/device label from the runtime or
     # node types.
-    # TODO (andreyvelich): Support other resource labels (e.g. NPUs).
     if constants.GPU_LABEL in resources.limits:
         device = constants.GPU_LABEL.split("/")[1]
         device_count = resources.limits[constants.GPU_LABEL].actual_instance
@@ -55,6 +54,13 @@ def get_container_devices(
         mig_key = mig_keys[0]
         device = mig_key.split("/")[1]
         device_count = resources.limits[mig_key].actual_instance
+    elif constants.NPU_LABEL in resources.limits:
+        # huawei.com/Ascend910 is the K8s extended resource key for Huawei's Ascend 910 NPU.
+        # Unlike GPU/TPU, the suffix ("Ascend910") doesn't generically describe the device
+        # type, so we explicitly return "npu" as the device string instead of deriving it
+        # from the key.
+        device = "npu"
+        device_count = resources.limits[constants.NPU_LABEL].actual_instance
     elif constants.CPU_LABEL in resources.limits:
         device = constants.CPU_LABEL
         device_count = resources.limits[constants.CPU_LABEL].actual_instance
--- a/kubeflow/trainer/backends/kubernetes/utils_test.py
+++ b/kubeflow/trainer/backends/kubernetes/utils_test.py
@@ -61,6 +61,18 @@ def _build_runtime() -> types.Runtime:
             },
             expected_error=ValueError,
         ),
+        TestCase(
+            name="single NPU limit returns device and count",
+            expected_status=SUCCESS,
+            config={
+                "resources": models.IoK8sApiCoreV1ResourceRequirements(
+                    limits={
+                        constants.NPU_LABEL: models.IoK8sApimachineryPkgApiResourceQuantity(2),
+                    }
+                )
+            },
+            expected_output=("npu", "2.0"),
+        ),
     ],
 )
 def test_get_container_devices(test_case: TestCase):
--- a/kubeflow/trainer/constants/constants.py
+++ b/kubeflow/trainer/constants/constants.py
@@ -104,6 +104,12 @@ GPU_MIG_PREFIX = "nvidia.com/mig-"
 # The label for TPU in the container resources.
 TPU_LABEL = "google.com/tpu"
 
+# The label for NPU in the container resources.
+# Unlike GPU (nvidia.com/gpu) and TPU (google.com/tpu), NPU vendors don't use
+# a consistent suffix that maps to device type, so we hardcode the full
+# Huawei Ascend 910 resource label until a general strategy is agreed upon.
+NPU_LABEL = "huawei.com/Ascend910"
+
 # The label key to identify the JobSet name of the Pod.
 JOBSET_NAME_LABEL = "jobset.sigs.k8s.io/jobset-name"
 


kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -41,7 +41,7 @@ def get_container_devices(
     # TODO (andreyvelich): We should discuss how to get container device type.
     # Potentially, we can use the trainer.kubeflow.org/device label from the runtime or
     # node types.
-    # TODO (andreyvelich): Support other resource labels (e.g. NPUs).
+    # TODO (andreyvelich): Support other resource labels.
```
andreyvelich: You can remove this TODO statement.


```suggestion
```

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -55,6 +55,13 @@ def get_container_devices(
         mig_key = mig_keys[0]
         device = mig_key.split("/")[1]
         device_count = resources.limits[mig_key].actual_instance
+    elif npu_keys := [k for k in resources.limits if k.endswith(constants.NPU_LABEL_SUFFIX)]:
+        if len(npu_keys) > 1:
+            raise ValueError(f"Multiple NPU resource types are not supported yet: {npu_keys}")
```
andreyvelich: Do we need this?

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -55,6 +55,13 @@ def get_container_devices(
...
+
+        npu_key = npu_keys[0]
```
andreyvelich: Why are you taking the first item? I thought NPU resources are set similarly to TPU and GPU

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -61,6 +61,31 @@ def _build_runtime() -> types.Runtime:
             },
             expected_error=ValueError,
         ),
+        TestCase(
+            name="single NPU limit returns device and count",
+            expected_status=SUCCESS,
+            config={
+                "resources": models.IoK8sApiCoreV1ResourceRequirements(
+                    limits={
+                        "huawei.com/npu": models.IoK8sApimachineryPkgApiResourceQuantity(2),
```
andreyvelich: Remove vendors from unit tests

```suggestion
                        "example.com/npu": models.IoK8sApimachineryPkgApiResourceQuantity(2),
```

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -61,6 +61,31 @@ def _build_runtime() -> types.Runtime:
...
+                    }
+                )
+            },
+            expected_output=("npu", "2.0"),
+        ),
+        TestCase(
+            name="multiple NPU resource types are not supported",
```
andreyvelich: This should not be the case.

kubeflow/trainer/constants/constants.py
```diff
@@ -104,6 +104,9 @@
 # The label for TPU in the container resources.
 TPU_LABEL = "google.com/tpu"
 
+# The Suffix label for NPU in the container resources.
+NPU_LABEL_SUFFIX = "/npu"
```
andreyvelich: I thought the resource name is: huawei.com/Ascend910.
@danish9039 Do you know details about it?
Ref: https://github.com/kubeflow/sdk/pull/264

kubeflow/trainer/backends/kubernetes/utils_test.py
sujalshah-bit: I understand the preference for vendor-neutral tests, but the current NPU implementation is tied to a specific vendor label (`huawei.com/Ascend910`) similar to how GPU is tied to nvidia.com/gpu and TPU to `google.com/tpu`. 
If we want `example.com/npu` in the test, we'd need to change the implementation to use suffix matching (`/npu`) instead of an exact vendor label, but as discussed, that's not guaranteed to work for all NPU vendors (e.g. huawei.com/Ascend910 has no /npu suffix). Happy to go either way, just need clarity on which approach we're taking.

andreyvelich: I see, good point. Let's keep it as NPU label for now

andreyvelich (commented):
I am fine with starting supporting NPU via `huawei.com/Ascend910` label.
/lgtm
/cc @astefanutti @akshaychitneni @kramaranya @Fiona-Waters

kubeflow/trainer/constants/constants.py
```diff
@@ -104,6 +104,9 @@
 # The label for TPU in the container resources.
 TPU_LABEL = "google.com/tpu"
 
+# The label for NPU in the container resources.
+NPU_LABEL = "/npu"
+
```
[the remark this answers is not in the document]
Fiona-Waters: I think we should be a little clearer in comments as to why we are hardcoding this.

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -55,6 +54,10 @@ def get_container_devices(
         mig_key = mig_keys[0]
         device = mig_key.split("/")[1]
         device_count = resources.limits[mig_key].actual_instance
+    elif constants.NPU_LABEL in resources.limits:
+        # Can't derive cleanly from key unlike gpu/tpu
```
Fiona-Waters: Could you? update the coment to explain what the device label is and why we return npu?

kubeflow/trainer/constants/constants.py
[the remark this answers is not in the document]
andreyvelich: Maybe @JesseStutler @lowang-bh can help us to find the better alias for Huawei's Ascend chips?
Do you know what is the name users usually say when they want to use this AI chip?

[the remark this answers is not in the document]
JesseStutler: NPU isn't limited to the 910 series chips, it's definitely a bit hard-coded here. However, I'm not familar with the context of this file. If a user wants to use the NPU, and it's a 310 series chip, the they'll request will be like `huawei.com/Ascend310P`. So only the prefix `huawei.com/` is the same, unlike with GPUs. The suffix might vary depending on the specific chip.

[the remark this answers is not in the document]
andreyvelich: Would it be better to ask users provide the dedicated chip model? For example:
```py
trainer=CustomTrainer(
        func=torch_dist,
        num_nodes=3,
        resources_per_node={"Ascend310P": 2},
    ),

trainer=CustomTrainer(
        func=torch_dist,
        num_nodes=3,
        resources_per_node={"Ascend910C": 5},
    ),
```

We would just add the prefix there: `huawei.com/`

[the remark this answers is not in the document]
JesseStutler: I prefer this style as @andreyvelich mentioned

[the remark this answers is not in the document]
andreyvelich: @sujalshah-bit please can you update the PR to let users set the `Ascend***` device?

andreyvelich:
@sujalshah-bit Please could you address @Fiona-Waters comment, so we can move this forward?

andreyvelich:
Ping @sujalshah-bit.
Did you get a chance to address the small comments, so we can merge this PR?

sujalshah-bit:
@andreyvelich, sorry for the delay and any inconvenience. I keep forgetting about this. Could you please review the latest change :)

andreyvelich (commented):
Thanks @sujalshah-bit!
/ok-to-test
/lgtm
/approve

PR #337, opened 2026-02-27
chore: Add make format target for Python code formatting

sujalshah-bit:
Add a `make format` target to the Makefile to simplify Python code formatting
using Ruff.

Previously, you could only check if files were formatted or not, but now you can fix it using a make command.

Fixes NONE

**Checklist:**


andreyvelich (commented):
We don't need this since we use pre-commit: https://github.com/kubeflow/sdk/blob/main/.pre-commit-config.yaml
/close

PR #339, opened 2026-02-28
fix(trainer): make get_job_logs optionally fail fast when pod is missing

### LINKED ISSUE #317: get_job_logs silently returns empty iterator when no pod is found
When no pod is found for the requested step, the generator silently returns. The caller gets an empty iterator with no way to distinguish "job produced no logs" from "pod was never found."

```python
if pod_name is None:
    return
```

| File affected | Line |
|--|--|
| `kubernetes/backend.py` | `line: 443` |
| `optimizer/backends/kubernetes/backend.py` | `line: 245` |

Same behavior appears in both files above

### Was this behavior intended or should it raise an error?

sjiang83:
Fixes #317

The Issue
Currently, when calling TrainerClient.get_job_logs() using the Kubernetes backend, it silently returns an empty iterator if a pod isn't found for the requested step. This usually happens if the step is still Pending or the selector just comes up empty. This behavior creates a lot of ambiguity, especially for MCP tools and AI agents that rely on clear feedback to know what is actually going on.

What this PR does
I've added an opt-in strict flag to the method:
- strict=False (the default): Keeps the current behavior exactly as is, which might yield nothing.
- strict=True: Explicitly raises a PodNotFoundError if no pod is found for the requested TrainJob or step.

I also updated the signatures across the other Trainer backends to accept the strict argument so the interface stays consistent (it simply ignores the flag for non-Kubernetes backends). I applied this same strict logic to the Optimizer Kubernetes backend log retrieval as well.

Why this helps MCP and Agents
For agents to debug effectively, they absolutely need to be able to tell the difference between three things:
- The job exists but the logs just aren't ready or outputting yet.
- The pod actually cannot be found (which we now catch explicitly with strict=True).
- Standard permission or RBAC errors (which still bubble up from the Kubernetes client just like before).

By opting into strict=True, we get a much more deterministic observability flow without forcing a breaking change on existing SDK users.

Testing and Compatibility
- Tests: I added a unit test to verify that strict=True correctly raises an error when all steps are still in the Pending state.
- Backward compatibility: 100% safe. Since the default behavior remains unchanged, existing users won't notice any difference.

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -37,9 +37,12 @@ from kubeflow.trainer.types import types
 
 logger = logging.getLogger(__name__)
 
+
 class PodNotFoundError(RuntimeError):
     """No pod resolved for the requested job/step."""
+
     pass
+
 
 class KubernetesBackend(RuntimeBackend):
     def __init__(self, cfg: KubernetesBackendConfig):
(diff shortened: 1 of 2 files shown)


sjiang83:
All the CI tests are green across the board now. 

Just waiting on an LGTM and an approve label whenever a maintainer has a moment to take a look. Thanks again!

PR #340, opened 2026-02-28
fix(trainer): return TRAINJOB_COMPLETE when all steps are done

### LINKED ISSUE #338: LocalProcessBackend.__get_job_status never returns TRAINJOB_COMPLETE
### What happened?

`LocalProcessBackend.__get_job_status()` returns `TRAINJOB_CREATED` when all steps have finished, instead of `TRAINJOB_COMPLETE`.

The else branch on line 272 of `trainer/backends/localprocess/backend.py`:
```python
else:
    status = constants.TRAINJOB_CREATED  # should be TRAINJOB_COMPLETE
```

When no step is Failed, Running, or Created i.e. all steps have completed and this should return `TRAINJOB_COMPLETE`. Because it doesn't, `wait_for_job_status(status={TRAINJOB_COMPLETE})` will always timeout on LocalProcessBackend since the job can never reach Complete status.

`__get_job_status` should return `TRAINJOB_COMPLETE` when all steps are done. The trainer's KubernetesBackend handles this correctly.

### Environment

Main branch, found while reviewing the code.

### Impacted by this bug?

priyank766:
`LocalProcessBackend.__get_job_status()` returns `TRAINJOB_CREATED` when all steps have finished, instead of `TRAINJOB_COMPLETE`. This causes [wait_for_job_status()](cci:1://file:///c:/Users/Priyank/Documents/CODE/gsoc-26/kubeflow-sdk/kubeflow/trainer/backends/localprocess/backend.py:214:4-249:92) to always timeout (600s) on the local backend even when jobs complete successfully. This is a one-line fix in the `else` branch to return the correct status.

Fixes #338, #315

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -499,39 +499,63 @@ def test_name_option_sets_job_name(local_backend, mock_train_environment):
     assert job_name == custom_name
 
 
-def test_get_job_status(local_backend):
-    """Test LocalProcessBackend.__get_job_status()."""
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="all steps complete returns TRAINJOB_COMPLETE",
+            expected_status=SUCCESS,
+            config={
+                "step_statuses": [constants.TRAINJOB_COMPLETE, constants.TRAINJOB_COMPLETE],
+            },
+            expected_output=constants.TRAINJOB_COMPLETE,
+        ),
+        TestCase(
+            name="any step running returns TRAINJOB_RUNNING",
+            expected_status=SUCCESS,
+            config={
+                "step_statuses": [constants.TRAINJOB_COMPLETE, constants.TRAINJOB_RUNNING],
+            },
+            expected_output=constants.TRAINJOB_RUNNING,
+        ),
+        TestCase(
+            name="any step failed returns TRAINJOB_FAILED",
+            expected_status=SUCCESS,
+            config={
+                "step_statuses": [constants.TRAINJOB_FAILED, constants.TRAINJOB_RUNNING],
+            },
+            expected_output=constants.TRAINJOB_FAILED,
+        ),
+        TestCase(
+            name="any step created returns TRAINJOB_CREATED",
+            expected_status=SUCCESS,
+            config={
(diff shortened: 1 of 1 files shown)


kubeflow/trainer/backends/localprocess/backend.py
```diff
@@ -270,7 +270,7 @@ def __get_job_status(self, job: LocalBackendJobs) -> str:
         elif constants.TRAINJOB_CREATED in statuses:
             status = constants.TRAINJOB_CREATED
         else:
-            status = constants.TRAINJOB_CREATED
+            status = constants.TRAINJOB_COMPLETE
```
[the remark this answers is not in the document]
Fiona-Waters: @priyank766 can you please add a test case to cover this?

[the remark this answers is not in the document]
priyank766: yes sure i will add it and will give it to you for review 
@Fiona-Waters

Fiona-Waters, line 275: I think in this case if steps haven't initialised yet it might wrongly default to complete. Would it work to add this at the top of the function to cover this?
```
if not job.steps:
        return constants.TRAINJOB_CREATED
```

priyank766: my bad yes you are correct added the early return at the top of __get_job_status() to handle the empty steps case, and also included a test for it. Thanks 
@Fiona-Waters

kubeflow/trainer/backends/localprocess/backend_test.py
```diff
@@ -497,3 +497,41 @@ def dummy_func():
     )
 
     assert job_name == custom_name
+
+
+def test_get_job_status(local_backend):
```
andreyvelich: @priyank766 Please can you refactor unit test with TestCase fixture as said in: https://github.com/priyank766/sdk/blob/117a720a17fad73505b388fc7593c90a18537b12/AGENTS.md#3-testing-requirements

priyank766: @andreyvelich check again refactored test_get_job_status to use @pytest.mark.parametrize with the TestCase fixture as per AGENTS.md, covering all 5 scenarios (complete, running, failed, created, and no steps registered).

PR #341, opened 2026-02-28
fix(trainer): fallback to cluster-scoped runtime on 404 in get_runtime

### LINKED ISSUE #335: ClusterRuntime un-getable
### What happened?

It looks like with recent changes to the Kubernetes backend, ClusterRuntime are not retrievable. 

When attempting to get a ClusterRuntime, it can never go past the exception caught on L199: 
https://github.com/kubeflow/sdk/blob/db76de0ca8c32540b8a177a713e73f41c1bf81af/kubeflow/trainer/backends/kubernetes/backend.py#L174-L223

The change seems to come from https://github.com/kubeflow/sdk/pull/307. Based on skimming that PR, it looks like it probably worked originally with the `pass` but that was pulling out due to a suggestion from copilot. It doesn't need to raise the error there as that requires something above it to catch it. It should probably just print/log it instead of catching and raising again.

Disclaimer: I am not familiar with this codebase nor have I had time to fully dig into it, so maybe I missed up something in my configuration. Either way, it warrants a deeper look.

`get_runtime` should first try namespaced runtimes and then fallback to cluster runtimes.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

prabindersinghh:
### What this PR does

This PR fixes the `get_runtime()` fallback behavior in the Kubernetes backend.

Previously, when a namespaced `TrainingRuntime` returned a 404 error,
`get_runtime()` could incorrectly raise an error instead of properly
falling back to the cluster-scoped `ClusterTrainingRuntime`.

This change:

- Allows fallback to cluster-scoped runtime only when the namespaced call returns 404
- Preserves existing error handling for other API exceptions
- Adds a unit test to validate the fallback behavior
- Ensures all existing backend tests pass

### Fixes

Fixes #335

### Checklist

- [x] Unit tests added
- [x] Existing tests pass
- [x] No breaking changes


astefanutti (commented):
Thanks @prabindersinghh!

/ok-to-test

prabindersinghh:
Hi @astefanutti @kramaranya, following up on this PR.

Formatting has been fixed as requested. The E2E test failures appear to be infrastructure-related rather than caused by this change, which only modifies the `polling_interval >= timeout` validation guard.

Could you confirm if a `/retest` would help, or if there is anything else needed to move this forward? Happy to make any additional changes.

Issue #164, opened 2025-11-18
Integrate Kubeflow SDK with OpenTelemetry


andreyvelich:
### What you would like to be added?

We would like to explore opportunities to integrate Kubeflow SDK with Open Telemetry for better observability and telemetry of AI Workloads.
For example, we can take a look at OTel and GenAI integrations: https://opentelemetry.io/blog/2024/otel-generative-ai/

This was one of the user asks during Kubeflow Summit at KubeCon + CloudNativeCon 2025 NA.

cc @kubeflow/kubeflow-sdk-team @kubeflow/kubeflow-trainer-team @Fiona-Waters @austinlparker @alolita

### Why is this needed?

Improve observability experience for Kubeflow SDK users.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

Mahimasai:
Hi @kramaranya @dhanishaphadate @jaiakash,

I’ve been going through the Kubeflow SDK documentation and reviewing the TrainerClient and PipelinesClient implementations to better understand where OpenTelemetry instrumentation could fit naturally.

From my current understanding, public SDK entry points like train() and run_pipeline() seem like good candidates for top-level spans, while internal API interactions (e.g., CRD creation and status polling) could form child spans. I’m also exploring whether introducing a shared instrumentation utility in kubeflow/common would help avoid duplication across clients.

I’ve started experimenting locally with manual OpenTelemetry instrumentation in Python to better understand span hierarchy and context propagation.

Before diving deeper, I’d appreciate guidance on preferred instrumentation granularity — should tracing primarily focus on SDK method boundaries, or also include lower-level Kubernetes API interactions?

Looking forward to contributing and refining this approach.

prabindersinghh:
Hi  @kramaranya, @dhanishaphadate, @kramaranya -

I’ve gone through the issue discussion, the GSoC proposal PR in kubeflow/website (#4294), and the current SDK documentation.
Given that the SDK’s core promise is a Pythonic abstraction over Kubernetes, I believe any OpenTelemetry integration should:

- Remain fully optional and zero-overhead for non-users
- Avoid leaking Kubernetes-level complexity into user-facing APIs
- Follow a consistent span model across clients

To start in a focused way, I propose beginning with a minimal reference implementation in TrainerClient:

1) Instrument the top-level train() entry point
2) Add child spans for CRD creation, submission, and status polling

Establish a consistent naming pattern such as:

1) kubeflow.sdk.trainer.train
[...]

PR #348, opened 2026-03-03
fix(trainer): raise ValueError for missing containers instead of Stop…

1Ayush-Petwal:
`next()` on a bare generator with no default raises `StopIteration`, which
  Python 3.7+ (PEP 479) promotes to a `RuntimeError` with no pod/container
  context. Hardens two functions in `utils.py`:

- `get_trainjob_initializer_step`: replaced bare `next(gen)` with  `next(..., None)` + explicit `ValueError` naming the pod and listingactual containers found
- `get_trainjob_node_step`: same pattern applied
- Added `Raises:` docstring section to both functions
- Added first unit test coverage for both functions (8 parametrized cases)

General SDK hardening (noticed while exploring the codebase for #164).

**Checklist:**

- DCO signed (`-s` flag on commit)
- Conventional commit format (`fix(trainer):`)
- No public API changes

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -155,13 +155,26 @@ def get_trainjob_initializer_step(
 ) -> types.Step:
     """
     Get the TrainJob initializer step from the given Pod name, spec, and status.
+
+    Raises:
+        ValueError: If the pod has no container named dataset-initializer or model-initializer.
     """
 
     container = next(
-        c
-        for c in pod_spec.containers
-        if c.name in {constants.DATASET_INITIALIZER, constants.MODEL_INITIALIZER}
+        (
+            c
+            for c in pod_spec.containers
+            if c.name in {constants.DATASET_INITIALIZER, constants.MODEL_INITIALIZER}
+        ),
+        None,
     )
+    if container is None:
+        raise ValueError(
+            f"Pod '{pod_name}' has no initializer container. "
+            f"Expected a container named '{constants.DATASET_INITIALIZER}' or "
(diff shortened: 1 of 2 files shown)


kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -155,13 +155,26 @@ def get_trainjob_initializer_step(
...
     container = next(
-        c
-        for c in pod_spec.containers
-        if c.name in {constants.DATASET_INITIALIZER, constants.MODEL_INITIALIZER}
+        (
+            c
+            for c in pod_spec.containers
+            if c.name in {constants.DATASET_INITIALIZER, constants.MODEL_INITIALIZER}
+        ),
+        None,
```
andreyvelich, line 169: How would it be possible that JobSet doesn't have such containers if we define label selector here to query the initializer Pods: https://github.com/1Ayush-Petwal/sdk/blob/cbb4ebd7ed7d24d3d4a70a579cc106c387f1efb7/kubeflow/trainer/backends/kubernetes/backend.py#L687-L689

We reserve the container name for the initializers, and validate it in TrainJob webhook: https://github.com/kubeflow/trainer/blob/master/pkg/webhooks/trainingruntime_webhook.go#L89-L101

1Ayush-Petwal: You're right. I missed the label selector, this case is effectively unreachable. Would it be ok to add a comment documenting this invariant for future contributors?

andreyvelich: Yes, please close this PR, and leave comment in the issue that these containers should always exist.

PR #352, opened 2026-03-04
feat(trainer): Add PyTorch Profiler integration to CustomTrainer

SoumyaRaikwar:
This PR introduces the ability to easily profile PyTorch-based [CustomTrainer](cci:2://file:///home/soumya/open-source/sdk/kubeflow/trainer/types/types.py:26:0-64:44) jobs using the official PyTorch Profiler. This significantly improves GPU observability and performance tuning capabilities for users running AI workloads with the Kubeflow SDK.

Changes include:
- Added `enable_profiler` (bool) and `profiler_dir` (string) configurations to the [CustomTrainer](cci:2://file:///home/soumya/open-source/sdk/kubeflow/trainer/types/types.py:26:0-64:44) configuration.
- Updated [get_command_using_train_func](cci:1://file:///home/soumya/open-source/sdk/kubeflow/trainer/backends/kubernetes/utils.py:318:0-409:18) in `kubernetes` and `localprocess` backends to automatically wrap the user's training function call with `torch.profiler.profile` when enabled.
- By default, traces are saved to `/artifacts/profile`, but this can be overridden via the `profiler_dir` parameter.
- Added corresponding unit tests to verify the injected profiler code snippet structure.

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -322,6 +322,8 @@ def get_command_using_train_func(
     train_func_parameters: dict[str, Any] | None,
     pip_index_urls: list[str],
     packages_to_install: list[str] | None,
+    enable_profiler: bool = False,
+    profiler_dir: str = "/artifacts/profile",
 ) -> list[str]:
     """
     Get the Trainer container command from the given training function and parameters.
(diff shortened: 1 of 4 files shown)


SoumyaRaikwar:
@astefanutti @kramaranya @szaher , PTAL whenever you have chance, Thanks!

astefanutti (commented):
@SoumyaRaikwar thanks, that's very useful!

I wonder if we should consider adding this on top of #308 and the new `TorchTrainer` as `CustomTrainer` doesn't guarantee it's a PyTorch runtime, WDYT?

cc @andreyvelich @szaher @kramaranya

HarshPopat23 (commented):
I reviewed this again against the current `main`. The PyTorch profiling use case is still valuable, but I don’t think this PR is ready to merge in its current form.

The main architectural concern remains unresolved: `CustomTrainer` is framework-neutral, and #308 has not yet introduced `TorchTrainer` into `main`. Adding PyTorch-specific fields to `CustomTrainer` would expose them for TensorFlow, MPI, and custom runtimes where `torch` may not exist.

I also noticed a few additional blockers:

1. `/artifacts/profile` is not connected to a documented persistent volume or artifact-retrieval mechanism and may be unwritable for local execution.
2. `profiler_dir` is interpolated directly into generated Python code and needs safe serialization.
3. The local backend changes `func({...})` to `func(**{...})`, which is an unrelated potentially breaking behavior change.
4.Profiling the entire training function with shape, memory, and stack collection enabled can create significant overhead and very large traces.
5.The tests only verify generated strings; execution-level and trace-output tests are needed.

I suggest moving this PR back to draft or superseding it, then rebasing the feature onto the approved specialized-trainer design with a typed profiler configuration, documented artifact lifecycle, and runtime tests.

PR #363, opened 2026-03-06
fix(trainer): replace os.chdir() with cwd= in subprocess.Popen to fix race condition

### LINKED ISSUE #356: Potential Race condition in LocalJob Executions
In the `LocalJob` class, which inherits from `threading.Thread`, the `run()` method uses `os.chdir()` to change the working directory before launching a background process, and then changes it back in a `finally` block
```
 os.chdir(self.execution_dir)

  self._process = subprocess.Popen(
      self.command,
      stdout=subprocess.PIPE,
      stderr=subprocess.STDOUT,
      text=True,
      encoding="utf-8",
      bufsize=1,
      env=self.env,
  )
```
https://github.com/kubeflow/sdk/blob/57ac0138eefa2cac18729fa9da611bfbd8fc00ed/kubeflow/trainer/backends/localprocess/job.py#L76C1-L87C14

If multiple `LocalJob `threads run concurrently, they will mix up each other's working directories, leading to chaotic path-resolution failures.

So instead of changing the path using the `os.chdir()`, can't we provide the path as an argument for the subprocess creation time? since this prevents changing the main process's path.

```
 self._process = subprocess.Popen(
                        self.command,
                        cwd=self.execution_dir, 
                        ......
)
```

MansiSingh17:
Description:
Summary
Fixes a race condition in LocalJob where concurrent threads could corrupt each other's working directory.
Problem
LocalJob inherits from threading.Thread. The run() method called os.chdir(self.execution_dir) before spawning the subprocess, then restored it in a finally block. Since os.chdir() mutates the working directory of the entire process (not just the thread), multiple concurrent LocalJob threads would overwrite each other's directory mid-execution, causing unpredictable path-resolution failures.
Fix
Removed os.chdir() and the finally cleanup entirely. Passed cwd=self.execution_dir directly to subprocess.Popen(), which scopes the directory change to the subprocess only — exactly as intended.
Changes

kubeflow/trainer/backends/localprocess/job.py — removed os.chdir() + finally block, added cwd= to Popen
kubeflow/trainer/backends/localprocess/backend_test.py — added test_concurrent_localjobs_do_not_change_cwd which spawns 5 concurrent jobs in different directories and verifies the parent process cwd is never mutated

Testing
24 passed in 0.85s
Fixes #356

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -563,10 +563,11 @@ def test_get_job_status(local_backend, test_case):
 
 def test_concurrent_localjobs_do_not_change_cwd():
     """Concurrent LocalJob threads must not mutate the parent process cwd."""
+    import os
     import threading
+
     from kubeflow.trainer.backends.localprocess.job import LocalJob
 
-    import os
     original_cwd = os.getcwd()
     errors = []
 
@@ -582,6 +583,7 @@ def test_concurrent_localjobs_do_not_change_cwd():
             errors.append(f"cwd changed to {os.getcwd()}")
 
     import tempfile
+
     dirs = [tempfile.mkdtemp() for _ in range(5)]
     threads = [threading.Thread(target=run_job, args=(d,)) for d in dirs]
     for t in threads:
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -559,3 +559,35 @@ def test_get_job_status(local_backend, test_case):
 
     status = local_backend._LocalProcessBackend__get_job_status(job)
     assert status == test_case.expected_output
+
+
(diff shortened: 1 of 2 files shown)


andreyvelich:
@MansiSingh17 Could you sign DCO please?
/assign @Fiona-Waters @szaher

kubeflow/trainer/backends/localprocess/backend_test.py
```diff
@@ -559,3 +559,37 @@ def test_get_job_status(local_backend, test_case):
 
     status = local_backend._LocalProcessBackend__get_job_status(job)
     assert status == test_case.expected_output
+
+
+def test_concurrent_localjobs_do_not_change_cwd():
```
andreyvelich, line 564: Please refactor this by using pytest + test_run_job(), like here: https://github.com/MansiSingh17/sdk/blob/3f2949cc1224c8d1c4b8ede6424ede54d91171ce/kubeflow/trainer/backends/localprocess/backend_test.py#L502

PR #368, opened 2026-03-07
feat: add update_trainjob_status utility for progress reporting

### LINKED ISSUE #367: Add utility for TrainJob progress reporting

Add a utility function to the Kubeflow SDK that allows training scripts to report progress and metrics to the Kubeflow Trainer controller. This enables the `trainerStatus` field introduced in [kubeflow/trainer#3227](https://github.com/kubeflow/trainer/pull/3227) and specified in [KEP-2779](https://github.com/kubeflow/trainer/issues/2779).

### Context

The Kubeflow Trainer controller ([PR #3227](https://github.com/kubeflow/trainer/pull/3227)) adds a Progress Plugin that:
1. Injects environment variables into training pods (`KUBEFLOW_TRAINER_STATUS_URL`, `KUBEFLOW_TRAINER_STATUS_TOKEN`, `KUBEFLOW_TRAINER_STATUS_CA_CERT`)
2. Runs a Status Server that accepts progress updates via HTTPS POST
3. Updates `TrainJob.status.trainerStatus` with progress, ETA, and metrics

The SDK needs a simple, safe utility function that training scripts can call to report status.

**Note:** A `KubeflowCallback` implementation has been submitted to HuggingFace Transformers ([PR #44487](https://github.com/huggingface/transformers/pull/44487)) which depends on this SDK utility. The callback auto-activates when running in Kubeflow and can call utility `update_runtime_status()` to report progress.

### Why is this needed?

**Problem:** AI practitioners running training jobs on Kubernetes have no native way to monitor training progress. They must either:
1. Parse container logs manually
2. Set up external tracking systems (MLflow/W&B) which adds infrastructure overhead
3. Wait blindly for jobs to complete

**Solution:** The Kubeflow Trainer controller ([PR #3227](https://github.com/kubeflow/trainer/pull/3227)) adds a Status Server that accepts progress updates from training pods. The SDK needs a client-side utility to POST updates to this server.

**User experience improvement:**

Before (no visibility):
```bash
kubectl get trainjob my-job
# NAME     STATUS    AGE
# my-job   Running   2h   ← Is it 10% done? 90% done? No idea.
```

After (with progress tracking):
```bash
kubectl get trainjob my-job -o jsonpath='{.status.trainerStatus}'
# {"progressPercentage": 67, "estimatedRemainingSeconds": 3600, "metrics": [...]}
```

**Dependency:** A `KubeflowCallback` has been submitted to HuggingFace Transformers ([PR #44487](https://github.com/huggingface/transformers/pull/44487) - In Review) which depends on this SDK utility. The callback auto-activates when running in Kubeflow and can call SDK's utility `update_runtime_status()` to report progress.

## Proposed API

```python
from kubeflow.trainer.progress import update_runtime_status

# Basic usage - SDK handles throttling (max 1 update/5s)
update_runtime_status(
    progress_percent=50,
    estimated_time_remaining=120,  # seconds or timedelta
    metrics={"loss": "0.234", "eval_accuracy": "0.89"}
)

# Force update (bypass throttling) - use for start/end
update_runtime_status(progress_percent=0, force=True)   # Training started
update_runtime_status(progress_percent=100, force=True) # Training complete
```

### Love this feature?

abhijeet-dhumal:
Adds update_runtime_status() utility for training scripts to report progress to the Kubeflow Trainer controller. This enables real-time visibility into training progress via standard Kubernetes APIs.

- Progress reporting: Report completion percentage (0-100%) and ETA
- Metrics tracking: Push training metrics (loss, accuracy, etc.) to TrainJob status
- Automatic throttling: Max 1 update per 5 seconds to avoid overwhelming the controller
- Token caching: 5-minute cache to minimize file I/O for service account tokens
- Silent failures: Network issues won't interrupt training - errors are logged but never raised

### Usage

```
from kubeflow.trainer import update_trainjob_status

update_trainjob_status(progress_percent=0, force=True)

for step in range(total_steps):
    loss = train_step(batch)
    update_trainjob_status(
        progress_percent=int((step / total_steps) * 100),
        estimated_remaining_seconds=eta_seconds,
        metrics={"loss": loss, "step": step}
    )  # SDK handles throttling

update_trainjob_status(progress_percent=100, force=True)
```

When the TrainJobStatus feature gate is enabled, the controller injects:

- KUBEFLOW_TRAINER_SERVER_URL - HTTPS endpoint for status updates
- KUBEFLOW_TRAINER_SERVER_CA_CERT - CA certificate path for TLS verification
- KUBEFLOW_TRAINER_SERVER_TOKEN - Service account token path for a

Note: HuggingFace Transformers users don't need this SDK utility - the built-in KubeflowCallback handles progress reporting automatically.
(huggingface/transformers#44487 - merged)

Fixes #367

**Checklist:**


robert-bell:
> Waiting for Kubeflow Trainer feature implementation to be merged : [kubeflow/trainer#3227](https://github.com/kubeflow/trainer/pull/3227) ...

@abhijeet-dhumal what's the status of this PR? Are we able to move it forward now the KFT implementation is merged?

abhijeet-dhumal:
@robert-bell @andreyvelich Apologies for the delayed follow-up! The PR is now ready for review.
The dependent controller implementation ([kubeflow/trainer#3227](https://github.com/kubeflow/trainer/pull/3227)) and transformers side utility (https://github.com/huggingface/transformers/pull/44487) is merged, so there are no remaining blockers.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+from __future__ import annotations
+
+from datetime import datetime, timedelta, timezone
+import logging
+import os
+import threading
+import time
+from typing import TYPE_CHECKING
+
+import requests
```
andreyvelich: `requests` isn't a direct dependency in `pyproject.toml` — it's only pulled in transitively via `kubernetes`. Per CLAUDE.md, new deps must be declared. Please add `requests` to `[project.dependencies]` so this public import doesn't break if the transitive chain changes.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utilities for Kubeflow Trainer.
+
+This module provides utility functions for training scripts running inside
+Kubeflow TrainJobs, including `update_runtime_status()` for reporting progress.
+
+Environment variables (injected by controller when TrainJobStatus feature gate is enabled):
```
andreyvelich: Feature gate name mismatch: docstring/docs use `TrainJobStatus` (matches the KEP), but the PR description says `TrainJobRuntimeStatus`. Please reconcile so users enable the right gate — likely the PR description is the outlier.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    try:
+        url = os.environ.get(_ENV_SERVER_URL)
+        if not url:
+            return False
+
+        with _throttle_lock:
+            if not force and _should_throttle():
+                return False
+            _update_last_time()
```
andreyvelich: `_update_last_time()` runs before the token check and the HTTP POST, so a failed / no-token attempt still consumes the 5s window and blocks the next legitimate retry. Consider advancing the timestamp only after a successful send (status 200), or document that throttling counts attempts rather than successes.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+
+        if not token_path or not os.path.exists(token_path):
+            return None
+
+        try:
+            with open(token_path) as f:
+                _cached_token = f.read().strip()
+                _token_read_time = now
+                return _cached_token
+        except OSError:
```
andreyvelich: An existing-but-unreadable token returns `None` with no log. Since the URL was set, the user *intended* reporting — add at least a debug/warning log here so a broken token mount is diagnosable.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+
+    Includes automatic throttling (max 1 update per 5 seconds) to avoid
+    overwhelming the controller. Use force=True to bypass throttling for
+    critical updates like training start (0%) and end (100%).
+
+    Args:
+        progress_percent: Training completion percentage (0-100).
+        estimated_time_remaining: ETA in seconds or as a timedelta.
+        metrics: Dict of metric name -> value. Values are converted to strings.
+            Truncated to 256 entries if exceeded (controller enforced limit).
```
andreyvelich: `256` is an SDK-side constant, so "(controller enforced limit)" states an unverified relationship. Either cite the KEP/CRD source, or soften to something like "Truncated to 256 entries by the SDK to avoid oversized payloads." (Same wording in the log message on line 180.)

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+            headers=headers,
+            verify=verify,
+            timeout=5,
+        )
+
+        if response.status_code == 200:
+            logger.debug("Status update sent: %s%%", progress_percent)
+            return True
+        else:
+            logger.warning("Status update failed: %s %s", response.status_code, response.text)
```
andreyvelich: Logging the full `response.text` at WARNING can leak server detail and bloat training logs (especially with `force=True` on a flaky network). Suggest `logger.warning("Status update failed: HTTP %s", response.status_code)` and move a truncated body to DEBUG.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
...
+            return False
+
+    except Exception as e:
```
andreyvelich: This catch wraps the whole body, so caller bugs (e.g. `progress_percent="50"`, a non-dict `metrics`) surface as the same `"Failed to send status update"` warning as a network error — misleading when debugging. Suggest catching `requests.RequestException` for transport and using a distinct message for other exceptions. Keeping the never-raise contract is correct.

docs/source/train/custom-training.rst
```diff
@@ -118,6 +118,42 @@ Tips and Best Practices
...
+   import time
+   from kubeflow.trainer.utils import update_runtime_status
+
+   def train():
+       total_steps = 1000
+       start_time = time.time()
+       update_runtime_status(progress_percent=0, force=True)  # Training start
+
+       for step in range(total_steps):
+           # ... training ...
```
andreyvelich: The loop body is just `# ... training ...`, but line 144 uses `metrics={"loss": loss, ...}` → `NameError` on copy-paste. Please assign it, e.g. `loss = train_step(batch)`. (The module docstring example in `utils.py` does this correctly.)

kubeflow/trainer/utils_test.py
```diff
@@ -0,0 +1,364 @@
...
+import tempfile
+from unittest.mock import MagicMock, patch
+
+from kubeflow.trainer import utils
+from kubeflow.trainer.utils import update_runtime_status
+
+
+class TestUpdateRuntimeStatus:
+    """Tests for update_runtime_status function."""
+
```
andreyvelich: Solid happy-path coverage. Please also add tests for: (a) token cache hit + TTL expiry re-read (`_get_cached_token` is untested); (b) the `OSError` token-read branch (utils.py:108-109); (c) a throttled call makes **no** HTTP POST (`post.assert_called_once()`); (d) negative ETA clamps to 0; (e) `metrics=None` / `{}` omits the `metrics` key; (f) assert the `Authorization: Bearer` header and request URL.

kubeflow/trainer/utils_test.py
```diff
@@ -0,0 +1,364 @@
...
+            f.write(content)
+            return f.name
+
+    def _mock_env(self, token_path: str) -> dict:
+        return {
+            "KUBEFLOW_TRAINER_SERVER_URL": "https://test",
+            "KUBEFLOW_TRAINER_SERVER_TOKEN": token_path,
+        }
+
+    def test_returns_false_when_not_in_kubeflow(self):
```
andreyvelich: CLAUDE.md prefers the parametrized `TestCase` dataclass pattern (see `backends/kubernetes/backend_test.py`). The env-dict + mock-session boilerplate is duplicated ~10×; consolidating value-variation cases (ETA int/timedelta/negative, progress clamp bounds) would cut a lot of duplication.

andreyvelich (commented):
Thanks for the work on this, @abhijeet-dhumal! 🙌

This review was performed with the help of AI tooling. I've left inline comments with specific suggestions and action items — please take a look. Happy to discuss any of them.

andreyvelich:
@abhijeet-dhumal Did you get a chance to review it, so we can move it forward?

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,216 @@
+# Copyright 2024 The Kubeflow Authors.
```
andreyvelich: Remove year too.

abhijeet-dhumal:
@andreyvelich Thanks for the detailed review! All feedback has been addressed, Please review !

andreyvelich:
@abhijeet-dhumal Can you rebase your PR too?

kubeflow/trainer/utils.py
andreyvelich: @abhijeet-dhumal Did you address this?

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,221 @@
...
+    now = time.monotonic()
+    return (now - _last_update_time) < _MIN_UPDATE_INTERVAL_SECONDS
+
+
+def _update_last_time() -> None:
+    global _last_update_time
+    _last_update_time = time.monotonic()
+
+
+def update_runtime_status(
```
andreyvelich: Can we move this util function, to the https://github.com/abhijeet-dhumal/sdk/blob/4ef89a354ec56389a9f2e6fa2fecb0b083056dc9/kubeflow/trainer/backends/kubernetes/utils.py?

And then, we can export this as part of __init__.py: https://github.com/abhijeet-dhumal/sdk/blob/4ef89a354ec56389a9f2e6fa2fecb0b083056dc9/kubeflow/trainer/__init__.py
So users can do: `from kubeflow.trainer import update_trainjob_status`

andreyvelich: @robert-bell @abhijeet-dhumal Any thoughts on renaming this to: `update_trainjob_status()`

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,221 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utilities for Kubeflow Trainer.
+
+This module provides utility functions for training scripts running inside
```
andreyvelich: You can move this doc string directly under your `update_trainjob_status()` function.

kubeflow/trainer/utils.py
andreyvelich: @abhijeet-dhumal Any thoughts on this?

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,221 @@
...
+def _get_cached_token(token_path: str | None) -> str | None:
+    global _cached_token, _token_read_time
+
+    now = time.monotonic()
+    with _token_lock:
+        if _cached_token and (now - _token_read_time) < _TOKEN_CACHE_TTL_SECONDS:
+            return _cached_token
+
+        if not token_path or not os.path.exists(token_path):
+            return None
```
andreyvelich: @robert-bell @abhijeet-dhumal Any particular reason to do Token Caching here?
Why not just always read token from the file system?

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _update_last_time() -> None:
+    global _last_update_time
+    _last_update_time = time.monotonic()
+
+
+def update_runtime_status(
+    progress_percent: int | None = None,
+    estimated_time_remaining: int | timedelta | None = None,
```
andreyvelich: Can we keep name consistent?

```suggestion
    estimated_remaining_seconds: int | timedelta | None = None,
```

andreyvelich: Do we need to accept timedelta?

kubeflow/trainer/utils_test.py
andreyvelich: @abhijeet-dhumal Did you address this?
We might need to refactor the rest of tests with TestCases.

Also, I suggest that we just have one single function to tests everything:
```
test_update_trainjob_status
```

andreyvelich (commented):
We might also need to add E2E tests for this feature.
After we merge this, can we add new example to `kubeflow/trainer` @abhijeet-dhumal @robert-bell?
cc @Sridhar1030 @h0pers

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,221 @@
...
+    metrics: dict[str, float | int | str] | None = None,
+    force: bool = False,
```
andreyvelich: What is the use-case to bypass throttling?
Shall we not allow users to update TrainJob status in more than `x` seconds?
cc @robert-bell

kubeflow/trainer/utils.py
abhijeet-dhumal: thinking since this moved to backends/kubernetes/utils.py, requests is already guaranteed via the kubernetes package which is direct dep.. right? 
The function can't be reached without kubernetes installed.

abhijeet-dhumal: Added requests>=2.25.0 to project.dependencies. maybe can be treated as cheap insurance regardless of where the function lives.

kubeflow/trainer/utils.py
abhijeet-dhumal: Fixed ✅

kubeflow/trainer/utils.py
abhijeet-dhumal: Addressed.. _update_last_time() now only runs after a 200. 
Failed attempts don't block the next retry.. also added tests for the same !

kubeflow/trainer/utils.py
abhijeet-dhumal: Added logger.warning("Unable to read token from %s: %s", ...) on OSError, and logger.debug("No authentication token available") when token is None after URL check passes

kubeflow/trainer/utils.py
abhijeet-dhumal: Softened to "Truncated to 256 entries by the SDK to avoid oversized payloads" in the docstring and "(SDK limit)" in the warning log.

kubeflow/trainer/utils.py
abhijeet-dhumal: Done.. status code at WARNING, truncated body (%.500s) moved to DEBUG.

kubeflow/trainer/utils.py
abhijeet-dhumal: Split into requests.RequestException ("Failed to send status update") and general Exception ("Unexpected error in update_trainjob_status"), different messages, same never-raise contract.

docs/source/train/custom-training.rst
abhijeet-dhumal: Fixed.. doc example now includes loss = train_step(batch) in the loop body.

kubeflow/trainer/utils_test.py
abhijeet-dhumal: All covered: TestTokenCaching (cache hit, TTL expiry, OSError), TestThrottling (no POST on throttled call, force bypass, failed send doesn't block), TestPayload parametrized (negative ETA = 0, metrics=None/{} omits key), TestAuthAndHeaders (Bearer header, request URL, CA cert verify).

kubeflow/trainer/utils_test.py
abhijeet-dhumal: Yup.. Refactored payload cases into a StatusTestCase dataclass with @pytest.mark.parametrize and kept separate test classes for throttling, token caching, and auth since they test distinct module behaviors (not just payload field variations) and have different fixture requirements.

abhijeet-dhumal: Happy to consolidate further if you'd prefer.

kubeflow/trainer/utils.py
abhijeet-dhumal: Done ✅

abhijeet-dhumal: Done. Public interface is from kubeflow.trainer import update_trainjob_status. A backward-compat shim in kubeflow/trainer/utils.py re-exports as update_runtime_status for any early adopters.

abhijeet-dhumal: Happy to consolidate further if you'd prefer.

kubeflow/trainer/utils.py
abhijeet-dhumal: The function gets called every training step , potentially hundreds/second in hot loops. Without caching, that's one open+read+close per step even though the projected SA token only rotates every 1 hour. The 5 minute TTL keeps file I/O near-zero while still picking up rotated tokens promptly.

That said, the throttle already limits actual HTTP calls to 1 per 5s, so we'd only hit the file 12 times/minute in practice. I can drop the cache if you'd prefer simplicity.. cost is one syscall per non-throttled call.

abhijeet-dhumal: @andreyvelich @robert-bell  let me know.

kubeflow/trainer/utils.py
abhijeet-dhumal: Yeah we can drop it 👍

kubeflow/trainer/utils.py
abhijeet-dhumal: guaranteed delivery of training start (0%) and end (100%)... without force, if a regular update was sent 2s before completion, the final 100% update gets silently dropped by the throttle.

abhijeet-dhumal: The HuggingFace KubeflowCallback uses the same pattern .. force=True on on_train_begin and on_train_end.

abhijeet-dhumal: but there is one more alternative .. by hardcoding bypass for 0%/100% only, or make it _force (internal). But explicit force is more flexible for custom callbacks without adding complexity.

abhijeet-dhumal: @andreyvelich @robert-bell thoughts?

andreyvelich: For the first implementation can we create this API without `force` option? If we get feedback from users that they want to bypass it, we can simple extend the parameter.

kubeflow/trainer/utils.py
andreyvelich: I see, I am fine with keeping caching for now, then.

kubeflow/trainer/utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+Prefer importing directly: ``from kubeflow.trainer import update_trainjob_status``
+"""
+
+from kubeflow.trainer.backends.kubernetes.utils import update_trainjob_status
+
+# Legacy alias
+update_runtime_status = update_trainjob_status
+
+__all__ = ["update_runtime_status", "update_trainjob_status"]
```
andreyvelich: I don't think we need this backward-compatible script.
We've never released this API to public, isn't @abhijeet-dhumal @robert-bell ?

```suggestion
```

kubeflow/trainer/utils_test.py
andreyvelich: Let's follow the rules defined here: https://github.com/kubeflow/sdk/blob/main/AGENTS.md#3-testing-requirements

Just keep one test function with multiple test cases, like we do in other scripts.

```suggestion
    def test_update_trainjob_status(test_case):
```

kubeflow/trainer/utils_test.py
```diff
@@ -0,0 +1,340 @@
+# Copyright The Kubeflow Authors.
```
andreyvelich: Please move the `test_update_trainjob_status()` to this file: https://github.com/abhijeet-dhumal/sdk/blob/ef9b6ad2e564d1b30f2ca0860579c049a5be5932/kubeflow/trainer/backends/kubernetes/utils_test.py#L78

pyproject.toml
```diff
@@ -28,6 +28,7 @@ classifiers = [
 dependencies = [
   "kubernetes>=27.2.0,<36.0.0",  # 36.0.0 has multiple regressions; see kubernetes-client/python#2592, #2596
   "pydantic>=2.10.0",
+  "requests>=2.25.0",
```
andreyvelich, line 31: you might need to run `uv lock` to update deps
cc @Fiona-Waters

kubeflow/trainer/utils.py
abhijeet-dhumal: ack

andreyvelich (commented):
Thanks for the updates @abhijeet-dhumal!
I think, we can move this forward to collect initial feedback from users.
/lgtm
cc @robert-bell @kramaranya @Fiona-Waters

PR #381, opened 2026-03-11
feat(trainer): replace PodTemplateOverrides with RuntimePatches API

Fiona-Waters:
 Refactors the Trainer SDK options to align with the upstream RuntimePatches API
  proposed in kubeflow/trainer#3199.

  - Replaces `PodTemplateOverrides`, `PodTemplateOverride`, `PodSpecOverride`, and
    `ContainerOverride` with the new `RuntimePatches` hierarchy (`RuntimePatch`,
    `TrainingRuntimeSpecPatch`, `JobSetTemplatePatch`, `ReplicatedJobPatch`,
    `PodSpecPatch`, `ContainerPatch`, etc.)
  - Removes `SpecLabels` and `SpecAnnotations` options (the upstream API removes
    `spec.labels` and `spec.annotations` from `TrainJobSpec`; these can now be set
    via `RuntimePatches` at the JobSet or Job metadata level)
  - Keeps `Labels` and `Annotations` options unchanged (they operate on
    `.metadata`, which is unaffected by the upstream change)
  - Updates the Kubernetes backend to pass `runtimePatches` to the API

  ## Open items (pending upstream PR merge)

  - [ ] **Generated models**: `kubeflow_trainer_api` models do not yet include
    `runtimePatches`. A dict-injection workaround is used in `backend.py` until
    the models are regenerated after kubeflow/trainer#3199 merges.
  - [ ] **Metadata type restriction**: The upstream PR discussion includes a
    suggestion to restrict `metadata` fields to only `labels` and `annotations`
    (rather than full `ObjectMeta`). If the upstream API adopts a restricted
    `MetadataPatch` type, we should add corresponding validation in the SDK.
  - [ ] **Convenience helpers**: Reviewers suggested that `RuntimePatches` should
    be encapsulated behind higher-level options for common patterns (e.g. PVC
    mounting, node selectors). This can be addressed in a follow-up once the
    upstream API stabilises.
  - [ ] **Conflict validation**: The upstream PR is discussing whether to forbid
    two managers from patching the same field. If server-side validation is added,
    the SDK may want client-side warnings or validation to match.

  ## Test plan

  - [x] All existing unit tests updated and passing (337 tests)
  - [x] Lint and format checks passing
  - [x] `RuntimePatches` option tests cover: backend validation, basic patches,
    node selectors, volumes with container mounts, JobSet metadata, multiple
    managers
  - [x] Backend integration tests cover: train with runtime patches, train with
    metadata labels combined with runtime patches

Fixes #

**Checklist:**

@kubeflow/kubeflow-sdk-team @abhijeet-dhumal


.github/workflows/trigger-distributed-workloads-e2e.yaml
```diff
@@ -0,0 +1,103 @@
+name: Trigger distributed-workloads E2E
```
andreyvelich: Why do we need this Workflow?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1242,39 +1248,117 @@ def test_get_runtime_packages(kubernetes_backend, test_case):
...
+                                                        "template": {
+                                                            "spec": {
+                                                                "nodeSelector": {
+                                                                    "node-type": "gpu-a100"
+                                                                },
+                                                            },
+                                                        },
+                                                    },
+                                                },
+                                            },
```
andreyvelich: Shall we re-use the Python models here?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1242,39 +1248,117 @@ def test_get_runtime_packages(kubernetes_backend, test_case):
...
                 labels={"owner": "ml-team"},
                 annotations={"description": "Fine-tuning job"},
-                spec_labels={"app": "training", "version": "v1.0"},
-                spec_annotations={"prometheus.io/scrape": "true"},
+                runtime_patches=[
+                    models.TrainerV1alpha1RuntimePatch.from_dict(
+                        {
+                            "manager": "trainer.kubeflow.org/kubeflow-sdk",
+                            "trainingRuntimeSpec": {
+                                "template": {
```
andreyvelich: Same for other tests.

kubeflow/trainer/options/kubernetes.py
```diff
@@ -90,141 +90,188 @@ def __post_init__(self):
...
+    Args:
+        manager: Who owns this patch entry. Used to track ownership and avoid
+                 conflicts. For example, Kueue sets this to "kueue.x-k8s.io/manager".
+        training_runtime_spec: Allowed patches for ClusterTrainingRuntime or
+                               TrainingRuntime-based jobs.
+    """
 
-        metadata = job_spec.setdefault("metadata", {})
-        metadata["annotations"] = self.annotations
+    manager: str
```
andreyvelich: IIRC, we discussed that manager will be automatically set to `trainer.kubeflow.org/kubeflow-sdk`.
So I suggest that initially we don't allow users to override that parameter.
cc @kubeflow/kubeflow-trainer-team @akshaychitneni @robert-bell
```suggestion
```

kubeflow/trainer/options/kubernetes.py
```diff
@@ -279,119 +323,76 @@ def __call__(
...
 
-        spec = job_spec.setdefault("spec", {})
-        spec["annotations"] = self.annotations
+        metadata = job_spec.setdefault("metadata", {})
+        metadata["annotations"] = self.annotations
 
 
-class PodTemplateOverrides:
-    """Add pod template overrides to the TrainJob (.spec.podTemplateOverrides).
+class RuntimePatches:
```
andreyvelich: Similar to the above comment, we should not allow users to set multiple patches via Kubeflow SDK initially.

```suggestion
```

.github/workflows/trigger-distributed-workloads-e2e.yaml
Fiona-Waters: Oops, this shouldn't be here.

Issue #85, opened 2025-09-08
Add structured and configurable logging support to Kubeflow SDK


szaher:
### What you would like to be added?

I would like the Kubeflow SDK to include structured, configurable logging across its modules. Specifically:

Consistent use of Python’s logging library instead of print statements.

Support for different logging levels (DEBUG, INFO, WARNING, ERROR).

Ability for users to configure log formatting and destinations (stdout, file, JSON, etc.).

Clear and actionable log messages for key SDK operations (e.g., pipeline compilation, run submission, experiment creation).

### Why is this needed?

Currently, the Kubeflow SDK provides limited visibility into its execution flow, making it difficult to debug and trace issues during development and runtime interactions. 

Proper logging would:

- Improve developer experience by making failures and warnings more transparent.
- Help users understand the sequence of SDK operations and their outcomes.
- Enable integration with observability tools (e.g., ELK, ...etc) by supporting structured log output.
- Align the SDK with standard Python practices for enterprise-grade software, making it easier to maintain and troubleshoot.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

szaher:
This is just about adding some meaningful log messages with debug level in the sdk with null handler so I believe it's suitable for new contributors to take on. 


In the SDK 
```python
# kubeflow/__init__.py
import logging

logger = logging.getLogger(__name__)
logger.addHandler(logging.NullHandler())

def some_function():
    logger.debug("This is an informational message from the SDK.")
```

Users that integrate the SDK in their libraries

```python
import logging
from kubeflow import some_function

# The user configures logging for their application
logging.basicConfig(level=logging.DEBUG)

# Calling the SDK function will now log to the console
# The NullHandler is effectively overridden by the user's configuration
some_function()
```

/good-first-issue

szaher:
@andreyvelich @astefanutti @kubeflow/kubeflow-sdk-team WDYT?

/remove-lifecycle stale

astefanutti:
+1 this would be a very useful addition.

Thanks @szaher!

khushiwayal:
Hi @andreyvelich and @szaher! I’m a first-year BE (ECE) student and I’ve been contributing to the SDK documentation (PR #304). I noticed this issue has been inactive for a while. I’d love to take this over and implement the structured logging support as my next contribution. May I be assigned to this?

andreyvelich:
@anencore94 @MansiSingh17 Are you still working on this issue?

[...]

PR #382, opened 2026-03-12
feat(docs): add OpenTelemetry integration KEP

### LINKED ISSUE #164: Integrate Kubeflow SDK with OpenTelemetry

XploY04:

Fixes #164 

**Checklist:**

@jaiakash @kramaranya @dhanishaphadate please review


andreyvelich:
We discussed this proposal during the last Kubeflow SDK call: https://youtu.be/dW_GSUBsUTg
@Rajneesh180 @kramaranya Appreciate your help with reviewing this proposal, so we can move it forward.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+- Collecting operational metrics (job counts, durations, error rates) for monitoring dashboards
+- Exporting telemetry to any OTel-compatible backend (Jaeger, Grafana Tempo, Datadog, etc.)
+- Running with zero performance cost when telemetry is not configured (no-op by default)
+
+### Goals
+
+- Instrument `TrainerClient` and all three backends (Kubernetes, Container, LocalProcess)
+  with OTel traces.
+- Instrument `PipelinesClient` ([PR #343](https://github.com/kubeflow/sdk/pull/343),
+  pending merge) with OTel traces across all public methods.
```
kramaranya: PipelinesClient is not part of the SDK yet, shall this be a stretch goal for now?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+- Build a shared telemetry module (`kubeflow/common/telemetry/`) that all SDK clients can
+  reuse.
+- Propagate trace context across pod boundaries via W3C `TRACEPARENT` env var injection.
+- Collect operational metrics: job counts, operation durations, active jobs, errors.
+- Provide a `configure_telemetry()` convenience helper so users can set up exporters and
+  sampling without writing boilerplate `TracerProvider` setup code.
```
kramaranya: Hm why do we want configure_telemetry() in the SDK itself, instead of keeping OTel provider/exporter setup in user application code and having the SDK only emit spans?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+|   |   +-- trainer_client.py         # MODIFIED add spans to public methods
+|   +-- backends/
+|   |   +-- kubernetes/backend.py     # MODIFIED add spans to backend methods
+|   |   +-- container/backend.py      # MODIFIED add spans to backend methods
+|   |   +-- localprocess/backend.py   # MODIFIED add spans to backend methods
+|   +-- ...
+|
++-- pipelines/
+|   +-- api/
+|       +-- pipelines_client.py       # MODIFIED add spans to public methods
```
kramaranya: PipelinesClient is going to live inside Pipelines repo though https://github.com/kubeflow/pipelines/pull/13405, I wonder how would that work with OTel integration

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+
+```toml
+# pyproject.toml
+dependencies = [
+    "kubernetes>=27.2.0",
+    "pydantic>=2.10.0",
+    "kubeflow-trainer-api>=2.0.0",
+    "kubeflow-katib-api>=0.19.0",
+    "opentelemetry-api>=1.4,<2.0",                      # NEW
+    "opentelemetry-semantic-conventions>=0.50b0",        # NEW
```
kramaranya, line 260: You don't mention this dependency in the summary

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+```
+INFO:kubeflow.trainer:Creating TrainJob a3f8b2c1d9e0f in namespace default
+```
+
+become correlated with the active span:
+
+```
+INFO:kubeflow.trainer:Creating TrainJob a3f8b2c1d9e0f in namespace default
+  [trace_id=4bf92f3577b01e8d span_id=00f067aa0ba902b7]
+```
```
kramaranya: This doesn’t happen automatically in existing logs, right? Don’t users need extra logging setup for that?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,707 @@
...
+| Attribute | Source | Used When |
+|---|---|---|
+| `k8s.namespace.name` | [K8s semconv](https://opentelemetry.io/docs/specs/semconv/resource/k8s/) | Kubernetes backend operations |
+| `error.type` | General semconv | On error spans |
+
+We do not use `k8s.job.name` for TrainJob names. That OTel attribute is defined for
+Kubernetes batch/v1 Jobs, and a TrainJob is a CRD not a batch Job. TrainJob names go in
+the Kubeflow-specific `kubeflow.job.name` attribute instead.
+
+#### GenAI Semantic Conventions
```
kramaranya, line 355: Do we really need to commit to GenAI semconv in v1? Shall we start with Kubeflow-specific attributes only?

kramaranya (commented):
Thank you for this proposal, @XploY04!
I left a few questions

docs/proposals/164-opentelemetry-integration/README.md
XploY04: Yes, I think we should move it to stretch goals because this KEP will mainly be focused on trainer client.

docs/proposals/164-opentelemetry-integration/README.md
XploY04: The SDK should primarily just emit spans through `opentelemetry-api` and respect whatever provider/exporter setup the user application already has.

The proposed `configure_telemetry()` is intended only as optional convenience for quickstarts and simple scripts, not as the required configuration path. Production users should still configure OTel directly using standard OTel APIs/env vars.

docs/proposals/164-opentelemetry-integration/README.md
XploY04: If PipelinesClient lives in `kubeflow/pipelines`, then this KEP should treat Pipelines instrumentation as cross-repo follow-up work, not part of the initial SDK implementation. The SDK side can still define compatible telemetry conventions, but the actual PipelinesClient spans should be implemented in the Pipelines repo.

docs/proposals/164-opentelemetry-integration/README.md
XploY04: Correct, this needs extra user-side logging setup. The SDK emitting spans is not enough to make existing `logging` output include trace IDs.

I’ll update this section to clarify that log correlation only happens when the user configures OTel logging integration, for example by adding an OTel `LoggingHandler` or a logging formatter/filter that reads the current trace context.

The SDK should not mutate global logging configuration automatically. It can document the setup and the optional `configure_telemetry()` helper can provide a convenience path, but existing logs remain unchanged unless the user opts into log correlation.

docs/proposals/164-opentelemetry-integration/README.md
XploY04: That makes sense. For v1, we can keep the attribute model Kubeflow-specific and avoid committing to GenAI semantic conventions until we have a concrete training-specific mapping that maintainers agree with.

I’ll update the KEP to remove GenAI semconv from the initial scope. We can keep it as future work to evaluate later, especially if OTel adds more training-oriented semantic conventions.

XploY04:
@andreyvelich can we merge this ?

andreyvelich:
Sure, @kramaranya @Fiona-Waters @abhijeet-dhumal @tariq-hasan @szaher Do we have any open questions before we can move this KEP forward?
Did you review the metrics?
/ok-to-test

abhijeet-dhumal:
Thanks @XploY04, Design is well-considered 🚀 
A few questions before 'lgtm'..

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,675 @@
...
+
+```python
+from opentelemetry import trace
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor, InMemorySpanExporter
+
+
+def setup_test_telemetry() -> InMemorySpanExporter:
+    """Set up an in-memory span exporter for test assertions."""
+    exporter = InMemorySpanExporter()
```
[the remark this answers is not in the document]
abhijeet-dhumal: @XploY04 Import is correct for opentelemetry-sdk>=1.0. Copilot's concern is outdated.. can resolve.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,713 @@
...
+
+If the training code does not use OTel, the `TRACEPARENT` env var is harmlessly ignored.
+
+### Metrics
+
+| Metric Name | Type | Unit | Description |
+|---|---|---|---|
+| `kubeflow.trainer.job.created` | Counter | `{job}` | Total training jobs submitted |
+| `kubeflow.trainer.operation.duration` | Histogram | `s` | Duration of each SDK API call |
+| `kubeflow.trainer.job.active` | UpDownCounter | `{job}` | Currently running jobs |
```
abhijeet-dhumal: QQ: How does the UpDownCounter for job.active decrement if the user submits a job but never calls wait_for_job_status?

docs/proposals/164-opentelemetry-integration/README.md
abhijeet-dhumal: +1 defer GenAI semconv. Training orchestration doesn't map to inference-oriented attributes today.

docs/proposals/164-opentelemetry-integration/README.md
abhijeet-dhumal: Maybe you can mention both opentelemetry-api and opentelemetry-semantic-conventions in the Summary section.. WDYT?

docs/proposals/164-opentelemetry-integration/README.md
XploY04: Yes, I didn't think about this, if wait_for_job_status is not called, the UpDownCounter will not decrement. So, we can remove it from the scope for now.

andreyvelich:
Please help us to review the final design for OTel integration.

/cc @kramaranya @Fiona-Waters @tariq-hasan @abhijeet-dhumal

XploY04:
@andreyvelich two more metrics we could add, both client-side and low-cardinality:

- `kubeflow.trainer.job.wait.duration` — Histogram, seconds. How long `wait_for_job_status` blocked. Dimensioned by terminal `kubeflow.job.status`. This gives queue-wait visibility, which on a busy GPU cluster is usually what users actually complain about.
- `kubeflow.trainer.job.terminal` — Counter. Jobs by final status. Dimensioned by `kubeflow.job.status` and `kubeflow.backend.type`. Gives the error-rate dashboard in Story 2.

Both reuse attributes already in the KEP, so no new dimensions and no cardinality risk. Happy to add them to the metrics table if you want them in scope.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+**`opentelemetry-api` as a core dependency.** The
+[OTel native instrumentation guidelines](https://opentelemetry.io/docs/concepts/instrumentation/libraries/)
+recommend libraries use `opentelemetry-api` for native instrumentation without requiring
+the OTel SDK or exporters. The package is under 100KB compressed, with `typing-extensions`
+and `importlib-metadata` as transitive dependencies, and is no-op by default. There is no
+runtime cost when users don't configure telemetry.
+
+**`opentelemetry-semantic-conventions`.** Provides Python constants for standard attribute
+names (e.g., `K8S_NAMESPACE_NAME` instead of the raw string `"k8s.namespace.name"`).
+Prevents silent typos and provides IDE autocomplete.
```
tariq-hasan, line 279: For the sake of importing only two constants is it worth installing this package or is it better to define these constants locally?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+`kubeflow.operation.name`. When `TimeoutError` spikes for `get_job`, I know the API server
+is under pressure.
+
+#### Story 3: End-to-End Trace Across Pod Boundary
+
+As an ML engineer, I configure OTel in both my SDK script and my training function. The
+SDK injects `TRACEPARENT` into the training pod's environment. My training code reads it
+and starts a child span. In Jaeger, I see one connected trace from `client.train()` all
+the way through to `training.epoch` inside the pod.
+
```
tariq-hasan, line 213: As it stands there is a missing middle - Kueue admission, JobSet creation, node scheduling, etc. - in the "end-to-end" that the KEP does not really cover - given that server-side observability is out of scope here. After all, this KEP cannot answer a user's question about whether their workload has been preempted or if it has been sitting in the queue for longer than expected.

Should we be explicit in the non-goals section that end-to-end tracing is not quite achievable as part of this KEP - unless we are able to rethink how we do context propagation?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+| `kubeflow.job.name` | Training job identifier | `a3f8b2c1d9e0f` |
+| `kubeflow.job.status` | Current job status | `Running`, `Complete`, `Failed` |
+| `kubeflow.backend.type` | Execution backend | `kubernetes`, `container`, `local-process` |
+| `kubeflow.container.runtime` | Container runtime | `docker`, `podman` |
+| `kubeflow.trainer.type` | Trainer class | `CustomTrainer`, `BuiltinTrainer` |
+| `kubeflow.runtime.name` | Training runtime name | `torch-distributed` |
+| `kubeflow.num_nodes` | Number of training nodes | `2`, `4` |
+| `kubeflow.initializer.type` | Initializer class | `HuggingFaceModelInitializer` |
+| `kubeflow.initializer.uri` | Model/dataset URI | `hf://meta-llama/Llama-3` |
+| `kubeflow.operation.name` | SDK operation (for metrics) | `train`, `get_job` |
```
tariq-hasan, line 382: Do we want to namespace attributes over the SDK clients as well, for example `kubeflow.trainer.job.name` vs `kubeflow.spark.job.name`? As it stands these attributes are likely to collide across clients.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
+# KEP-164: Integrate Kubeflow SDK with OpenTelemetry
```
tariq-hasan, line 1: We should move the proposal to the [`proposals`](https://github.com/kubeflow/sdk/tree/main/proposals) directory and add a `kep.yaml` to match the existing convention.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+
```
tariq-hasan, line 2: Also it would be nice to have a header and table of contents as per the [most recent](https://github.com/kubeflow/sdk/tree/main/proposals/125-kfp-client) KEP to standardize on the overall structure.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+## Design Details
+
+### Architecture
+
+```
+kubeflow/
++-- common/
+|   +-- telemetry/                    # NEW shared OTel infrastructure
+|   |   +-- __init__.py               # configure_telemetry() and module-level tracer/meter setup
```
tariq-hasan, line 228: Is `configure_telemetry()` meant to go in `__init__.py` or in `configure.py`?

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+(`Service.Operation`):
+
+```
+TrainerClient.train
+TrainerClient.get_job
+TrainerClient.wait_for_job_status
+KubernetesBackend.train
+ContainerBackend.train
+LocalProcessBackend.train
+```
```
tariq-hasan: I believe that in the May community call there was a consensus to remove the backend-specific spans and keep the `TrainerClient` spans for now.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+
+Make OTel an optional extra (`pip install kubeflow[telemetry]`) with `try/except` import
+guards.
+
+Rejected because every instrumentation call would need an `if _tracer:` guard, duplicating
+control flow throughout the codebase. The OTel API is under 100KB compressed, with only `typing-extensions` and
+`importlib-metadata` as transitive dependencies, and is no-op by default. The SDK already
+depends on `kubernetes` which is much heavier. The OTel spec explicitly says libraries
+should depend on the API.
+
```
tariq-hasan, line 682: The argument that every instrumentation call would need an `if _tracer:` guard may not really hold if we decide we want to create one internal tracer module as follows:

```
# kubeflow/common/telemetry/_tracer.py

try:
    from opentelemetry import trace
    tracer = trace.get_tracer("kubeflow")
except ImportError:
    tracer = NoOpTracer()
```

Then the rest of the code in the individual client packages can remain as is unaffected and won't need to care if OTel is installed or not despite being wired in with OTel.

```
from kubeflow.common.telemetry._tracer import tracer

with tracer.start_as_current_span("TrainerClient.train"):
    ...
```

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+| Attribute | Description | Example Values |
+|---|---|---|
+| `kubeflow.job.name` | Training job identifier | `a3f8b2c1d9e0f` |
+| `kubeflow.job.status` | Current job status | `Running`, `Complete`, `Failed` |
+| `kubeflow.backend.type` | Execution backend | `kubernetes`, `container`, `local-process` |
+| `kubeflow.container.runtime` | Container runtime | `docker`, `podman` |
+| `kubeflow.trainer.type` | Trainer class | `CustomTrainer`, `BuiltinTrainer` |
+| `kubeflow.runtime.name` | Training runtime name | `torch-distributed` |
+| `kubeflow.num_nodes` | Number of training nodes | `2`, `4` |
```
tariq-hasan, line 379: I think replacing `kubeflow.num_nodes` with `kubeflow.node.count` aligns the attribute naming with the OTel conventions as that creates a naming hierarchy that makes it easy to categorize attributes. This is more so about naming semantics and its consistency thereof for all proposed attributes.

Moving one step further we might want to call this `kubeflow.trainer.node.count` to align with the other comment regarding namespacing attributes.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+    carrier: dict[str, str] = {}
+    TraceContextTextMapPropagator().inject(carrier, context.get_current())
+    return {k.upper(): v for k, v in carrier.items()}
+```
+
+Injection points:
+
+| Backend | Where TRACEPARENT is injected |
+|---|---|
+| Kubernetes | Pod template env vars in TrainJob CR spec |
```
tariq-hasan, line 409: As it stands the `TrainJob` CRD does not directly expose a pod template field where environment variables can be injected. There are three primary ways to inject environment variables into the training runtime on the control plane and all three of them are relevant here: [`spec.trainer.env`](https://github.com/kubeflow/trainer/blob/master/pkg/apis/trainer/v1alpha1/trainjob_types.go#L267-L273), [`spec.initializer.dataset.env`](https://github.com/kubeflow/trainer/blob/master/pkg/apis/trainer/v1alpha1/trainjob_types.go#L204-L210) and [`spec.initializer.model.env`](https://github.com/kubeflow/trainer/blob/master/pkg/apis/trainer/v1alpha1/trainjob_types.go#L230-L236). This is especially so because dataset or model download latencies are relevant observability data points that would form part of the overall Trainer observability story.

Secondly, this injection mechanism would also require a clearer specification of the propagation contract - for example, if TRACEPARENT and TRACESTATE env variables are pre-existing do we overwrite, skip, or error, how do we handle the 128-variable limit, etc.

The specification of the contract extends to the client-side as well. For example, what should be the precedence rule when a user decides to inject the trace variables using [`CustomTrainer.env`](https://github.com/XploY04/sdk/blob/6215abc89a260fef466f37644a72de3bce7d4017/kubeflow/trainer/types/types.py#L62) - should the SDK win, should the client win, or should an error be thrown, etc.

tariq-hasan, line 409: If anything the cleanest approach for context propagation may not be to use direct injection into the pods or to use https://github.com/kubeflow/sdk/pull/447.

It may be to combine both approaches by having the SDK write the `traceparent` annotation on the custom resource and for the controller to read the annotation and materialize it in every container.

This would apply equally well for both Trainer and Spark operator.

If I am not wrong here there is already a direct precedent for this approach in terms of what the status server already accomplishes in Kubeflow Trainer.

Wiring context propagation in this way would also enable the "end to end" framing in story 3 to be achieved as the controller would be able to use the annotations to emit its own spans.

Would appreciate further thoughts and suggestions on this approach. @andreyvelich @kramaranya @abhijeet-dhumal @nabuskey

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+```python
+# kubeflow/common/telemetry/__init__.py
+
+from opentelemetry import trace, metrics
+
+def configure_telemetry(
+    exporter: str = "otlp",
+    endpoint: str | None = None,
+    sampling_rate: float = 1.0,
+    service_name: str = "kubeflow-sdk",
```
tariq-hasan, line 488: I am thinking that the `service_name` should not be `kubeflow-sdk` but rather the name of the application itself. A single app or service could very well have multiple libraries running within it - kubeflow sdk being just one of them. 

That said there would also be a need to identify the scope of the sdk instrumentation within the app and `trace.get_tracer("kubeflow.trainer", __version__)` would very well satisfy that requirement.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+    Raises ImportError with install instructions if they are missing.
+
+    Args:
+        exporter: Exporter type. "otlp", "console", or "none".
+        endpoint: Collector endpoint. Defaults to OTEL_EXPORTER_OTLP_ENDPOINT or localhost:4317.
+        sampling_rate: Fraction of traces to sample, 0.0 to 1.0.
+        service_name: Value for the service.name resource attribute.
+    """
+    ...
+```
```
tariq-hasan, line 502: The KEP already writes explicitly in the proposal section that "user-owned OTel configuration is the primary path." In light of this what would be the policy around precedence rules in the case when the app already has pre-configured tracer and meter providers and in the case when it does not? Essentially we'd want to avoid the double-configuration failure mode by defining a clear contract around what happens if a provider already exists.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+|---|---|---|
+| `kubeflow.job.name` | Training job identifier | `a3f8b2c1d9e0f` |
+| `kubeflow.job.status` | Current job status | `Running`, `Complete`, `Failed` |
+| `kubeflow.backend.type` | Execution backend | `kubernetes`, `container`, `local-process` |
+| `kubeflow.container.runtime` | Container runtime | `docker`, `podman` |
+| `kubeflow.trainer.type` | Trainer class | `CustomTrainer`, `BuiltinTrainer` |
+| `kubeflow.runtime.name` | Training runtime name | `torch-distributed` |
+| `kubeflow.num_nodes` | Number of training nodes | `2`, `4` |
+| `kubeflow.initializer.type` | Initializer class | `HuggingFaceModelInitializer` |
+| `kubeflow.initializer.uri` | Model/dataset URI | `hf://meta-llama/Llama-3` |
```
tariq-hasan, line 381: Given we are declaring `kubeflow.initializer.uri` how would we handle security or privacy concerns regarding this attribute? For example, do we record the URI as-is, remove its query parameters, redact any credentials in the URL, hash its value, record only the scheme, allow users to disable the attribute, etc.? Same goes for `k8s.namespace.name`.

More generally speaking, what should be the policy around security and privacy when it comes to all of these attributes? I'm thinking it may be ideal to have rules for when an attribute is recorded, redacted/sanitized, never recorded, not automatically copied, etc.?

The security policy might need a dedicated section in the KEP since enterprise use cases are a primary target.

tariq-hasan: Other than the security and privacy concern a discussion for how the SDK would handle cases where telemetry is explicitly disabled (for example, with `OTEL_SDK_DISABLED=true`) and the case of Kubeflow usage in an air-gapped cluster may be good to note as well.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+```
+INFO:kubeflow.trainer:Creating TrainJob a3f8b2c1d9e0f in namespace default
+  [trace_id=4bf92f3577b01e8d span_id=00f067aa0ba902b7]
+```
+
+This lets users query logs by trace ID in tools like Grafana Loki and see every log line
+that happened during a specific `train()` call. The SDK should not mutate global logging
+configuration automatically. The `configure_telemetry()` helper can optionally set up log
+correlation when users ask for that convenience path.
+
```
tariq-hasan, line 476: For log correlation would it be better to coordinate with the effort underway in https://github.com/kubeflow/sdk/issues/85 to enable structured logging in the SDK?

The KEP as written does not make any assumptions about the structured logging stream of work.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+traces, I can see that `KubernetesBackend.train` spent 9.5 seconds on
+`create_namespaced_custom_object`. The Kubernetes API server was slow, not the SDK.
+
+#### Story 2: Monitoring Job Error Rates
+
+As a platform operator, I configure an OTel Collector to export metrics to Prometheus. I
+build a Grafana dashboard showing `kubeflow.trainer.errors` by `error.type` and
+`kubeflow.operation.name`. When `TimeoutError` spikes for `get_job`, I know the API server
+is under pressure.
+
```
tariq-hasan, line 206: If controller-side telemetry is out of scope then story 2 is out of scope as well.

This KEP fundamentally addresses the client-side observability story that results from a SDK user spinning up a training job or interactive Spark workload from their local.

Since a SDK user can configure their application to not expose telemetry or expose only a sample then the platform operator would not be able to reliably observe the complete picture of all workloads running on the cluster.

The only way that a platform operator can reliably view observability metrics for all jobs in the cluster is if the controller exposes those metrics to the exporter.

tariq-hasan, line 206: In addition the multi-tenancy story can only be addressed cleanly if there is a clear identifier for a user's identity in the observability signals - be it metrics, traces or logs - that are flushed out to the exporter. A trace ID or a span ID, by itself, would not be enough to identify a user or a user application.

To add to this the exporter would also need to be a centralized OpenTelemetry collector. This goes back to the prior comment regarding a platform operator's need to have a reliable mechanism to collect observability metrics. If every SDK user were to configure their own exporter that reliable mechanism would be missing altogether from the platform operator's point of view.

In other words a true multi-tenant architecture can only be addressed by resolving each telemetry item to its originating namespace and tenant identity as well as using a shared OpenTelemetry collector to flush out metrics from all the SDK processes to some external dashboard.

tariq-hasan: The fundamental idea here is that both client-side and server-side observability have value that they add - the client-side more so for developer debugging and server-side for platform operations. That said, the context propagation from client-side to server-side can only be fully addressed once the server-side observability story is also fleshed out for all relevant upstream components.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+| Metric | Dimensions |
+|---|---|
+| `job.created` | `kubeflow.backend.type`, `kubeflow.trainer.type` |
+| `operation.duration` | `kubeflow.operation.name`, `kubeflow.backend.type`, `error.type` |
+| `errors` | `kubeflow.operation.name`, `error.type`, `kubeflow.backend.type` |
+
+To prevent cardinality explosion, per-job identifiers (`kubeflow.job.name`,
+`kubeflow.runtime.name`) are recorded on spans only, not as metric dimensions.
+
```
tariq-hasan, line 450: Regarding the [following](https://github.com/kubeflow/sdk/pull/382#issuecomment-5271752334) comment these two metrics on the client-side would not fully capture the semantics that they seemingly promise.

For `kubeflow.trainer.job.wait.duration` specifically one may think that this captures "queue-wait visibility, which on a busy GPU cluster is usually what users actually complain about." That would not actually be true since `kubeflow.trainer.job.wait.duration` would be measured on the runtime of the SDK process not on the runtime of the controller. The SDK process runtime is local (it connects to the cluster and captures only a subset of the true cluster state that spans over the time that this process runs on the local) while the controller runtime is what actually captures the workload's true and complete state on the cluster.

A similar argument can be made for `kubeflow.trainer.job.terminal`. This metric, on the user's local setup, would undercount or count only a subset of whatever is the true state on the cluster as the metric itself is dependent on the client maintaining a connection to the remote cluster and observing its state - one cannot expect that to happen reliably in all cases.

```
@andreyvelich two more metrics we could add, both client-side and low-cardinality:

- `kubeflow.trainer.job.wait.duration` — Histogram, seconds. How long `wait_for_job_status` blocked. Dimensioned by terminal `kubeflow.job.status`. This gives queue-wait visibility, which on a busy GPU cluster is usually what users actually complain about.
- `kubeflow.trainer.job.terminal` — Counter. Jobs by final status. Dimensioned by `kubeflow.job.status` and `kubeflow.backend.type`. Gives the error-rate dashboard in Story 2.

Both reuse attributes already in the KEP, so no new dimensions and no cardinality risk. Happy to add them to the metrics table if you want them in scope.
```

tariq-hasan, line 450: The above comment ties in with story 2 and the fundamental premise that client-side observability cannot be relied upon to expose the true state on the cluster.

If a stateless SDK can only measure user behavior and not the cluster state I believe that this would need to be the yardstick that helps determine if a proposed metric belongs in the SDK as a client-side observability metric or not.

All three of `kubeflow.trainer.job.created` (if interpreted as job submission count), `kubeflow.trainer.operation.duration` and `kubeflow.trainer.errors` seem to satisfy this criteria. On the other hand [`job.active`](https://github.com/kubeflow/sdk/pull/382#discussion_r3610508240) did not seem to meet this criteria and that was the reason why it needed to be removed.

tariq-hasan, line 450: To nail down this thought process and exemplify this with an upstream component we can use the case of Spark operator for example.

The operator in this case already exposes controller-level lifecycle [metrics](https://github.com/kubeflow/spark-operator/blob/master/pkg/common/metrics.go#L19-L44) that the client-side SDK attempts to define in terms of `kubeflow.trainer.job.created`, `kubeflow.trainer.job.active`, `kubeflow.trainer.job.wait.duration` and `kubeflow.trainer.job.terminal`. 

This would then mean that the SDK maintains a second lossier version of client-observed state when the upstream operator is already the authoritative source for these metrics.

Therefore, we would want to ensure that there is a strong architectural boundary between client-side metrics and server-side metrics in terms of their semantics and the use cases they address.

docs/proposals/164-opentelemetry-integration/README.md
```diff
@@ -0,0 +1,711 @@
...
+### Stretch Deliverables: OptimizerClient, ModelRegistryClient, SparkClient
+
+These clients follow the same pattern established by `TrainerClient` and reuse the
+same `kubeflow/common/telemetry/` module:
+
+| Client | Architecture | Tracing Pattern | Complexity |
+|---|---|---|---|
+| `OptimizerClient` | Kubernetes backend only | Two-level: INTERNAL / CLIENT | Medium (same pattern as TrainerClient with fewer methods) |
+| `ModelRegistryClient` | REST API wrapper (delegates to `model_registry.ModelRegistry`) | Single-level: INTERNAL only | Low (no backend layer) |
+| `SparkClient` | Kubernetes backend only | Two-level: INTERNAL / CLIENT | Medium (same pattern as TrainerClient) |
```
tariq-hasan, line 557: Upstream Spark already supports robust mechanisms for observability such as Spark UI and Spark history server. These provide more thorough observability data points than what OTel can provide. Add to that the KEP's identifier for the workload -  [`kubeflow.job.name`](https://github.com/XploY04/sdk/blob/6215abc89a260fef466f37644a72de3bce7d4017/docs/proposals/164-opentelemetry-integration/README.md?plain=1#L373) would need to correlate with what Spark uses - the Spark application ID and submission ID - for the Spark UI and history server. And the two-level span model may work for Spark batch job but not for interactive workloads. 

For these reasons primarily the OTel model would only address one part of the observability story for Spark SDK. Spark UI and Spark History Server would be equally important capabilities to support in the SDK.

The above points may deserve some level of acknowledgement in the KEP.

docs/proposals/164-opentelemetry-integration/README.md
tariq-hasan, line 206: The fundamental idea here is that both client-side and server-side observability have value that they add - the client-side more so for developer debugging and server-side for platform operations.

tariq-hasan (commented):
Hi @XploY04! I have added some further comments on the KEP. The main pointers are around resolving the user stories, what client-side observability on its own (without its server-side counterpart) can really accomplish and its impact on the multi-tenancy question and what metrics we choose, context propagation mechanism that scales across different workload types - Trainer or Spark, and why OTel can only address part of the client-side Spark SDK observability problem.

PR #383, opened 2026-03-13
feat(trainer): add job_exists helper method to TrainerClient

### LINKED ISSUE #380: TrainerClient: Add helper method `job_exists()` to check TrainJob presence
## Problem

Currently, checking whether a TrainJob exists requires calling `get_job()` and handling an exception.

Example:

```python
try:
    job = client.get_job("train123")
except RuntimeError:
    # job does not exist
```

This pattern is slightly inconvenient and may lead to repetitive error handling in user code.

## Proposed Solution

Add a small helper method to `TrainerClient`:

```python
def job_exists(self, name: str) -> bool:
```

The method would internally call `get_job()` and return a boolean indicating whether the job exists.

Example usage:

```python
client = TrainerClient()

if client.job_exists("train123"):
    print("TrainJob exists")
```

## Benefits

* Simplifies common existence checks
* Reduces repetitive exception handling
* Improves developer experience when working with TrainJobs

## Implementation Idea

The helper could internally wrap `get_job()`:

```python
def job_exists(self, name: str) -> bool:
    try:
        self.get_job(name)
        return True
    except RuntimeError:
        return False
```

Would this be a useful addition to the SDK?

### Love this feature?

priyank766:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
This PR adds a [job_exists(name: str) -> bool] method to the [TrainerClient]

Currently, verifying if a TrainJob exists requires calling [get_job()] and handling backend-specific exceptions (`RuntimeError` for Kubernetes and `ValueError` for LocalProcess). This helper method provides a unified way to check for job presence by internally handling these exceptions and returning a simple boolean value.

#### Which issue(s) this PR fixes:
Fixes #380

#### Does this PR introduce a user-facing change?
Yes. Users can now use `client.job_exists("job_name")` for existence checks.

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -180,6 +180,21 @@ class TrainerClient:
 
         return self.backend.get_job(name=name)
 
+    def job_exists(self, name: str) -> bool:
+        """Check if a TrainJob exists.
+
+        Args:
+            name: Name of the TrainJob.
+
+        Returns:
+            True if the TrainJob exists, False otherwise.
+        """
+        try:
+            self.get_job(name=name)
+            return True
+        except (RuntimeError, ValueError):
+            return False
+
     def get_job_logs(
         self,
         name: str,
--- a/kubeflow/trainer/api/trainer_client_test.py
+++ b/kubeflow/trainer/api/trainer_client_test.py
@@ -23,6 +23,7 @@ import pytest
 from kubeflow.common.types import KubernetesBackendConfig
 from kubeflow.trainer.api.trainer_client import TrainerClient
 from kubeflow.trainer.backends.localprocess.types import LocalProcessBackendConfig
+from kubeflow.trainer.test.common import TestCase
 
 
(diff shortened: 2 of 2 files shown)


priyank766:
@andreyvelich @kramaranya  should i add the other Exceptions suggested by co-pilot 
current files handles the Value Error and Runtime Errors what do you think ?

kubeflow/trainer/api/trainer_client.py
```diff
@@ -180,6 +180,21 @@ def get_job(self, name: str) -> types.TrainJob:
 
         return self.backend.get_job(name=name)
 
+    def job_exists(self, name: str) -> bool:
```
andreyvelich, line 183: I don't think that this API is really necessary since `get_job()` API can be used to check existance.
@priyank766 Do you get any particular user requests for this API?
/hold

priyank766: @andreyvelich That makes sense — get_job() already covers this use case. I'll close this PR. Thanks for the feedback!

PR #384, opened 2026-03-13
fix(trainer): ignore PEP 668 system python check

robert-bell:

The `pytorch/pytorch:2.10.0-cuda12.8-cudnn9-runtime` image brought in by https://github.com/kubeflow/trainer/pull/3320 has [PEP 668](https://peps.python.org/pep-0668) enabled which causes `pip install` to fail when creating custom train jobs, causing the trainjob to fail.

Generally, it sounds like python images shouldn't be enabling PEP 668, but I think it's a good idea for us to explicitly disable the check whenever we're creating jobs in a container.

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -294,10 +294,10 @@ def get_script_for_python_packages(
         LOG_FILE="{install_log_file}"
         rm -f "$LOG_FILE"
 
-        if PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+        if PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_BREAK_SYSTEM_PACKAGES=1 python -m pip install --quiet \\
             --no-warn-script-location $PIP_OPTS --user $PACKAGES >"$LOG_FILE" 2>&1; then
             echo "Successfully installed Python packages: $PACKAGES"
-        elif PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+        elif PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_BREAK_SYSTEM_PACKAGES=1 python -m pip install --quiet \\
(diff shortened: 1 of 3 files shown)


kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -294,10 +294,10 @@ def get_script_for_python_packages(
         LOG_FILE="{install_log_file}"
         rm -f "$LOG_FILE"
 
-        if PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+        if PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_BREAK_SYSTEM_PACKAGES=1 python -m pip install --quiet \\
```
andreyvelich, line 297: I am wondering if PIP_BREAK_SYSTEM_PACKAGES can break other user's deployment.
WDYT @kubeflow/kubeflow-sdk-team ?

robert-bell: It should be a safe default given it's only set in a container. It just means that pip doesn't complain if it tries to modify an externally-managed python. It doesn't change how pip installs the packages.

kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -294,10 +294,10 @@ def get_script_for_python_packages(
...
             --no-warn-script-location $PIP_OPTS --user $PACKAGES >"$LOG_FILE" 2>&1; then
             echo "Successfully installed Python packages: $PACKAGES"
-        elif PIP_DISABLE_PIP_VERSION_CHECK=1 python -m pip install --quiet \\
+        elif PIP_DISABLE_PIP_VERSION_CHECK=1 PIP_BREAK_SYSTEM_PACKAGES=1 python -m pip install --quiet \\
```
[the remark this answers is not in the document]
robert-bell: The flag should not be added to the localprocess backend. It is not safe to install packages into the system python outside containerised environments, and it's not necessary as this backend uses a venv.

PR #412, opened 2026-03-20
fix(trainer): validate polling_interval is strictly less than timeout in ContainerBackend

### LINKED ISSUE #411: bug(trainer): ContainerBackend.wait_for_job_status missing polling_interval validation
## What happened?

`ContainerBackend.wait_for_job_status` does not validate that `polling_interval`
is less than `timeout`. Both the kubernetes and localprocess backends have this
guard, but the container backend is missing it entirely.

When `polling_interval >= timeout`, the time-based while loop exits before
the first sleep completes — job status is checked zero times — and a
`TimeoutError` is raised immediately with a misleading message.
No error is raised for clearly invalid input.

## What did you expect?

A clear `ValueError` should be raised before the loop starts, matching
the behavior of the other two backends.

## Affected file

`kubeflow/trainer/backends/container/backend.py` — `wait_for_job_status`

## Related

Same class of bug as #400 (kubernetes backend), fixed in #402.

Varun-39:
## Summary

`ContainerBackend.wait_for_job_status` was missing input validation that both the kubernetes and localprocess backends already have.

## The Bug

When `polling_interval >= timeout`, the time-based while loop exits before the first `time.sleep()` completes — job status is checked zero times — and a `TimeoutError` fires immediately with a misleading message. No error is raised for clearly invalid input.

## The Fix

Added a `ValueError` guard matching the established pattern in the other two backends.

## Validation

- Added unit test for `polling_interval == timeout`
- 43/43 tests pass

Fixes #411
Related: #402


Fiona-Waters (commented):
Thanks for this @Varun-39 
Could you please update the tests you added to match syntax and format of existing tests - to use config-driven values instead of name-based branching with hardcoded parameters.

PR #416, opened 2026-03-23
fix(trainer): serialize TorchTune dtype/loss enums by value, add unit tests for get_args_using_torchtune_config and get_trainer_cr_from_builtin_trainer

prabhaharanv:
### What this PR does

Adds comprehensive unit tests for two previously untested utility functions in `kubeflow/trainer/backends/kubernetes/utils.py`:

**`test_get_args_using_torchtune_config`** — 8 parametrized test cases:
1. Empty `TorchTuneConfig` produces empty args
2. `dtype` only produces single dtype arg
3. All scalar fields (`dtype`, `batch_size`, `epochs`, `loss`)
4. `peft_config` with `LoraConfig` appends LoRA args
5. `dataset_preprocess_config` appends dataset args
6. HuggingFace dataset URI — directory path via initializer (`hf://tatsu-lab/alpaca`)
7. HuggingFace dataset URI — file path via initializer (`hf://tatsu-lab/alpaca/data.json`)
8. Invalid dtype raises `ValueError`

**`test_get_trainer_cr_from_builtin_trainer`** — 4 parametrized test cases:
1. Config with `num_nodes` and `batch_size`
2. Config with `resources_per_node`
3. Empty config produces trainer with empty args
4. Invalid config type raises `ValueError`

### Bug fix uncovered by these tests
While writing the tests, I found `get_args_using_torchtune_config` interpolated the
`DataType`/`Loss` enum members directly (e.g. `dtype=DataType.BF16`,
`loss=Loss.CEWithChunkedOutputLoss`) instead of `.value`, so torchtune received the
enum repr rather than the intended `dtype=bf16` /
`loss=torchtune.modules.loss.CEWithChunkedOutputLoss`. Fixed to use `.value`.
The original assertions reused the implementation's own f-string, which made them
tautological and hid this — they now assert the literal produced strings.
Also updated `backend_test.py::test_train` whose expected args encoded the old repr.

### Why

Both `get_args_using_torchtune_config()` and `get_trainer_cr_from_builtin_trainer()` had **zero test coverage** — confirmed via grep across the entire test suite. These functions are critical for converting SDK `BuiltinTrainer` types into Kubernetes CR models and CLI arguments for torchtune.

### How

- Tests follow existing patterns in `utils_test.py` using `TestCase` dataclass and `pytest.mark.parametrize`
- Added helper `_build_builtin_runtime()` to construct a torchtune runtime with `TORCH_TUNE_COMMAND`
- All 93 tests pass (81 existing + 12 new), zero regressions
- Passes `ruff check`, `ruff format --check`, and pre-commit hooks

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils_test.py
+++ b/kubeflow/trainer/backends/kubernetes/utils_test.py
@@ -890,7 +890,7 @@ def _build_builtin_runtime() -> types.Runtime:
                     dtype=types.DataType.BF16,
                 ),
             },
-            expected_output=[f"dtype={types.DataType.BF16}"],
+            expected_output=["dtype=bf16"],
         ),
         TestCase(
             name="all scalar fields produce correct args",
@@ -904,10 +904,10 @@ def _build_builtin_runtime() -> types.Runtime:
                 ),
             },
             expected_output=[
-                f"dtype={types.DataType.FP32}",
+                "dtype=fp32",
                 "batch_size=8",
                 "epochs=3",
-                f"loss={types.Loss.CEWithChunkedOutputLoss}",
+                "loss=torchtune.modules.loss.CEWithChunkedOutputLoss",
             ],
         ),
         TestCase(
@@ -956,7 +956,7 @@ def _build_builtin_runtime() -> types.Runtime:
                 ),
             },
             expected_output=[
-                f"dtype={types.DataType.BF16}",
+                "dtype=bf16",
                 f"dataset.data_dir={constants.DATASET_PATH}/.",
             ],
         ),
@@ -973,6 +973,21 @@ def _build_builtin_runtime() -> types.Runtime:
             },
             expected_output=[
                 f"dataset.data_files={constants.DATASET_PATH}/data.json",
+            ],
+        ),
+        TestCase(
+            name="nested directory dataset uri produces data_dir arg",
+            expected_status=SUCCESS,
+            config={
+                "fine_tuning_config": types.TorchTuneConfig(),
+                "initializer": types.Initializer(
+                    dataset=types.HuggingFaceDatasetInitializer(
+                        storage_uri="hf://tatsu-lab/alpaca/train",
+                    ),
+                ),
+            },
+            expected_output=[
+                f"dataset.data_dir={constants.DATASET_PATH}/train",
             ],
         ),
         TestCase(
@@ -1061,6 +1076,48 @@ def test_get_args_using_torchtune_config(test_case: TestCase):
             expected_output=models.TrainerV1alpha1Trainer(
                 command=["tune", "run"],
                 args=[],
+            ),
+        ),
+        TestCase(
+            name="num_nodes=0 is treated as unset",
+            expected_status=SUCCESS,
+            config={
+                "runtime": _build_builtin_runtime(),
+                "trainer": types.BuiltinTrainer(
+                    config=types.TorchTuneConfig(
+                        num_nodes=0,
+                        batch_size=8,
+                    ),
+                ),
+            },
+            # numNodes is intentionally left unset: `if config.num_nodes:` is falsy for 0.
+            expected_output=models.TrainerV1alpha1Trainer(
+                command=["tune", "run"],
+                args=["batch_size=8"],
+            ),
+        ),
+        TestCase(
+            name="initializer threads dataset arg into trainer args",
+            expected_status=SUCCESS,
+            config={
+                "runtime": _build_builtin_runtime(),
+                "trainer": types.BuiltinTrainer(
+                    config=types.TorchTuneConfig(
+                        batch_size=8,
+                    ),
+                ),
+                "initializer": types.Initializer(
+                    dataset=types.HuggingFaceDatasetInitializer(
+                        storage_uri="hf://tatsu-lab/alpaca/data.json",
+                    ),
+                ),
+            },
+            expected_output=models.TrainerV1alpha1Trainer(
+                command=["tune", "run"],
+                args=[
+                    "batch_size=8",
+                    f"dataset.data_files={constants.DATASET_PATH}/data.json",
+                ],
             ),
         ),
         TestCase(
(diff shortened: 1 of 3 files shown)


kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -857,3 +857,238 @@ def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
...
+        ),
+        TestCase(
+            name="dtype only produces dtype arg",
+            expected_status=SUCCESS,
+            config={
+                "fine_tuning_config": types.TorchTuneConfig(
+                    dtype=types.DataType.BF16,
+                ),
+            },
+            expected_output=[f"dtype={types.DataType.BF16}"],
```
andreyvelich: These expected values reuse the implementation's own f-string, so they evaluate to the enum repr (`dtype=DataType.BF16`, and `loss=Loss.CEWithChunkedOutputLoss` on line 910) — making the assertion tautological: it passes regardless of how the enum serializes. Suggest asserting the literal produced string instead. Also worth double-checking whether torchtune expects the `.value` (`dtype=bf16`) rather than the enum repr. Same applies to line 959.

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -857,3 +857,238 @@ def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
...
+    "test_case",
+    [
+        TestCase(
+            name="valid config with num_nodes and batch_size",
+            expected_status=SUCCESS,
+            config={
+                "runtime": _build_builtin_runtime(),
+                "trainer": types.BuiltinTrainer(
+                    config=types.TorchTuneConfig(
+                        num_nodes=2,
```
andreyvelich, line 1032: Consider adding a `num_nodes=0` (and/or `1`) case to pin the `if config.num_nodes:` boundary in `get_trainer_cr_from_builtin_trainer`.

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -857,3 +857,238 @@ def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
...
+                "runtime": _build_builtin_runtime(),
+                "trainer": types.BuiltinTrainer(
+                    config="invalid_config",
+                ),
+            },
+            expected_error=ValueError,
+        ),
+    ],
+)
+def test_get_trainer_cr_from_builtin_trainer(test_case: TestCase):
```
andreyvelich, line 1136: None of these cases pass an `initializer`, so the initializer→args wiring (the real user path) isn't exercised here. Suggest one case threading an `initializer` and asserting the `data_dir`/`data_files` arg lands in `trainer_cr.args`.

kubeflow/trainer/backends/kubernetes/utils_test.py
```diff
@@ -857,3 +857,238 @@ def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
...
+                    dtype=types.DataType.BF16,
+                ),
+                "initializer": types.Initializer(
+                    dataset=types.HuggingFaceDatasetInitializer(
+                        storage_uri="hf://tatsu-lab/alpaca",
+                    ),
+                ),
+            },
+            expected_output=[
+                f"dtype={types.DataType.BF16}",
```
andreyvelich: Optional: add a single-segment URI (e.g. `hf://alpaca`) to cover the `relative_path == "."` else-branch of the data_dir/data_files split.

andreyvelich (commented):
Thanks for adding this coverage, @prabhaharanv! A few inline suggestions below. This review was done with the help of AI tools.

kubeflow/trainer/backends/kubernetes/utils_test.py
prabhaharanv: You were right, and it was masking an actual bug: the implementation interpolated the enum member (`dtype=DataType.BF16`) instead of `.value` (`dtype=bf16`), which is what torchtune expects. Fixed `get_args_using_torchtune_config` to use `.value` for both `dtype` and `loss`, and changed the assertions to hardcoded literals. This also required updating `backend_test.py::test_train`, whose expected args had encoded the old repr confirming the fix. Retitled the PR accordingly.

kubeflow/trainer/backends/kubernetes/utils_test.py
prabhaharanv: Added a `num_nodes=0` case asserting `numNodes` stays unset, pinning the if `config.num_nodes:` truthiness boundary. The existing `num_nodes=2` case already covers the truthy side.

kubeflow/trainer/backends/kubernetes/utils_test.py
prabhaharanv: Added a case threading a HuggingFace initializer through `get_trainer_cr_from_builtin_trainer` and asserting the `dataset.data_files` arg lands in `trainer_cr.args`.

kubeflow/trainer/backends/kubernetes/utils_test.py
prabhaharanv: `hf://alpaca` can't actually be constructed. `HuggingFaceDatasetInitializer.__post_init__` requires a non-empty path (`hf://<user>/<dataset>`), so it raises before reaching this code. And the existing "directory dataset" case (`hf://tatsu-lab/alpaca`) already exercises `relative_path == "."`. So I added a nested-directory URI (`hf://tatsu-lab/alpaca/train`) instead, which covers the other `data_dir` sub-branch `relative_path != "."` with no file extension that wasn't previously hit.

andreyvelich (commented):
Thanks for the update @prabhaharanv!
Can you sign your commits for DCO please?
/lgtm

prabhaharanv:
> Thanks for the update @prabhaharanv! Can you sign your commits for DCO please? /lgtm

Thanks for the tag @andreyvelich. It's done.

andreyvelich (commented):
/lgtm
/approve

PR #440, opened 2026-03-30
fix(hub): add params to register_model()

### LINKED ISSUE #439: bug(hub): `ModelRegistryClient.register_model()` call drops 3 parameters
### What happened?

`ModelRegistryClient.register_model()` in `sdk/kubeflow/hub/api/model_registry_client.py` does not accept `storage_key`,
  `storage_path`, or `service_account_name` keyword arguments, even though:                                                     
   
  1. The method's **own docstring** (lines 116-120) documents `storage_key` and `storage_path` as available parameters          
  2. The upstream `ModelRegistry.register_model()` in `model-registry/clients/python/src/model_registry/_client.py` fully
  supports all three                                                                                                            
                                                        
  These parameters map to `ModelArtifact` fields used by KServe's `StorageInitializer` to pull models at inference time. Without
   them, SDK users cannot register models with S3/MinIO data connection credentials or Kubernetes ServiceAccount-based auth in a
   single call.                                                                                                                 
                                                        
  **Workaround exists but is poor:** Users must call `register_model()`, then `get_model_artifact()`, mutate the fields, and    
  call `update_model_artifact()` — 3 extra API calls with a race window where the artifact has no storage credentials.
                                                                                                                                
  **Suggested fix:** Add the 3 kwargs to the `register_model()` signature and forward them to `self._registry.register_model()`.
   This is ~10 lines, fully backwards-compatible (all default to `None`), and aligns the docstring with the implementation.
                                                                                                                                
  **File:** `sdk/kubeflow/hub/api/model_registry_client.py`, lines 98-149

`ModelRegistryClient.register_model()` should accept `storage_key`, `storage_path`, and `service_account_name` as keyword
  arguments (matching the upstream `ModelRegistry.register_model()` signature) and forward them to the underlying call, so that 
  models can be registered with storage credentials in a single API call.
                                                                                                                                
  Example of expected usage:                            
  ```python
  client.register_model(
      name="my-model",                                                                                                          
      uri="s3://bucket/model",
      version="1.0",                                                                                                            
      model_format_name="pytorch",                      
      model_format_version="2.0",                                                                                               
      storage_key="my-s3-secret",
      storage_path="models/v1",                                                                                                 
  )

### Environment

Kubernetes version:
N/A

Kubeflow Trainer version:
N/A

Kubeflow Python SDK version:
 Latest main branch — sdk/kubeflow/hub/api/model_registry_client.py  


### Impacted by this bug?

1Ayush-Petwal:
**What this PR does **:

Adds storage_key, storage_path, and service_account_name kwargs to `ModelRegistryClient.register_model()`. These were documented in the docstring but silently dropped — they are now forwarded to the upstream `ModelRegistry.register_model()`.

**Which issue(s) this PR fixes** : 

Fixes #439

**Checklist:**

### DIFF
--- a/kubeflow/hub/api/model_registry_client.py
+++ b/kubeflow/hub/api/model_registry_client.py
@@ -17,6 +17,8 @@ from __future__ import annotations
 from collections.abc import Iterator, Mapping
 from typing import TYPE_CHECKING
 
+from kubeflow.hub.types.types import StorageConfig
+
 if TYPE_CHECKING:
     from model_registry.types import (
         ModelArtifact,
@@ -107,20 +109,18 @@ class ModelRegistryClient:
         owner: str | None = None,
         version_description: str | None = None,
         metadata: Mapping[str, SupportedTypes] | None = None,
-        storage_key: str | None = None,
-        storage_path: str | None = None,
-        service_account_name: str | None = None,
+        storage_config: StorageConfig | None = None,
     ) -> RegisteredModel:
         """Register a model.
 
         This registers a model in the model registry. The model is not downloaded,
         and has to be stored prior to registration.
 
-        Most models can be registered using their URI, along with optional
-        connection-specific parameters, `storage_key` and `storage_path` or,
-        simply a `service_account_name`. URI builder utilities are recommended
-        when referring to specialized storage; for example `utils.s3_uri_from`
-        helper when using S3 object storage data connections.
+        Most models can be registered using their URI, along with an optional
+        `storage_config` describing how KServe should fetch the model at
+        inference time. URI builder utilities are recommended when referring to
+        specialized storage; for example `utils.s3_uri_from` when using S3
+        object storage data connections.
 
         Args:
             name: Name of the model.
@@ -135,15 +135,15 @@ class ModelRegistryClient:
             owner: Owner of the model. Defaults to the client author.
             version_description: Description of the model version.
             metadata: Additional version metadata.
-            storage_key: Name of a Kubernetes Secret containing storage credentials
-                        (e.g., S3 access key/secret for MinIO or AWS).
-            storage_path: Subpath within the storage bucket where the model resides.
-            service_account_name: Kubernetes ServiceAccount name with IRSA/Workload
-                                 Identity for cloud-native auth.
+            storage_config: Storage credentials for the model artifact. Groups
+                           `storage_key`, `storage_path`, and
+                           `service_account_name` used by KServe's
+                           StorageInitializer. See `StorageConfig` for details.
 
         Returns:
             Registered model.
         """
+        storage = storage_config or StorageConfig()
         return self._registry.register_model(
             name=name,
             uri=uri,
@@ -154,9 +154,9 @@ class ModelRegistryClient:
             owner=owner,
             description=version_description,
             metadata=metadata,
-            storage_key=storage_key,
-            storage_path=storage_path,
-            service_account_name=service_account_name,
+            storage_key=storage.storage_key,
+            storage_path=storage.storage_path,
+            service_account_name=storage.service_account_name,
         )
 
     def update_model(self, model: RegisteredModel) -> RegisteredModel:
--- a/kubeflow/hub/api/model_registry_client.py
+++ b/kubeflow/hub/api/model_registry_client.py
@@ -107,6 +107,9 @@ class ModelRegistryClient:
         owner: str | None = None,
         version_description: str | None = None,
         metadata: Mapping[str, SupportedTypes] | None = None,
+        storage_key: str | None = None,
+        storage_path: str | None = None,
+        service_account_name: str | None = None,
     ) -> RegisteredModel:
         """Register a model.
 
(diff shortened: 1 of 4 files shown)


1Ayush-Petwal:
Hi @szaher @fege, would appreciate a review when you get a chance.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -107,6 +107,9 @@ def register_model(
         owner: str | None = None,
         version_description: str | None = None,
         metadata: Mapping[str, SupportedTypes] | None = None,
+        storage_key: str | None = None,
+        storage_path: str | None = None,
+        service_account_name: str | None = None,
```
andreyvelich: We haven't decided yet if we want to export these parameters via Kubeflow SDK APIs.
@kubeflow/kubeflow-hub-team @kubeflow/kubeflow-sdk-team any thoughts?

jonburdo: @1Ayush-Petwal thanks for bringing this up and working on it! Much appreciated :)

A few notes for context: We originally wrote the underlying `ModelRegistry` client class somewhat independently of other sdks under kubeflow I believe. Now we have made an effort to unify things a bit more under this repo and also to make APIs simpler and less overwhelming.

That said, I do think this functionality may be something users should be able to do. I just want to make sure we are thoughtful about the signatures. IMO, following conventions within this repo and making things user-friendly is more important than following the underlying client signature or flattening everything to fit the API.

Therefore, I'm leaning towards suggesting a config data structure rather than flat args. Perhaps something like:
```python
@dataclass
class StorageConfig:
    storage_key: str | None = None
    storage_path: str | None = None
    service_account_name: str | None = None
```

And then `register_model` might take:
```python
  storage_config: StorageConfig | None = None,
```

Thinking ahead, if in the future we wanted to add support for our `model_source_*` args as well, we might add to `register_model`:
```python
source_config: ModelSourceConfig | None = None,
```
(I don't suggest we do this now, but wanted to point it out)

This way we don't have a hug proliferation of args

jonburdo: Some additional context:
- this comment and surrounding ones: https://github.com/kubeflow/sdk/issues/59#issuecomment-3564058861
- PR to add this wrapper client and comments there: https://github.com/kubeflow/sdk/pull/186

1Ayush-Petwal: Thanks @andreyvelich @jonburdo, I went back through PR #186 and issue #59. I earlier thought that it was a documented-but-not-implemented bug,  but got to know these fields were intentionally dropped in PR #186 with the expectation that users needing storage creds would call `update_model_artifact` after registration. 
                                                                                                                              
The `StorageConfig` shape makes sense to me, and it actually addresses something @tarilabs also flagged in #186 of not grouping `model_source_*` in a dataclass.

Here's what I'd propose:         
- New `kubeflow/hub/types/types.py` with `@dataclass class StorageConfig` (3 fields, all None defaults). Hub doesn't have a `types`/module yet; trainer/optimizer/spark all do. Happy to place it inline in `model_registry_client.py` instead if you'd prefer not to introduce the new submodule yet.                                                                                                  
- `register_model(..., storage_config: StorageConfig | None = None)` — unpack at the forwarding call.
- Update the prose docstring at lines 116–123 to describe `storage_config` instead of the flat field names — this is the text that made issue #439 look like a doc/impl mismatch.                        
- Skipping `ModelSourceConfig` per your note.
                                                                                                                                                                                                                     
@andreyvelich @jonburdo let me know if you'd prefer a different direction.

jonburdo: I want to get confirmation that there are no objections from other maintainers, but I believe this is good plan! Thanks for the details.

> New kubeflow/hub/types/types.py

I agree with adding this

> The StorageConfig shape makes sense to me, and it actually addresses something @tarilabs also flagged in https://github.com/kubeflow/sdk/pull/186 of not grouping model_source_* in a dataclass.

Good point - for anyone else reading, [this is the comment](https://github.com/kubeflow/sdk/pull/186#discussion_r2588004688) I believe.

And yes, I agree on updating the docstring. It is currently incorrect on `main`!

jonburdo: I hear no objections! I would say, feel free to go ahead with this plan @1Ayush-Petwal

kubeflow/hub/api/model_registry_client.py
```diff
@@ -146,6 +154,9 @@ def register_model(
             owner=owner,
             description=version_description,
             metadata=metadata,
+            storage_key=storage.storage_key,
+            storage_path=storage.storage_path,
+            service_account_name=storage.service_account_name,
```
jonburdo, line 159: Just a note for awareness: Passing explicitly does mean that if the underlying `ModelRegistry.register_model` were to change the defaults from `None` to something else, we would still pass `None` here rather than using the default.

IMO this is perfectly fine. This just means that this wrapper defines its own defaults as `None` effectively.

jonburdo (commented):
/lgtm
/approve

Great work @1Ayush-Petwal ! Thank you for this contribution.

I've tested locally, and the changes include tests to verify args are forwarded as expected.

PR #447, opened 2026-04-05
feat: inject W3C trace context into TrainJob annotations

Rajneesh180:
Addresses #446.

When `opentelemetry-api` is installed, `train()` now injects the active W3C trace context (`traceparent`, `tracestate`) into TrainJob annotations under the `opentelemetry.io/` prefix. This lets traces started by the SDK propagate to the training controller and worker pods.

The implementation uses a lazy import with a fallback — when the package isn't installed (the common case today), the function is a no-op that returns annotations unchanged with zero overhead. The `opentelemetry.io/` annotation prefix follows the same convention Tekton Pipelines uses for CRD-level context propagation.

**Changes:**

- `inject_trace_context()` in `utils.py` — uses the global OTel propagator to write trace headers into a carrier dict, then merges them as prefixed annotations
- One-line call in `KubernetesBackend.train()` right before TrainJob construction
- 6 unit tests covering all branches (no OTel, no active context, injection, preservation of existing annotations)

No new required dependencies. A follow-up could add `opentelemetry-api` as an optional extra (`pip install kubeflow[telemetry]`).

Relates to #164

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -656,3 +656,39 @@ def get_model_initializer(
         )
 
     raise ValueError(f"Model initializer type is invalid: {type(model)}")
+
+
+# Annotation key prefix for W3C trace context propagation.
+_TRACE_ANNOTATION_PREFIX = "opentelemetry.io/"
+
+
+def inject_trace_context(
+    annotations: dict[str, str] | None,
+) -> dict[str, str] | None:
+    """Inject W3C trace context into annotations when OpenTelemetry is available.
+
+    Uses the global OTel propagator to inject ``traceparent`` / ``tracestate``
+    headers into a carrier dict, then merges them into *annotations* under the
+    ``opentelemetry.io/`` key prefix.
+
+    Returns *annotations* unchanged when the ``opentelemetry`` package is not
+    installed or when no active span context exists.
+    """
+    try:
+        from opentelemetry.propagate import inject
+    except ImportError:
+        return annotations
+
+    carrier: dict[str, str] = {}
+    inject(carrier)
+
+    if not carrier:
+        return annotations
+
+    if annotations is None:
+        annotations = {}
+
+    for key, value in carrier.items():
+        annotations[f"{_TRACE_ANNOTATION_PREFIX}{key}"] = value
+
+    return annotations
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -661,6 +661,10 @@ def get_model_initializer(
 # Annotation key prefix for W3C trace context propagation.
 _TRACE_ANNOTATION_PREFIX = "opentelemetry.io/"
 
+# Only propagate W3C Trace Context headers — not baggage or other
+# propagator-injected keys that could carry PII or unbounded values.
+_W3C_TRACE_HEADERS = frozenset({"traceparent", "tracestate"})
+
 
 def inject_trace_context(
     annotations: dict[str, str] | None,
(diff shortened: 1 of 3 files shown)


kubeflow/trainer/backends/kubernetes/utils.py
```diff
@@ -656,3 +656,39 @@ def get_model_initializer(
...
+    inject(carrier)
+
+    if not carrier:
+        return annotations
+
+    if annotations is None:
+        annotations = {}
+
+    for key, value in carrier.items():
+        annotations[f"{_TRACE_ANNOTATION_PREFIX}{key}"] = value
```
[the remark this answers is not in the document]
Rajneesh180: Good catch. The global propagator can include baggage (and composite propagators could inject arbitrary keys), so filtering to just `traceparent`/`tracestate` is the right call. Fixed in ee38abd — added a `_W3C_TRACE_HEADERS` frozenset and skip anything not in it. Also added a test that verifies baggage gets dropped.

Rajneesh180:
@andreyvelich Quick ping — this is the SDK-side complement to the OTel integration in #164. The approach is intentionally minimal (lazy import, no new required deps). Would appreciate your take on whether annotations are the right propagation mechanism before I invest more time refining the implementation. Happy to adjust the approach based on your feedback.

andreyvelich:
Sorry for the late reply @Rajneesh180!
@kramaranya @astefanutti @abhijeet-dhumal @Fiona-Waters @szaher @dhanishaphadate @XploY04  Would you be able to check this OpenTelemetry integration since you plan to propose this as a GSoC project?
I know we have open KEP for that too: https://github.com/kubeflow/sdk/pull/382

XploY04:
Sure @andreyvelich I will review it.

XploY04:
@andreyvelich should we discuss the KEP first before moving forward with this PR? This implementation is quite different from my KEP (#382), and I think we should align on the design before we lock in the code.

A few differences I noticed:

- The KEP propagates `TRACEPARENT` and `TRACESTATE` by injecting them as environment variables into pods, so the training code can pick up the parent span. This PR adds them as CRD annotations, which helps with controller-side correlation but does not pass the context into the training containers.
- The KEP uses `TraceContextTextMapPropagator` directly, so propagation is W3C-only. This PR uses the global composite propagator through `opentelemetry.propagate.inject`, then filters the result with an allowlist. That works, but it feels like a workaround.
- The KEP puts shared telemetry code in `kubeflow/common/telemetry/propagation.py`, so other backends can reuse it. This PR puts the helper in `kubeflow/trainer/backends/kubernetes/utils.py`, which makes it Kubernetes-specific.

My concern is that if we merge this shape now, we may need to rework the propagation code once the KEP lands. I think it would be better to agree on the KEP design first, then make the PR follow that structure. What do you think?

andreyvelich:
Yes, let's firstly discuss the implementation in the KEP.

I added this item to the next Kubeflow SDK call: https://docs.google.com/document/d/1jH2WAX2ePxOfI4JuiVK9nPlesDMiyg67xzLwhpR7wTQ/edit?tab=t.0

cc @kramaranya @Fiona-Waters

SahilKumar75:
This has been inactive since mid May, so summarizing the open items to help it move forward or be parked cleanly:

* Design: per the SDK call, the annotation based propagation here is pending alignment with KEP #382 (@XploY04). Probably worth settling that before more code.
* Open review points: `inject_trace_context()` merges every propagator key, so it can attach baggage or large/PII values beyond `traceparent`/`tracestate`; and it can overwrite a user provided `opentelemetry.io/*` annotation. Both still need handling.
* The branch is now conflicting with main and needs a rebase.

@Rajneesh180 are you planning to continue this once the KEP direction is settled? Happy to help on the OTel side if useful.

kubeflow/trainer/backends/kubernetes/utils.py
abhijeet-dhumal, line 1: Is this PR waiting for KEP https://github.com/kubeflow/sdk/pull/382 to land first ?
IIUC This belongs in `kubeflow/common/telemetry/propagation.py` per the [KEP](https://github.com/kubeflow/sdk/pull/382/changes#diff-1aace4704c3c7f5dddd31b369977af07719e177febb5cc82f008b9e78a44c261R38) architecture, not in a backend-specific utils file. Container and LocalProcess backends need the same propagation logic (env vars for those), so a shared module avoids duplication when those are added. WDYT?
cc: @XploY04

XploY04: Yes, we have discussed this before in a previous meet, we are waiting for the KEP to be merged, so that the direction is clear.

PR #452, opened 2026-04-09
fix(deps): Bump Kubeflow Trainer API to 2.1 version

andreyvelich:
KF Trainer v2.2 introduced several breaking changes that makes Kubeflow SDK 0.3 release unusable with the `kubeflow-trainer-api=2.2.0` package.
We should not allow to install `2.2.0` version.

Error:
```
if ml_policy.torch and ml_policy.torch.num_proc_per_node: 
AttributeError: 'dict' object has no attribute 'num_proc_per_node'
```

/assign @astefanutti @kramaranya @Fiona-Waters @abhijeet-dhumal @szaher @akshaychitneni

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -1,7 +1,6 @@
 name: E2E Test
 
-on:
-  pull_request
+on: pull_request
 
 concurrency:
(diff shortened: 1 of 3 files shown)


kramaranya (commented):
Thanks!
/lgtm

Issue #458, opened 2026-04-14
Define shared exception hierarchy across Kubeflow SDK clients


MStokluska:
### What

Today, Kubeflow SDK clients (`TrainerClient`, `ModelRegistryClient`, and the upcoming `PipelinesClient`) raise generic Python exceptions (`ValueError`, `RuntimeError`, `TimeoutError`) for domain-specific errors like name resolution failures, compilation errors, and run failures. This makes it difficult for users to catch and handle errors precisely without accidentally catching unrelated exceptions.

### How

Define a shared exception hierarchy in `kubeflow.common.exceptions` (or similar) that all SDK clients adopt. Proposed starting set:

| Exception | Use case |
|---|---|
| `KubeflowError` | Base class for all Kubeflow SDK errors |
| `NameResolutionError(KubeflowError)` | Resource not found by name (pipeline, experiment, job, model, etc.) |
| `CompilationError(KubeflowError)` | Pipeline or job spec failed to compile |
| `RunFailedError(KubeflowError)` | A run/job reached a failed state during a wait operation |
| `KubeflowTimeoutError(KubeflowError)` | A wait operation exceeded its timeout |

The hierarchy should be minimal and extensible — clients can subclass further if needed (e.g. `PipelineNotFoundError(NameResolutionError)`), but the common set covers the majority of cases.

### Acceptance

- [ ] Shared exception module exists in the SDK (e.g. `kubeflow.common.exceptions`)
- [ ] At least one existing client (`TrainerClient` or `ModelRegistryClient`) migrates to the new exceptions
- [ ] `PipelinesClient` (KEP-125) adopts the shared exceptions from the start
- [ ] Exceptions are documented with when each is raised

### Context

Raised during review of [KEP-125: PipelinesClient](https://github.com/kubeflow/sdk/pull/343) — the error handling table uses generic exceptions, and reviewers suggested custom exceptions that are shared across all SDK clients.

Goku2099:
Thanks for opening this, the problem makes sense. Using generic exceptions across clients definitely makes error handling less precise.

I’d be interested in working on this  starting with introducing the shared exception module and then migrating one client (probably TrainerClient) to validate the approach.

nancychetty25:
Hi! I would like to work on this issue as part of my recent open-source project. If it's available, could you please assign it to me? Thank you!

PR #459, opened 2026-04-14
feat: add shared exception hierarchy for Kubeflow SDK

Goku2099:
Introduce a shared exception module (kubeflow.common.exceptions) with a base KubeflowError and a minimal set of specific exceptions.
This lays the foundation for consistent and precise error handling across SDK clients. Migration of existing clients will be done in follow-up PRs.

#458

### DIFF
--- a/kubeflow/common/__init__.py
+++ b/kubeflow/common/__init__.py
@@ -1,0 +1,15 @@
+from .exceptions import (
+    KubeflowError,
+    NameResolutionError,
+    CompilationError,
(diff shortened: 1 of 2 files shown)


Goku2099:
cc @tariq-hasan @Shekharrajak 

I have made some updates to the PR based on the initial feedback and kept the scope focused on introducing the shared exception hierarchy.

Would appreciate your thoughts on the approach whenever you get a chance.

PR #461, opened 2026-04-15
fix: port-forward process cleanup in connect() failure paths

### LINKED ISSUE #460: Port-forward process not cleaned up in connect() failure paths
## Problem

In `kubeflow/spark/backends/kubernetes/backend.py` inside `connect()`, the port-forward process created via `subprocess.Popen` is not properly cleaned up in failure scenarios.

Specifically:
- During retry scenarios (when port-forward dies and is recreated), the previous process is not terminated and reaped.
- In exception and timeout paths, the process may remain running or not be reaped (`wait()` not called).

Goku2099:
## Problem

In `kubeflow/spark/backends/kubernetes/backend.py` inside `connect()`, the port-forward process created via `subprocess.Popen` is not properly cleaned up in failure scenarios.

Specifically:
- During retry scenarios (when port-forward dies and is recreated), the previous process is not terminated and reaped.
- In exception and timeout paths, the process may remain running or not be reaped (`wait()` not called).

## Fix

- Added cleanup before reassigning `pf_proc` in retry paths
- Ensured `wait()` is always called to properly reap subprocesses
- Added a `try/finally` block to guarantee cleanup in failure scenarios
- Preserved active connection on success by avoiding termination of the running process

## Impact

- Prevents orphaned `kubectl port-forward` processes
- Avoids zombie subprocesses
- Improves subprocess lifecycle handling in failure cases

## Related
Fixes #460


test_pf_leak.py
```diff
@@ -0,0 +1,16 @@
...
+proc = subprocess.Popen(
+    ["python", "-m", "http.server", "8000"],
+    creationflags=subprocess.CREATE_NEW_PROCESS_GROUP  # IMPORTANT (Windows)
+)
+
+print("PID:", proc.pid)
+
+time.sleep(3)
+
+print("Exiting WITHOUT cleanup...")
```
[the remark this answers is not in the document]
Goku2099: that file was just for local testing and has been removed from the repo.

Goku2099:
cc @andreyvelich @Shekharrajak @tariq-hasan 
As discussed on Slack, I’ve raised this PR , happy to get feedback if anyone is available.

Shekharrajak (commented):
We should refine this for production; the current method is quite cluttered and could use some simplification. 

Always  add some unit tests (even though unit tests not present) show that we can validate the changes and requirements .

Goku2099:
> We should refine this for production; the current method is quite cluttered and could use some simplification.
> 
> Always add some unit tests (even though unit tests not present) show that we can validate the changes and requirements .

Thanks for the review, I will simplify the logic and add a basic unit test to validate the failure scenarios.

Goku2099:
@Shekharrajak 
I have updated the PR please review whenever you have time. I will keep in mind to keep the code cleaner and include tests going forward.

Goku2099:
@Shekharrajak @tariq-hasan  gentle ping on this PR, would appreciate a review when you have time.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -399,7 +399,24 @@ def get_connect_url(
...
+        except subprocess.TimeoutExpired:
+            proc.kill()
+            try:
+                proc.wait()
+            except Exception:
+                logger.exception(
+                    "Failed to reap port-forward process after kill (pid=%s)",
+                    proc.pid,
+                )
+
```
tariq-hasan: I think it is better to move the try-except block within the second conditional - it only makes sense to wait and perform a SIGKILL if the process is still alive i.e., if `proc.poll()` returns None.

You'd also need to add a third conditional at the end to close the open read file descriptor on `stderr`:
```python
        if proc.stderr:
            proc.stderr.close()
```

Perhaps also a good idea to remove the `self` parameter and make this a module-level function since it is more of a utility function independent of the backend lifecycle.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -426,104 +443,115 @@ def connect(
...
+                f"Spark Connect connection to {connect_url} did not complete "
+                f"within {connect_timeout}s. "
+                "Verify: (1) port-forward target is the Spark Connect server pod, "
+                "(2) PySpark and server Spark major.minor match, "
+                "(3) driver pod logs for gRPC/auth errors; "
+                "see Spark sql/connect for server config."
+            )
+            if pf_proc is not None and pf_proc.poll() is not None:
+                stderr_b = pf_proc.stderr.read() if pf_proc.stderr else b""
+                stderr_str = stderr_b.decode("utf-8", errors="replace").strip() if stderr_b else ""
```
tariq-hasan: A utility function such as

```python
def _read_proc_stderr(proc: subprocess.Popen) -> str:
    """Read and decode stderr from a subprocess.

    Args:
        proc: A subprocess.Popen instance.

    Returns:
        Decoded stderr string, or empty string if unavailable.
    """
    stderr_b = proc.stderr.read() if proc.stderr else b""
    return stderr_b.decode("utf-8", errors="replace").strip() if stderr_b else ""
```

would be helpful since this pattern tends to repeat quite often in the module.

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan, line 618: Although unrelated to the port-forward issue resource cleanup for the CR is also important. We should delete the session if an exception is raised during the connection process.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -426,104 +443,115 @@ def connect(
...
-        if pf_proc is not None and pf_proc.poll() is not None:
-            stderr_b = pf_proc.stderr.read() if pf_proc.stderr else b""
-            stderr_str = stderr_b.decode("utf-8", errors="replace").strip() if stderr_b else ""
-            raise RuntimeError(
-                f"Port-forward process exited with code {pf_proc.returncode} "
-                f"before connect. stderr: {stderr_str}"
-            )
-
-        # Log connection info
+        pf_proc = None
```
tariq-hasan, line 455: Do you want to consider refactoring the `connect` function using a context manager to manage the port-forwarding lifecycle? This would be the cleanest way to contain resource leaks.

```python
@contextlib.contextmanager
def _port_forward(self, info, local_port=None):
    proc = None
    try:
        url, proc = self._start_port_forward(info, port)
        yield (url, proc)
    finally:
        _cleanup_proc(proc)
```

tariq-hasan (changes requested):
@Goku2099 Thanks for raising the PR. I have left a few comments.

/cc @Shekharrajak @andreyvelich

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Updated the cleanup logic to only terminate and wait when the process is alive, added stderr close and moved it to a module level helper.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Added a small helper to read and decode stderr and replaced the duplicated code.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: I will handle session cleanup on failure in a follow-up PR to keep this change focused.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Thanks, this makes sense. I will keep this PR focused on the cleanup fix and can follow up with a context manager refactor separately.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -61,6 +61,29 @@ def _enable_spark_debug_logging() -> None:
...
+            proc.kill()
+            try:
+                proc.wait()
+            except Exception:
+                logger.exception(
+                    "Failed to reap port-forward process after kill (pid=%s)",
+                    proc.pid,
+                )
+    if proc.stderr:
+        proc.stderr.close()
```
tariq-hasan, line 83: You'd need to add a docstring for the `_cleanup_proc` method.

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan, line 404: There should be a `_cleanup_proc(proc)` before `continue` to clean up the open read file descriptor.

tariq-hasan, line 416: There should be a similar `_cleanup_proc(proc)` before `continue` here as well.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -399,7 +420,7 @@ def get_connect_url(
             proc.wait(timeout=5)
```
tariq-hasan: You can replace the terminate and wait with `_cleanup_proc(proc)`.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -399,7 +420,7 @@ def get_connect_url(
...
             logger.warning("Port %s did not become reachable in time for %s", port, key)
         raise RuntimeError(f"Port-forward failed for all candidates in {info.namespace}")
-
+
```
tariq-hasan: You'll need to run the commands in the Makefile to make sure all the linting and oher CI checks are in place. Technically there should be no indentations on empty lines and spacing gaps between module-level functions and classes should match the codebase conventions.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -426,104 +447,113 @@ def connect(
...
+            if pf_proc is not None and pf_proc.poll() is not None:
+                logger.warning("Port-forward died before connect, restarting...")
+                _cleanup_proc(pf_proc)
+                pf_proc = None
+                connect_url, pf_proc = self.get_connect_url(info, local_port=local_port)
+
+            # Create SparkSession with timeout
+            result: list = []
+            exc_holder: list = []
+
```
tariq-hasan, line 523: For consistency it is best to remove the type references as we do not adopt this practice in other places.

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -737,3 +737,20 @@ def test_extract_name_option(kubernetes_backend, test_case):
...
+        "get_connect_url",
+        return_value=("sc://127.0.0.1:15002", mock_proc),
+    ), patch(
+        "kubeflow.spark.backends.kubernetes.backend.SparkSession.builder.remote",
+        side_effect=Exception("connect failed"),
+    ):
+        with pytest.raises(Exception):
+            kubernetes_backend.connect(info=Mock())
+
+    assert mock_proc.wait.called
```
tariq-hasan: The unit test is currently failing. You'd need to review and include tests covering both valid and invalid cases.

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan: Sounds good. Do you want to create a separate tracking issue for this item in Kubeflow SDK? @Goku2099

It appears that resource cleanup is a gap on both Spark SDK as well as Spark Operator when it comes to the SparkConnect CRD. Unlike the SparkApplication CRD, SparkConnect does not have built-in garbage collection and resource cleanup and also does not transition to the `Failed` state in its reconciliation loop. I have raised an issue in Spark Operator for this gap: https://github.com/kubeflow/spark-operator/issues/2927.

On another note there also appears to be a gap in how the states map from the SDK to the control plane - there's a [Running](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/types/types.py#L27) state in the SDK but none in the [SparkConnect CRD](https://github.com/kubeflow/spark-operator/blob/master/api/v1alpha1/sparkconnect_types.go#L175-L179). I am wondering if this might have been intentional or if this was an undetected gap. @andreyvelich @Shekharrajak

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan: Sure. Should there be a separate issue to track this item?

One additional question in this regard is if we want to spawn a `kubectl` subprocess and manage its lifecycle or if we want to use a more Python-native approach such as `kubernetes.stream.portforward` that handles all the lifecycle management and resource cleanup on our behalf.

cc @andreyvelich @Shekharrajak

tariq-hasan (changes requested):
@Goku2099 Thanks for addressing the review comments. I have left a few nits and additional comments on the tests.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Added a docstring for _cleanup_proc.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: added

Goku2099: added _cleanup_proc(proc) before continue here as well

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Replaced terminate and wait with _cleanup_proc(proc) to keep cleanup consistent.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Ran the Makefile checks and fixed formatting and spacing issues.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Removed the type annotations to keep it consistent with the rest of the codebase.

kubeflow/spark/backends/kubernetes/backend_test.py
Goku2099: Fixed the failing test and added coverage for both failure and success paths.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Yeah, that makes sense. I have created a tracking issue for this (#476) and linked it here. Also interesting point about the Spark Operator gap and the state mismatch, would be good to get thoughts from @andreyvelich and @shekharrajak on whether this is intentional or something we should align.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Yeah, makes sense. I can create a separate tracking issue for the context manager and refactor part.
For the port-forward approach, I think kubectl subprocess keeps it closer to current behavior, but open to switching to a more Python-native approach if that aligns better would be good to get thoughts from @andreyvelich and @shekharrajak.

kubeflow/spark/backends/kubernetes/backend_test.py
tariq-hasan: Adding tests for `connect()` is good since these are missing currently, But they would need to follow the pytest test case enumeration pattern like in other tests for consistency.

More broadly speaking test cases for cleanup would also need to be added for `test_get_connect_url()`. The goal is basically to add tests to verify that the subprocess is cleaned up on every exit path.

tariq-hasan (changes requested):

kubeflow/spark/backends/kubernetes/backend_test.py
Goku2099: Added tests for cleanup in both connect() and get_connect_url() across failure paths, and aligned them with the existing parametrized test pattern.
happy to get feedback

PR #465, opened 2026-04-17
fix: Safely parse fractional kubernetes cpu limits

### LINKED ISSUE #466: ValueError crash when parsing fractional Kubernetes CPU limits in Spark Client
### Description
Currently, the `kubeflow.spark.backends.kubernetes.utils` module expects Kubernetes CPU resource requests to be implicitly convertible to standard python integers using `int()`.

Specifically, in `get_server_spec_from_driver` and `get_executor_spec_from_executor`, the following logic exists:

```python
cores = int(resource_dict["cpu"])
```

This is fundamentally incompatible with standard Kubernetes resource limits, which are almost entirely specified via strings defining fractional hardware capacities like `"500m"` (millicores) or `"1.5"`. Since `int("500m")` and `int("1.5")` fail directly with `ValueError: invalid literal for int() with base 10`, setting fractional custom CPU limits currently crashes the Spark client wrapper entirely.

### Expected Behavior
Spark mandates that allocated CPU cores for Spark Submit and PySpark sessions correspond strictly to whole integers (as they represent concrete JVM worker threads).

Instead of raising a `ValueError` by performing a blunt typecast upon encountering a trailing `"m"` character or decimal, the Spark client should round Kubernetes millicores to the next ceiling integer using `math.ceil()`. For example:

- `"500m"` -> `1` core
- `"1500m"` -> `2` cores
- `"2.5"` -> `3` cores

A specialized `_cpu_kubernetes_to_spark` conversion utility should be implemented mirroring `_memory_kubernetes_to_spark` to handle this logic securely.

### Steps to Reproduce
Attempt to create a Spark Connect session with fractional driver or executor cores:

```python
client = SparkClient()
spark = client.connect(
    driver=Driver(resources={"cpu": "500m", "memory": "2Gi"}),
    # OR 
    resources_per_executor={"cpu": "1.5"}
)
```
The client will crash locally returning a built-in `ValueError`.

chahatsagarmain:

## Problem:
When defining custom driver or executor options in the Spark client, allocating CPU limits via `resources={"cpu": "500m"}` or using decimal values like `"1.5"` routinely crashes logic in `utils.py`. The backend was explicitly casting this configuration metric directly to an integer `int(resource_dict["cpu"])`, which triggers an immediate `ValueError` for fractional millicore definitions (e.g. `500m`).

## Fix:

Fix : #466 

Implemented a custom parser utility `_cpu_kubernetes_to_spark` that safely handles fractional limits. It parses decimal values and standard Kubernetes `"m"` suffixes into floats, and computes the mathematical ceiling (`math.ceil`) so that PySpark allocates the next highest whole number of JVM threads/cores successfully.

**Checklist:**

### DIFF
--- a/kubeflow/spark/backends/kubernetes/utils.py
+++ b/kubeflow/spark/backends/kubernetes/utils.py
@@ -14,6 +14,7 @@
 
 """Utility functions for Kubernetes Spark backend."""
 
+import math
 import re
 from typing import Any
 from urllib.parse import urlparse
@@ -71,6 +72,21 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
     return num + spark_suffix
 
 
+def _cpu_kubernetes_to_spark(cpu: str) -> int:
+    """Convert Kubernetes CPU request to integer Spark cores (rounded up).
+
+    Spark expects integer cores. Kubernetes allows fractional CPUs (e.g., "500m", "1.5").
+    """
+    if not cpu:
+        return 1
+    try:
+        if cpu.endswith("m"):
+            return math.ceil(float(cpu[:-1]) / 1000)
+        return math.ceil(float(cpu))
+    except ValueError:
+        return 1
+
+
 def build_service_url(info: SparkConnectInfo) -> str:
     """Build Spark Connect URL from session info.
 
@@ -102,7 +118,7 @@ def get_server_spec_from_driver(
     if driver:
         if driver.resources:
             if "cpu" in driver.resources:
-                cores = int(driver.resources["cpu"])
+                cores = _cpu_kubernetes_to_spark(driver.resources["cpu"])
             if "memory" in driver.resources:
                 memory = _memory_kubernetes_to_spark(driver.resources["memory"])
 
@@ -162,7 +178,7 @@ def get_executor_spec_from_executor(
 
     if resource_dict:
         if "cpu" in resource_dict:
-            cores = int(resource_dict["cpu"])
+            cores = _cpu_kubernetes_to_spark(resource_dict["cpu"])
         if "memory" in resource_dict:
             memory = _memory_kubernetes_to_spark(resource_dict["memory"])
 
--- a/kubeflow/spark/backends/kubernetes/utils_test.py
+++ b/kubeflow/spark/backends/kubernetes/utils_test.py
@@ -19,6 +19,7 @@ import pytest
 
 from kubeflow.spark.backends.kubernetes import constants
 from kubeflow.spark.backends.kubernetes.utils import (
+    _cpu_kubernetes_to_spark,
     _memory_kubernetes_to_spark,
     build_service_url,
     build_spark_connect_cr,
@@ -46,6 +47,27 @@ class TestMemoryKubernetesToSpark:
     )
     def test_conversion(self, k8s_memory: str, expected_spark: str) -> None:
         assert _memory_kubernetes_to_spark(k8s_memory) == expected_spark
+
+
+class TestCpuKubernetesToSpark:
+    """Tests for _cpu_kubernetes_to_spark."""
+
+    @pytest.mark.parametrize(
+        "k8s_cpu,expected_spark",
+        [
+            ("500m", 1),
+            ("1500m", 2),
+            ("1", 1),
+            ("1.5", 2),
+            ("2", 2),
+            ("0.1", 1),
+            ("", 1),
+            (None, 1),
+            ("invalid", 1),
+        ],
+    )
+    def test_conversion(self, k8s_cpu: str, expected_spark: int) -> None:
+        assert _cpu_kubernetes_to_spark(k8s_cpu) == expected_spark
 
 
 class TestGenerateSessionName:


Goku2099:
@chahatsagarmain 
I went through this PR, and currently the behavior is that invalid inputs or parsing failures default to 1. This can silently mask misconfigurations, so maybe adding a small log/message here would make it clearer for users when their input isn’t parsed as expected.

chahatsagarmain:
> @chahatsagarmain I went through this PR, and currently the behavior is that invalid inputs or parsing failures default to 1. This can silently mask misconfigurations, so maybe adding a small log/message here would make it clearer for users when their input isn’t parsed as expected.

Yes , I have left it for now for review by maintainers .

Goku2099:
I went through the reference shared by Shekhar and the validation logic. Since invalid CPU formats are already rejected via ValidationError, this helper mostly handles valid inputs.
So maybe the focus should be ensuring it always returns a valid Spark value (≥1 core), rather than relying on fallback behavior.

chahatsagarmain:
> I went through the reference shared by Shekhar and the validation logic. Since invalid CPU formats are already rejected via ValidationError, this helper mostly handles valid inputs. So maybe the focus should be ensuring it always returns a valid Spark value (≥1 core), rather than relying on fallback behavior.

That makes sense , I have left it for now for review by maintainers based on their decision we can make a slight change here and in the unit test .

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -71,6 +72,21 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
...
+    Spark expects integer cores. Kubernetes allows fractional CPUs (e.g., "500m", "1.5").
+    """
+    if not cpu:
+        return 1
+    try:
+        if cpu.endswith("m"):
+            return math.ceil(float(cpu[:-1]) / 1000)
+        return math.ceil(float(cpu))
+    except ValueError:
+        return 1
```
tariq-hasan, line 87:
```suggestion
    cpu = cpu.strip()
    if cpu.endswith("m"):
        cores = int(cpu[:-1]) / 1000
    else:
        cores = float(cpu)
    if not math.isfinite(cores) or cores <= 0:
        raise ValueError(f"Invalid CPU value: {cpu!r}")
    return math.ceil(cores)
```

A few things to note about this change.

A default `1` should not be returned for `""`, `None` or for invalid strings such as `"four"`, `"50O0m"`  and `"invalid"`. Otherwise we'd be hiding real user errors and using a default that won't make sense. `int()` and `float()` raise `ValueError` for these kinds of inputs that are outright invalid so no need for the exception block there.

More nuanced cases such as`"inf"`, `"infinity"`, `"nan"`, `"0"`, negatives like `"-1"` or `"-500m"`, and decimal millicores like `"1.5m"` should raise errors instead of being let through.

A value such as `" 1500m  "` could be considered valid and be supported through `cpu = cpu.strip()`.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -71,6 +72,21 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
     return num + spark_suffix
 
 
+def _cpu_kubernetes_to_spark(cpu: str) -> int:
+    """Convert Kubernetes CPU request to integer Spark cores (rounded up).
+
+    Spark expects integer cores. Kubernetes allows fractional CPUs (e.g., "500m", "1.5").
+    """
```
tariq-hasan, line 79: This is more of a nit but the docstrings for both `_cpu_kubernetes_to_spark` and  `_memory_kubernetes_to_spark` should match the module's convention in terms of having Args/Returns/Raises blocks.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -48,6 +49,27 @@ def test_conversion(self, k8s_memory: str, expected_spark: str) -> None:
...
+            ("0.1", 1),
+            ("", 1),
+            (None, 1),
+            ("invalid", 1),
+        ],
+    )
+    def test_conversion(self, k8s_cpu: str, expected_spark: int) -> None:
+        assert _cpu_kubernetes_to_spark(k8s_cpu) == expected_spark
+
+
```
tariq-hasan, line 72: The code currently tests only valid cases.

We'd want to add additional tests for invalid cases as well - for both `TestCpuKubernetesToSpark` and `TestMemoryKubernetesToSpark` - something along the lines of

```
def test_invalid_input(self) -> None:
    with pytest.raises(ValueError):
        _cpu_kubernetes_to_spark("abc")
```

tariq-hasan (changes requested):
@chahatsagarmain Thanks for raising the PR. I have left a few comments.

/cc @Shekharrajak @andreyvelich

tariq-hasan:
The changes tracked by this PR have been covered as part of https://github.com/kubeflow/sdk/pull/521.

Ref: https://github.com/kubeflow/sdk/pull/521#discussion_r3524007793.

/close

PR #485, opened 2026-05-07
chore(optimizer): add comprehensive unit tests for KubernetesBackend

Slowlybomb:
#### What this PR does / why we need it:

This PR adds comprehensive unit tests for the Optimizer SDK's `KubernetesBackend`.

Previously, the test file only contained a single test (`test_optimize`) verifying that `optimize()` does not mutate caller-provided inputs. This PR expands coverage from 1 test function to 8, with 25 parametrized test cases covering all public methods: `optimize`, `get_job`, `list_jobs`, `get_job_logs`, `get_best_results`, `wait_for_job_status`, `delete_job`, and `get_job_events`. Each method is tested with success paths and error scenarios (`TimeoutError`, `RuntimeError`, `ValueError`) using mock Kubernetes APIs.

The test structure mirrors `kubeflow/trainer/backends/kubernetes/backend_test.py`, using the shared `TestCase` dataclass with `pytest.mark.parametrize` and helper factories for Experiment/Trial CRs.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

No. This PR only adds unit tests for existing backend functionality.

#### Checklist:

### DIFF
--- a/kubeflow/optimizer/backends/kubernetes/backend_test.py
+++ b/kubeflow/optimizer/backends/kubernetes/backend_test.py
@@ -15,18 +15,65 @@
 """Unit tests for the KubernetesBackend class in the Kubeflow Optimizer SDK.
 
 This module uses pytest and unittest.mock to simulate Kubernetes API interactions.
-It tests that KubernetesBackend.optimize() does not mutate caller-provided inputs.
+It tests KubernetesBackend's behavior across job listing, resource creation, deletion,
+log retrieval, event filtering, and status waiting.
 """
 
+from dataclasses import asdict
+import datetime
+import multiprocessing
+from typing import Any, TypeVar
 from unittest.mock import Mock, patch
 
+from kubeflow_katib_api import models
 import pytest
 
 from kubeflow.common.types import KubernetesBackendConfig
 from kubeflow.optimizer.backends.kubernetes.backend import KubernetesBackend
-from kubeflow.optimizer.types.search_types import Search
-from kubeflow.trainer.test.common import FAILED, SUCCESS, TestCase
-from kubeflow.trainer.types.types import CustomTrainer, TrainJobTemplate
+from kubeflow.optimizer.constants import constants
+from kubeflow.optimizer.types.algorithm_types import RandomSearch
+from kubeflow.optimizer.types.optimization_types import (
+    Metric,
+    Objective,
+    OptimizationJob,
+    Result,
+    Trial,
+    TrialConfig,
+)
+from kubeflow.optimizer.types.search_types import (
+    ContinuousSearchSpace,
+    Distribution,
+    Search,
+)
+import kubeflow.trainer.constants.constants as trainer_constants
+from kubeflow.trainer.test.common import (
+    DEFAULT_NAMESPACE,
+    FAILED,
+    RUNTIME,
+    SUCCESS,
+    TIMEOUT,
+    TestCase,
+)
+from kubeflow.trainer.types.types import (
+    CustomTrainer,
+    Event,
+    Runtime,
+    RuntimeTrainer,
+    Step,
+    TrainerType,
+    TrainJob,
+    TrainJobTemplate,
+)
+
+T = TypeVar("T")
+
+BASIC_OPTIMIZATION_JOB_NAME = "basic-opt-job"
+BASIC_TRIAL_NAME = "basic-trial"
+BASIC_TRIAL_NAME_2 = "basic-trial-2"
+
+# --------------------------
+# Fixtures
+# --------------------------
 
 
 @pytest.fixture
@@ -37,10 +84,22 @@ def optimizer_backend():
         patch(
             "kubernetes.client.CustomObjectsApi",
             return_value=Mock(
-                create_namespaced_custom_object=Mock(return_value=None),
+                create_namespaced_custom_object=Mock(side_effect=conditional_error_handler),
+                delete_namespaced_custom_object=Mock(side_effect=conditional_error_handler),
+                get_namespaced_custom_object=Mock(
+                    side_effect=get_namespaced_custom_object_response
+                ),
+                list_namespaced_custom_object=Mock(
+                    side_effect=list_namespaced_custom_object_response
+                ),
             ),
         ),
-        patch("kubernetes.client.CoreV1Api", return_value=Mock()),
+        patch(
+            "kubernetes.client.CoreV1Api",
+            return_value=Mock(
+                list_namespaced_event=Mock(side_effect=mock_list_namespaced_event),
+            ),
+        ),
         patch(
             "kubeflow.trainer.backends.kubernetes.backend.KubernetesBackend.verify_backend",
             return_value=None,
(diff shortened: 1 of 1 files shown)


Slowlybomb:
Hey @kramaranya, ready for review

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -112,8 +485,7 @@ def test_optimize(optimizer_backend, test_case):
         assert trial_template.trainer.func_args == original_func_args
 
     except Exception as e:
-        assert test_case.expected_status != SUCCESS
```
Fiona-Waters, line 115: What is the reason for removing this?
It guards against unexpected exceptions in success-path cases being silently swallowed, maybe it should be re-added and something similar added to the additional test cases?

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +494,376 @@ def test_optimize(optimizer_backend, test_case):
...
+        TestCase(
+            name="valid flow with all defaults",
+            expected_status=SUCCESS,
+            config={"name": BASIC_OPTIMIZATION_JOB_NAME},
+            expected_output=None,
+        ),
+        TestCase(
+            name="timeout error when deleting job",
+            expected_status=FAILED,
+            config={"namespace": TIMEOUT},
```
Fiona-Waters: For completeness, `"name": BASIC_OPTIMIZATION_JOB_NAME` could be added to the `config` in the timeout and runtime error cases here too, so `delete_job` isn't called with `None` as the job name.

Fiona-Waters (commented):
Thanks for this contribution @Slowlybomb - just a couple of comments

Fiona-Waters (commented):
/lgtm

andreyvelich (commented):
Thanks @Slowlybomb for the substantial expansion of test coverage here! This review was produced with AI tools, with inline comments focused on coverage gaps and consistency with the trainer reference pattern.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
-            name="negative polling_interval raises ValueError",
+            name="polling interval is more than timeout error",
             expected_status=FAILED,
-            config={"name": "test-job", "timeout": 10, "polling_interval": -1},
+            config={
+                "name": BASIC_OPTIMIZATION_JOB_NAME,
+                "timeout": 1,
+                "polling_interval": 2,
+            },
             expected_error=ValueError,
```
andreyvelich, line 970: Coverage regression from the original test: the original file had three `polling_interval` validation cases (equal-to-timeout, zero, negative). Only the "greater-than" case is kept here. The backend at `backend.py:287-295` has two distinct validation branches (`<= 0` and `>= timeout`); please restore at least a `polling_interval=0` case and a `polling_interval==timeout` case so each branch is exercised.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
+        ),
+        TestCase(
+            name="runtime error when getting job logs",
+            expected_status=FAILED,
+            config={"name": RUNTIME},
+            expected_error=RuntimeError,
+        ),
+    ],
+)
+def test_get_job_logs(optimizer_backend, test_case):
```
andreyvelich, line 746: `get_job_logs` has several branches that aren't exercised here: (a) explicit `trial_name` argument (skips `_get_best_trial`), (b) `follow=True` passed through, (c) `best_trial` non-None path, (d) empty `trials` early return, (e) `pod_name is None` early return after the step loop. Please add parametrized cases to cover these.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
+
+    def mock_get_job(name):
+        job = original_get_job(name)
+        if test_case.config.get("name") == "failed-job":
+            job.status = constants.OPTIMIZATION_JOB_FAILED
+        elif test_case.expected_status == SUCCESS:
+            job.status = constants.OPTIMIZATION_JOB_COMPLETE
+        return job
+
+    optimizer_backend.get_job = mock_get_job
```
andreyvelich: Replacing `optimizer_backend.get_job` with a closure bypasses the real CR→OptimizationJob conversion and the status-mapping logic in `__get_optimization_job_from_cr` (backend.py:503-512). Also, the `callbacks` parameter is never tested. Consider (a) adding a case that passes a callback and asserts it was invoked, and (b) using the CR-based mock with evolving status conditions rather than a closure.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="wait for complete status (default)",
+            expected_status=SUCCESS,
+            config={"name": BASIC_OPTIMIZATION_JOB_NAME},
+            expected_output=get_optimization_job_data_type(
+                name=BASIC_OPTIMIZATION_JOB_NAME,
+            ),
```
andreyvelich, line 914: `expected_output` is populated here and at L719 but never asserted against in the test body — only `isinstance(job, OptimizationJob)` and `job.status` are checked (L783-784). Either assert `asdict(job) == asdict(test_case.expected_output)` (matches the pattern used in `test_get_job`), or drop the unused field.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
             expected_status=FAILED,
-            config={"name": "test-job", "timeout": 10, "polling_interval": 10},
-            expected_error=ValueError,
+            config={"name": RUNTIME},
+            expected_error=RuntimeError,
+        ),
+    ],
+)
+def test_get_job(optimizer_backend, test_case):
+    """Test KubernetesBackend.get_job with basic success path."""
```
andreyvelich: Docstring says "basic success path" but the parametrization includes TIMEOUT and RUNTIME error cases. Suggest "Test KubernetesBackend.get_job with success and error paths" to match the sibling pattern at L602/L649/L763. Same comment applies to L561 (`test_list_jobs`) and L817 (`test_delete_job`).

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -50,9 +107,293 @@ def optimizer_backend():
...
+    )
+    return mock_thread
+
+
+def mock_trainer_get_job(name: str) -> TrainJob:
+    """Return a mock TrainJob for the given trial name."""
+    return create_mock_trainjob(name)
+
+
+T = TypeVar("T")
```
andreyvelich: `TypeVar` declarations conventionally go near the top of the file with imports. Please move `T = TypeVar("T")` to just below the imports block (~L66) for consistency with standard Python style.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -97,12 +468,14 @@ def test_optimize(optimizer_backend, test_case):
...
     try:
-        optimizer_backend.optimize(
+        optimizer_backend.namespace = test_case.config.get("namespace", DEFAULT_NAMESPACE)
+        job_name = optimizer_backend.optimize(
             trial_template=trial_template,
             search_space=search_space,
         )
 
         assert test_case.expected_status == SUCCESS
+        assert isinstance(job_name, str) and len(job_name) > 0
```
andreyvelich, line 492: Success path only checks the returned `job_name` is a non-empty string. It never verifies that `create_namespaced_custom_object` was called with an Experiment containing the expected parameters / trialParameters / objective / algorithm — a regression that silently dropped `parameters_spec` would pass this test. Consider adding `optimizer_backend.custom_api.create_namespaced_custom_object.assert_called_once()` and inspecting the payload, matching the approach used by the trainer reference at `backend_test.py:1373-1379`.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -122,33 +495,383 @@ def test_optimize(optimizer_backend, test_case):
...
+            expected_error=TimeoutError,
         ),
         TestCase(
-            name="polling_interval equal to timeout raises ValueError",
+            name="runtime error when getting job",
             expected_status=FAILED,
-            config={"name": "test-job", "timeout": 10, "polling_interval": 10},
-            expected_error=ValueError,
+            config={"name": RUNTIME},
+            expected_error=RuntimeError,
```
andreyvelich, line 538: No test constructs an Experiment CR with `status.conditions` set, so the status-mapping logic in `backend.py:503-512` (mapping `ExperimentSucceeded`, `OptimizationJobFailed`, and the trial-running → `OPTIMIZATION_JOB_RUNNING` fallback) is completely untested. Consider extending `create_experiment_cr` callers with conditions so each status branch is exercised.

kubeflow/optimizer/backends/kubernetes/backend_test.py
```diff
@@ -50,9 +107,293 @@ def optimizer_backend():
...
+    if args[2] == TIMEOUT:
+        raise multiprocessing.TimeoutError()
+    elif args[2] == RUNTIME:
+        raise RuntimeError()
+
+
+def get_namespaced_custom_object_response(*args, **kwargs):
+    """Return a mocked Experiment object."""
+    mock_thread = Mock()
+    if args[4] == TIMEOUT:
```
andreyvelich, line 143: `conditional_error_handler` keys errors on `args[2]` (namespace) while this handler keys on `args[4]` (name). This is correct given each API signature but confusing for readers decoding positional indexes, and brittle to a future refactor that switches to kwargs. Consider adding a brief comment explaining which positional arg is used, or switching to kwargs-based matching.

Slowlybomb: Fair enough. I'll add inline comments

kubeflow/optimizer/backends/kubernetes/backend_test.py
Slowlybomb: Will move it up to the imports

kubeflow/optimizer/backends/kubernetes/backend_test.py
Slowlybomb: You're right, just checking the return value isn't enough. I'll add an `assert_called_once()` and verify the key fields in the Experiment payload.

kubeflow/optimizer/backends/kubernetes/backend_test.py
Slowlybomb: I'll add cases with status.conditions set to exercise the different status branches.

Slowlybomb:
@andreyvelich, I applied required changes to commit. Please review it. Thank you for reviewing!

andreyvelich (commented):
Thanks @Slowlybomb!
Can you sign your commits please?
/lgtm
/approve

kramaranya (approved):
Thank you @Slowlybomb!
/lgtm

PR #495, opened 2026-05-15
feat(ci): Replace Trivy CVE scanning with uv audit + OSV-Scanner

### LINKED ISSUE #478: feat(ci): Replace Trivy CVE scanning with uv audit + OSV-Scanner
The Trivy-based CVE scanning workflows were removed in March 2026 ([#427](https://github.com/kubeflow/sdk/pull/427), [f2dca02](https://github.com/kubeflow/sdk/commit/f2dca02c544e0136dce1ed62d390a373c3955c18)) after the aquasecurity/trivy-action was [compromised](https://github.com/kubeflow/katib/issues/2642). This left the SDK without:

 - Nightly CVE scan + auto-fix (was trivy-cve-scan.yaml) — scanned uv.lock, uploaded SARIF to the Security tab, and auto-created fix PRs for CVSS 7.0+ vulnerabilities
 - PR lockfile validation (was validate-lockfile.yaml) — diff-aware scan that flagged new vulnerabilities introduced by a PR.
 
 The helper scripts (.github/scripts/), the cleanup-overrides.yaml workflow, and Dependabot are all still in place. Only the scanner itself needs replacing.

Proposal
Replace Trivy with two open-source, free tools:

[uv audit](https://docs.astral.sh/uv/reference/cli/#uv-audit) — already bundled with uv (no new CI dependencies). Scans uv.lock natively against the PyPI Advisory Database. Use as the PR gate.
[OSV-Scanner](https://github.com/google/osv-scanner) (Apache-2.0, Google) — aggregates 30+ vulnerability databases (NVD, PyPA, GitHub, etc.), supports uv.lock since v2.0, produces SARIF. Use as the nightly scan with SARIF upload to the Security tab.

This approach extends to Go repos (Trainer, Katib, Spark Operator) by pairing govulncheck (PR gate) with OSV-Scanner (nightly) — same pattern, consistent across the project.

/area security

Fiona-Waters:
## Summary

Replaces the Trivy-based CVE scanning workflows (removed in #427) with two new tools, as proposed in #478:

- **`validate-lockfile.yaml`**: Non-blocking PR check using `uv audit` that diffs vulnerabilities between PR and base branch. Posts an informational comment if new vulns are introduced — does not block merging.
- **`osv-scanner.yaml`**: Nightly scan using OSV-Scanner CLI (v2.3.8) with SARIF upload to the GitHub Security tab and auto-fix PRs. Integrates with existing `.github/scripts/` utilities (`update_overrides.py`, `compare_versions.py`, `extract_version.py`) and the `cleanup-overrides` workflow for the full fix lifecycle.
- **`osv-scanner.toml`**: Minimal config for suppressing false positives with expiry dates.

### Design decisions
- Both `uv audit` and OSV-Scanner query the same OSV.dev database — coverage is identical for Python/PyPI
- `uv audit` is used for the PR check because it's zero-dependency (bundled with `uv`), but it is experimental — a note in the workflow documents the fallback path (switch to OSV-Scanner JSON) if the text output format changes
- Both workflows are **non-blocking** — engineers can always merge PRs
- The nightly auto-fix flow tries `uv lock --upgrade-package` first, then falls back to `override-dependencies` via the existing `update_overrides.py` script when natural upgrades can't reach the fix version
- OSV-Scanner binary version is pinned (v2.3.8) — Dependabot won't auto-bump this, so manual updates are needed

Closes #478

## Test plan
- [ ] Verify `validate-lockfile.yaml` triggers on PRs that modify `uv.lock` or `pyproject.toml`
- [ ] Verify PR comment is posted when new vulns are introduced and deleted when resolved
- [ ] Trigger `osv-scanner.yaml` via `workflow_dispatch` and verify SARIF appears in Security tab
- [ ] Verify nightly auto-fix creates a PR when fixable vulns exist
- [ ] Confirm neither workflow blocks PR merging

### DIFF
--- a/.github/workflows/osv-scanner.yaml
+++ b/.github/workflows/osv-scanner.yaml
@@ -1,0 +1,235 @@
+name: OSV-Scanner Vulnerability Scan
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+
+jobs:
+  osv-scan:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      security-events: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install OSV-Scanner
+        # Bump cadence: check https://github.com/google/osv-scanner/releases monthly
+        # Dependabot cannot bump curl-installed binaries — update version + checksum manually
+        run: |
+          OSV_VERSION="2.3.8"
+          EXPECTED_SHA="bc98e15319ed0d515e3f9235287ba53cdc5535d576d24fd573978ecfe9ab92dc"
+          curl -fsSL "https://github.com/google/osv-scanner/releases/download/v${OSV_VERSION}/osv-scanner_linux_amd64" -o osv-scanner
+          ACTUAL_SHA=$(sha256sum osv-scanner | awk '{print $1}')
+          if [ "$ACTUAL_SHA" != "$EXPECTED_SHA" ]; then
+            echo "::error::OSV-Scanner checksum mismatch! Expected ${EXPECTED_SHA}, got ${ACTUAL_SHA}"
+            exit 1
+          fi
+          chmod +x osv-scanner
+          sudo mv osv-scanner /usr/local/bin/
+
+      - name: Run OSV-Scanner (SARIF for Security Tab)
+        # continue-on-error so a SARIF failure doesn't block the JSON scan + auto-fix below
+        continue-on-error: true
+        run: |
+          rc=0
+          osv-scanner scan \
+            --format sarif \
+            --output osv-results.sarif \
+            -L uv.lock || rc=$?
+          # Exit code 1 = vulns found (expected), 0 = clean. Any other code is a real failure.
+          if [ "$rc" -ne 0 ] && [ "$rc" -ne 1 ]; then
+            echo "::error::OSV-Scanner failed with unexpected exit code $rc"
+            exit 1
+          fi
+
+      - name: Upload SARIF to GitHub Security tab
+        if: always() && hashFiles('osv-results.sarif') != ''
+        uses: github/codeql-action/upload-sarif@v4
+        with:
+          sarif_file: osv-results.sarif
+          category: kubeflow-sdk-osv-scanner
+
+      - name: Run OSV-Scanner (JSON for auto-fix)
+        run: |
+          rc=0
+          osv-scanner scan \
+            --format json \
+            --output osv-results.json \
+            -L uv.lock || rc=$?
+          if [ "$rc" -ne 0 ] && [ "$rc" -ne 1 ]; then
+            echo "::error::OSV-Scanner failed with unexpected exit code $rc"
+            exit 1
+          fi
+          # Validate JSON output is parseable before downstream steps consume it
+          if [ -f osv-results.json ] && [ -s osv-results.json ]; then
+            jq empty osv-results.json
+          fi
+
+      - name: Process vulnerabilities and apply fixes
+        id: fixer
+        run: |
+          if [ ! -f osv-results.json ] || [ ! -s osv-results.json ]; then
+            echo "No scan results found."
+            echo "updates_found=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          CURRENT_DATE=$(date +%Y-%m-%d)
+
+          # Extract fixable vulnerabilities: package name, current version, fix version, advisory ID, URL
+          # The OSV JSON embeds full advisory objects; fixed versions are in affected[].ranges[].events
+          # Use group_by + last to deduplicate: keep the highest fix version per package+advisory pair
+          FIX_DATA=$(jq -r '
+            [
+              .results[]?.packages[]? |
+              .package as $pkg |
+              .vulnerabilities[]? |
+              . as $vuln |
+              .affected[]? |
+              select(.package.name == $pkg.name) |
+              .ranges[]? |
+              .events[] |
+              select(.fixed != null) |
+              {
+                pkg: $pkg.name,
+                ver: $pkg.version,
+                fix: .fixed,
+                id: $vuln.id,
+                url: ($vuln.references[0].url // "https://osv.dev/vulnerability/\($vuln.id)")
+              }
+            ] |
+            group_by([.pkg, .id]) |
+            map(sort_by(.fix | split(".") | map(tonumber? // 0)) | last) |
+            .[] |
+            "\(.pkg)|\(.ver)|\(.fix)|\(.id)|\(.url)"
+          ' osv-results.json 2>/dev/null | sort -u)
+
+          if [ -z "$FIX_DATA" ]; then
+            echo "No fixable vulnerabilities found."
+            echo "updates_found=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          echo "Fixable vulnerabilities found:"
+          echo "$FIX_DATA"
+          echo ""
+
+          rm -f applied_fixes.txt
+
+          while IFS='|' read -r pkg current_ver fix_ver advisory_id advisory_url; do
+            [ -z "$pkg" ] && continue
+
+            echo ""
+            echo "=== Processing: $pkg $current_ver -> $fix_ver ($advisory_id) ==="
+
+            # Try natural upgrade first
+            echo "Attempting: uv lock --upgrade-package $pkg"
+            LOCK_RC=0
+            uv lock --upgrade-package "$pkg" 2>&1 || LOCK_RC=$?
+            if [ "$LOCK_RC" -ne 0 ]; then
+              echo "Warning: uv lock failed for $pkg (exit code $LOCK_RC), skipping"
+              continue
+            fi
+
+            # Check resulting version
+            TREE_RC=0
+            TREE_OUTPUT=$(uv tree --package "$pkg" 2>&1) || TREE_RC=$?
+            if [ "$TREE_RC" -ne 0 ]; then
+              echo "Warning: uv tree failed for $pkg (exit code $TREE_RC), skipping"
+              continue
+            fi
+            EXTRACT_RC=0
+            NATURAL_VER=$(echo "$TREE_OUTPUT" | uv run python3 .github/scripts/extract_version.py "$pkg") || EXTRACT_RC=$?
+            if [ "$EXTRACT_RC" -ne 0 ]; then
+              echo "Warning: Could not parse version for $pkg from uv tree output, skipping"
+              continue
+            fi
+
+            if [ -z "$NATURAL_VER" ]; then
+              echo "Warning: Could not determine version for $pkg after upgrade"
+              continue
+            fi
+
+            echo "Natural upgrade resulted in version: $NATURAL_VER"
+
+            # Compare: does natural version meet the fix?
+            # compare_versions.py exits: 0 = upgrade needed, 1 = sufficient, 2 = parse error
+            CMP_EXIT=0
+            uv run python3 .github/scripts/compare_versions.py "$NATURAL_VER" "$fix_ver" || CMP_EXIT=$?
+
+            if [ "$CMP_EXIT" -eq 2 ]; then
+              echo "Warning: Could not parse versions for $pkg ($NATURAL_VER vs $fix_ver), skipping"
+              continue
+            elif [ "$CMP_EXIT" -eq 0 ]; then
+              # NATURAL_VER < fix_ver — natural upgrade insufficient, add override
+              echo "Natural upgrade insufficient ($NATURAL_VER < $fix_ver). Adding override..."
+              uv run python3 .github/scripts/update_overrides.py "$pkg" "${pkg}==${fix_ver}" "$CURRENT_DATE" "$advisory_url"
+              LOCK_RC=0
+              uv lock --upgrade-package "$pkg" 2>&1 || LOCK_RC=$?
+              if [ "$LOCK_RC" -ne 0 ]; then
+                echo "Warning: uv lock failed after adding override for $pkg (exit code $LOCK_RC), skipping"
+                continue
+              fi
+              printf '%s\n' "- ${pkg}==${fix_ver} (override) | [$advisory_id]($advisory_url)" >> applied_fixes.txt
+            else
+              # NATURAL_VER >= fix_ver — natural upgrade worked
+              echo "Natural upgrade sufficient ($NATURAL_VER >= $fix_ver)"
+              printf '%s\n' "- ${pkg}==${NATURAL_VER} (upgraded) | [$advisory_id]($advisory_url)" >> applied_fixes.txt
+            fi
+          done <<< "$FIX_DATA"
+
+          if [ ! -s applied_fixes.txt ]; then
+            echo "No fixes were applied."
+            echo "updates_found=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          echo "updates_found=true" >> $GITHUB_OUTPUT
+          {
+            echo "fix_details<<EOF"
+            cat applied_fixes.txt
+            echo "EOF"
+          } >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request
+        if: steps.fixer.outputs.updates_found == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "fix: nightly automated dependency update for security vulnerabilities"
+          title: "fix: nightly security dependency updates"
+          add-paths: |
+            pyproject.toml
+            uv.lock
+          body: |
+            ## Security Update
+
+            This is an automated PR triggered by the nightly OSV-Scanner security scan.
+
+            The following dependencies were updated to resolve known vulnerabilities:
+
+            | Package & Version | Advisory Link |
+            | :--- | :--- |
+            ${{ steps.fixer.outputs.fix_details }}
+
+            **Verification:** Updated via `uv lock --upgrade-package` or `override-dependencies`.
+
+            > **Note:** This PR was created with `GITHUB_TOKEN`, which does not trigger `pull_request` workflows.
+            > CI checks will not run automatically. Push an empty commit (`git commit --allow-empty -m "trigger CI"`)
+            > to kick off CI before merging.
+            >
+            > **TODO:** Migrate to a GitHub App token via `actions/create-github-app-token` to trigger CI automatically.
+          branch: security-nightly-updates
+          delete-branch: true
+          labels: |
+            "area/security"
--- a/.github/workflows/validate-lockfile.yaml
+++ b/.github/workflows/validate-lockfile.yaml
@@ -1,0 +1,257 @@
+name: Validate Lockfile Security
+
+on:
+  pull_request:
+    paths:
+      - 'uv.lock'
+      - 'pyproject.toml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  check-security-regressions:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v6
+        with:
+          path: pr-code
+
+      - name: Checkout base branch
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ github.base_ref }}
+          path: base-code
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Verify lockfile is in sync with pyproject.toml
+        run: |
+          cd pr-code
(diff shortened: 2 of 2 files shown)


.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,178 @@
...
+            if uv run python3 .github/scripts/compare_versions.py "$NATURAL_VER" "$fix_ver"; then
+              # NATURAL_VER < fix_ver — natural upgrade insufficient, add override
+              echo "Natural upgrade insufficient ($NATURAL_VER < $fix_ver). Adding override..."
+              uv run python3 .github/scripts/update_overrides.py "$pkg" "${pkg}==${fix_ver}" "$CURRENT_DATE" "$advisory_url"
+              uv lock --upgrade-package "$pkg" 2>&1 || true
+              printf '%s\n' "- ${pkg}==${fix_ver} (override) | [$advisory_id]($advisory_url)" >> applied_fixes.txt
+            else
+              # NATURAL_VER >= fix_ver — natural upgrade worked
+              echo "Natural upgrade sufficient ($NATURAL_VER >= $fix_ver)"
+              printf '%s\n' "- ${pkg}==${NATURAL_VER} (upgraded) | [$advisory_id]($advisory_url)" >> applied_fixes.txt
```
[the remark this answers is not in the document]
Fiona-Waters: Fixed, now captures the exit code explicitly and skips with a warning on exit 2 instead of falling through to the else branch.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,178 @@
...
+        with:
+          enable-cache: true
+
+      - name: Install OSV-Scanner
+        run: |
+          OSV_VERSION="2.3.8"
+          curl -fsSL "https://github.com/google/osv-scanner/releases/download/v${OSV_VERSION}/osv-scanner_linux_amd64" -o osv-scanner
+          chmod +x osv-scanner
+          sudo mv osv-scanner /usr/local/bin/
+
```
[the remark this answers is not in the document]
Fiona-Waters: Fixed, binary SHA-256 is verified against the published checksum. Workflow fails if there's a mismatch.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+          echo "Base branch vulnerabilities: $BASE_COUNT"
+          echo "PR branch vulnerabilities: $PR_COUNT"
+
+          # Find NEW vulns (in PR but not in base) — compare by GHSA|package
+          if [ -n "$PR_VULNS" ]; then
+            BASE_KEYS=$(echo "$BASE_VULNS" | cut -d'|' -f1,2 | sort -u)
+            NEW_VULNS=""
+            while IFS='|' read -r ghsa pkg ver; do
+              KEY="${ghsa}|${pkg}"
+              if ! echo "$BASE_KEYS" | grep -qF "$KEY"; then
```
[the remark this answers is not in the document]
Fiona-Waters: Fixed, changed to grep -qxF for exact line matching.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+      - name: Run OSV-Scanner (SARIF for Security Tab)
+        run: |
+          osv-scanner scan \
+            --format sarif \
+            --output osv-results.sarif \
+            --config osv-scanner.toml \
+            -L uv.lock || true
```
andreyvelich: `|| true` here (and on line 59) collapses scanner crashes (network failure, malformed lockfile, panic) and the documented "vulns found" exit into a single success path — the next step then reads an empty/missing JSON and concludes "no fixes needed". Suggest capturing `rc`, treating only the documented codes as success, and validating with `jq empty osv-results.json` before line 64's check.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+              select(.fixed != null) |
+              {
+                pkg: $pkg.name,
+                ver: $pkg.version,
+                fix: .fixed,
+                id: $vuln.id,
+                url: ($vuln.references[0].url // "https://osv.dev/vulnerability/\($vuln.id)")
+              }
+            ] |
+            group_by(.pkg) |
```
andreyvelich: `group_by(.pkg)` keeps one row **per package**, not per package+advisory as the comment on line 74 implies. Multi-advisory packages silently lose entries from `applied_fixes.txt` and the PR table. Either group by `[.pkg, .id]`, or update the comment to: "one row per package — the highest fix wins; the recorded advisory is whichever shipped that fix."

andreyvelich (commented):
Thanks for the contribution, @Fiona-Waters — solid replacement for the Trivy setup. Review done with AI assistance; comments inline.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+          while IFS='|' read -r pkg current_ver fix_ver advisory_id advisory_url; do
+            [ -z "$pkg" ] && continue
+
+            echo ""
+            echo "=== Processing: $pkg $current_ver -> $fix_ver ($advisory_id) ==="
+
+            # Try natural upgrade first
+            echo "Attempting: uv lock --upgrade-package $pkg"
+            uv lock --upgrade-package "$pkg" 2>&1 || true
```
andreyvelich: `|| true` hides every resolver failure mode (conflict, network, malformed override). The loop then sees the unchanged version and falsely concludes "natural upgrade insufficient", writing a spurious override. Same issue on line 144 — if the post-override re-lock fails, the PR is opened with a stale `uv.lock`. Capture `rc` (cf. `cleanup-overrides.yaml` line ~76) and `continue` on non-zero.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+            # Check resulting version
+            NATURAL_VER=$(uv tree --package "$pkg" 2>/dev/null | uv run python3 .github/scripts/extract_version.py "$pkg" 2>/dev/null || echo "")
```
andreyvelich: Suppressing both `2>/dev/null` and falling through to `|| echo ""` collapses every distinct failure mode (uv tree crash, parse error, transient I/O) into one bland "Could not determine version" warning, then `continue`. Drop the redirects so real failures are visible in the run log; distinguish "package not present" from "tooling failure".

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+            echo "fix_details<<EOF"
+            cat applied_fixes.txt
+            echo "EOF"
+          } >> $GITHUB_OUTPUT
+
+      - name: Create Pull Request
+        if: steps.fixer.outputs.updates_found == 'true'
+        uses: peter-evans/create-pull-request@v8
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
andreyvelich, line 208: PRs created with the default `GITHUB_TOKEN` do **not** trigger downstream `pull_request` workflows — so the auto-PR will appear with no checks (`validate-lockfile`, `test-python`, `test-e2e`) and reviewers may merge without CI. Either pass a GitHub App token via `actions/create-github-app-token`, or document this gap in the auto-PR body so contributors know to push an empty commit to kick CI.

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,191 @@
...
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install OSV-Scanner
+        run: |
+          OSV_VERSION="2.3.8"
+          EXPECTED_SHA="bc98e15319ed0d515e3f9235287ba53cdc5535d576d24fd573978ecfe9ab92dc"
```
andreyvelich, line 30: Binary version + checksum pinning is the right posture, but Dependabot doesn't bump `curl`-installed binaries. Consider a short `# Bump cadence: check https://github.com/google/osv-scanner/releases monthly` comment, or a follow-up issue tracking the manual update.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+            echo ""
+            echo "ERROR: uv.lock is out of sync with pyproject.toml"
+            echo "This means pyproject.toml was modified but 'uv lock' was not run."
+            echo ""
+            echo "To fix this:"
+            echo "  1. Run: uv lock"
+            echo "  2. Commit the updated uv.lock file"
+            echo "  3. Push your changes"
+            echo ""
+            exit 0
```
andreyvelich: `exit 0` after a sync error lets the audit/compare steps run against a known out-of-sync lockfile and report "no new vulnerabilities" against a misleading state — the gate doesn't gate. Either `exit 1`, or set a step output and add `if: steps.<id>.outputs.in_sync == 'true'` to the audit/compare steps.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+          echo "Lockfile is in sync"
+
+      - name: Audit PR branch lockfile
+        run: |
+          cd pr-code
+          # NOTE: uv audit is experimental (preview feature) and its text output format may
+          # change without notice. If the parsing in the compare step breaks after a uv upgrade,
+          # check the output format or consider switching to osv-scanner JSON output.
+          # Both tools query the same OSV.dev database — coverage is identical.
+          uv audit --preview-features audit 2>&1 | tee ../audit-pr.txt || true
```
andreyvelich: `uv audit … || true` (here and on line 68) hides tool failures; an empty output yields `BASE_COUNT=0`/`PR_COUNT=0` and stamps "no new vulnerabilities" on a PR where the audit just crashed. Capture `${PIPESTATUS[0]}` and, if non-zero AND the file lacks any vuln-format header, emit `::error::` and exit 1.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+          fi
+          echo "Lockfile is in sync"
+
+      - name: Audit PR branch lockfile
+        run: |
+          cd pr-code
+          # NOTE: uv audit is experimental (preview feature) and its text output format may
+          # change without notice. If the parsing in the compare step breaks after a uv upgrade,
+          # check the output format or consider switching to osv-scanner JSON output.
+          # Both tools query the same OSV.dev database — coverage is identical.
```
andreyvelich: Two suggestions: (a) soften "coverage is identical" — both query OSV.dev, but filtering and version-range logic differ between `uv audit` (preview) and OSV-Scanner; (b) since the parser is fragile against the experimental `uv audit` format, consider running `osv-scanner --format json -L uv.lock` here too — same data, structured parsing, single source of format risk shared with the nightly job.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+            # Build markdown table
+            echo "new_cves<<EOF" >> $GITHUB_OUTPUT
+            echo "| Advisory ID | Package | Version |" >> $GITHUB_OUTPUT
+            echo "| :--- | :--- | :--- |" >> $GITHUB_OUTPUT
+
+            while IFS='|' read -r ghsa pkg ver; do
+              [ -z "$ghsa" ] && continue
+              # Look up advisory URL from the PR audit output
+              ADVISORY=$(grep -A3 "$ghsa" audit-pr.txt | grep -oP 'https://\S+' | head -1 || true)
```
andreyvelich: `grep -A3 "$ghsa"` matches the GHSA anywhere in the file (including descriptions of unrelated advisories), then takes the first URL within 3 lines, which may belong to another package's block. Anchor on `^- ${ghsa}:` and parse the next URL line, or pull the URL from JSON output if you adopt the `osv-scanner --format json` suggestion above.

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,230 @@
...
+
+            echo "EOF" >> $GITHUB_OUTPUT
+            exit 0
+          else
+            echo "No new vulnerabilities introduced by this PR"
+            echo "new_cves_found=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Comment on PR with regression details
+        continue-on-error: true
```
andreyvelich, line 166: `continue-on-error: true` on the only mechanism that surfaces a regression to the developer: if the comment API call fails (rate limit, transient 5xx, fork permissions), the alert is silently dropped. Keep the flag for fork PRs, but add a fallback that writes the regression to `$GITHUB_STEP_SUMMARY` and emits a `::warning::` annotation so the alert is visible regardless.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Fixed. Both scan steps now capture the exit code and only treat 0 (clean) and 1 (vulns found) as success — any other code surfaces an ::error annotation and fails the step. The SARIF step uses continue-on-error: true so a SARIF-specific failure doesn't block the JSON scan and auto-fix pipeline. Also added jq empty validation on the JSON output before downstream consumption.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Fixed. Changed to group_by([.pkg, .id]) so each package+advisory pair is preserved.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Fixed. Both uv lock calls now capture the exit code and continue to the next package on non-zero, matching the pattern in cleanup-overrides.yaml. Error output is preserved in the log.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Fixed. Removed the stderr suppression and split into two explicit checks — one for uv tree failure and one for version extraction failure — each with a descriptive warning and continue. Uses the rc=0; cmd || rc=$? pattern to stay safe under GitHub Actions' default set -e.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Documented the gap in the auto-PR body with instructions to push an empty commit to trigger CI. Added a TODO noting the long-term fix is migrating to a GitHub App token via actions/create-github-app-token. Maybe we can create a follow on issue for this?

.github/workflows/osv-scanner.yaml
Fiona-Waters: Added a comment on the install step noting monthly bump cadence and that Dependabot can't track curl-installed binaries.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Fixed. Changed to exit 1 and upgraded the annotation from ::warning to ::error. The job now fails when the lockfile is out of sync, preventing the audit steps from running against a misleading lockfile state.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Fixed. Replaced | tee file || true with > file 2>&1 || rc=$? followed by cat — this correctly captures uv audit's exit code (the pipe+tee pattern would have captured tee's exit code instead). Exit code 1 (vulns found) is treated as expected; anything else fails the step with ::error.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Softened the comment to note that while both tools query the same OSV.dev database, filtering and version-range logic differ between uv audit (preview) and OSV-Scanner, so results may not be identical.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Fixed. Changed to grep -P "^- ${ghsa}:" to anchor the match to the start of a GHSA line, preventing false matches from descriptions of unrelated advisories.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Fixed. Added id: comment to the PR comment step and a new fallback step that fires when steps.comment.outcome == 'failure'. The fallback writes the regression details to $GITHUB_STEP_SUMMARY and emits a ::warning annotation, so the alert is visible even if the API call fails.

Fiona-Waters:
> Thanks for the contribution, @Fiona-Waters — solid replacement for the Trivy setup. Review done with AI assistance; comments inline.

Thanks for the review @andreyvelich I've addressed all comments, please re-review. Thanks

osv-scanner.toml
```diff
@@ -0,0 +1,9 @@
+# OSV-Scanner configuration
```
andreyvelich: If this is file empty, do we still need it?

Fiona-Waters: Its a placeholder that could be used later for suppressing false positives or disputed advisories - but I can delete it for now and we can add it back in the future if needed. wdyt?

andreyvelich: Let's add it in the future once we need them if that sounds good @Fiona-Waters.

Fiona-Waters: Removed in latest commit. Thanks

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,257 @@
...
+            local current_ver=""
+            while IFS= read -r line; do
+              # Match package header: "aiohttp 3.13.3 has N known vulnerabilities:"
+              if echo "$line" | grep -qP '^[a-zA-Z].*has \d+ known vulnerabilit'; then
+                current_pkg=$(echo "$line" | awk '{print $1}')
+                current_ver=$(echo "$line" | awk '{print $2}')
+              fi
+              # Match GHSA line: "- GHSA-xxxx-xxxx-xxxx: description"
+              if echo "$line" | grep -qP '^\- GHSA-'; then
+                local ghsa=$(echo "$line" | grep -oP 'GHSA-[a-z0-9-]+')
```
kramaranya: Why do we only catch GHSA prefixed ids?

andreyvelich (commented):
Thanks @Fiona-Waters!
Could you also add it to the Trainer project too!
/lgtm
/approve

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+                fix: .fixed,
+                id: $vuln.id,
+                url: ($vuln.references[0].url // "https://osv.dev/vulnerability/\($vuln.id)")
+              }
+            ] |
+            group_by([.pkg, .id]) |
+            map(sort_by(.fix) | last) |
+            .[] |
+            "\(.pkg)|\(.ver)|\(.fix)|\(.id)|\(.url)"
+          ' osv-results.json 2>/dev/null | sort -u)
```
kramaranya: would this sorting pick the correct fix version for example with 1.9.0 and 1.10.0

.github/workflows/validate-lockfile.yaml
```diff
@@ -0,0 +1,257 @@
...
+      - name: Audit PR branch lockfile
+        run: |
+          cd pr-code
+          # NOTE: uv audit is experimental (preview feature) and its text output format may
+          # change without notice. If the parsing in the compare step breaks after a uv upgrade,
+          # check the output format or consider switching to osv-scanner JSON output.
+          # Both tools query the same OSV.dev database, but filtering and version-range logic
+          # differ between uv audit (preview) and OSV-Scanner — results may not be identical.
+          rc=0
+          uv audit --preview-features audit > ../audit-pr.txt 2>&1 || rc=$?
```
kramaranya: is this correct or a typo?

.github/workflows/osv-scanner.yaml
```diff
@@ -0,0 +1,235 @@
...
+            ${{ steps.fixer.outputs.fix_details }}
+
+            **Verification:** Updated via `uv lock --upgrade-package` or `override-dependencies`.
+
+            > **Note:** This PR was created with `GITHUB_TOKEN`, which does not trigger `pull_request` workflows.
+            > CI checks will not run automatically. Push an empty commit (`git commit --allow-empty -m "trigger CI"`)
+            > to kick off CI before merging.
+            >
+            > **TODO:** Migrate to a GitHub App token via `actions/create-github-app-token` to trigger CI automatically.
+          branch: security-nightly-updates-${{ github.run_id }}
```
kramaranya: this could spam with prs, can we use a fixed branch name so it updates the existing pr instead?

kramaranya (changes requested):

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Updated to match any advisory ID prefix.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Fixed to split and sort numerically.

.github/workflows/validate-lockfile.yaml
Fiona-Waters: Thanks for catching, its a typo. Removed now.

.github/workflows/osv-scanner.yaml
Fiona-Waters: Updated to remove the run_id and use `security-nightly-updates` fixed branch.

kramaranya (commented):
Looks good to me!
@Fiona-Waters could you please squash your commits?

Fiona-Waters:
> Looks good to me! @Fiona-Waters could you please squash your commits?

Done. Thanks @kramaranya 🥳

kramaranya:
Thank you, @Fiona-Waters!
/lgtm

PR #504, opened 2026-05-26
fix(trainer): resolve E2E log streaming incompatibility with kubernetes client v35+

HKanoje:
Fix log streaming to properly yield line-terminated chunks instead of raw 64KB byte buffers, maintaining Iterator[str] contract for caller code.

## Changes
- Replace watch.Watch().stream() with response.stream() for pod log streaming
- Pass _preload_content=False to read_namespaced_pod_log when following logs
- Remove unused watch import that caused ApiTypeError in newer kubernetes versions
- Update mock_read_namespaced_pod_log to support streaming responses
- Add parameterized test case for get_job_logs with follow=True

## Issue
Fixes the RuntimeError in E2E tests: 'Got an unexpected keyword argument watch to method read_namespaced_pod_log' which occurs with kubernetes client >= 35.0.0

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -25,7 +25,7 @@ from typing import Any
 import uuid
 
 from kubeflow_trainer_api import models
-from kubernetes import client, config, watch
+from kubernetes import client, config
 
 import kubeflow.common.constants as common_constants
 from kubeflow.common.types import KubernetesBackendConfig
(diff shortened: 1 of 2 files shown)


Fiona-Waters:
This issue is being tracked upstream in kubernetes python client - https://github.com/kubernetes-client/python/issues/2596#issuecomment-4540544379
We should wait for this fix, or we could pin kubernetes<36 until it's fixed
wdyt @andreyvelich @tariq-hasan ?

tariq-hasan:
There's no definite timeline yet for a patch release for the bug.

There has been a recent 36.0.1 which has been breaking the tests again because it does not address the logging error.

I suppose best to pin <36.0.0.

SahilKumar75:
This has been inactive since late May, consolidating the open items:

* Direction: @Fiona-Waters and @tariq-hasan suggested pinning `kubernetes<36` and waiting on the upstream client fix (kubernetes-client/python#2596) rather than reworking the streaming logic. Is this PR still the intended fix, or should it be superseded by a version pin?
* Open review points: `response.stream()` yields arbitrary byte chunks rather than lines, so decoding and splitting per chunk can merge or split log lines; the `follow=True` stream is not explicitly closed, which can leak a connection if the caller stops early; and the tests only cover a single chunk with no newlines.

@HKanoje are you still working on this? Happy to help move it along.

PR #509, opened 2026-05-26
fix(ci): fix --preview-features flag usage in uv audit

Fiona-Waters:

  The `validate-lockfile.yaml` workflow fails on every run with exit code 2 because
  `uv audit` is invoked with `--preview-features` as a bare flag. The flag requires
  a value naming the feature to opt into: `--preview-features audit`.

  ## Changes

  - Fixed both `uv audit` invocations to use `--preview-features audit`
  - Added comment explaining why `audit` is needed after `--preview-features`

  ## How to verify

  The workflow should now pass on PRs that modify `uv.lock` or `pyproject.toml`.

Fixes #

**Checklist:**

### DIFF
--- a/.github/workflows/validate-lockfile.yaml
+++ b/.github/workflows/validate-lockfile.yaml
@@ -56,13 +56,15 @@ jobs:
       - name: Audit PR branch lockfile
         run: |
           cd pr-code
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
Thanks @Fiona-Waters!
/lgtm
/approve

PR #518, opened 2026-06-02
fix(spark): align Spark Connect service-name fallback with operator default

### LINKED ISSUE #515: KEP-107: Mismatch between SparkConnect SDK and operator service name fallback
### What happened?

The Spark SDK [`build_service_url`](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/utils.py#L74-L84) helper falls back to `f"{info.name}-svc"` when `info.service_name` is unset. The Spark Operator's [`GetServerServiceName`](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkconnect/util.go#L66-L76) helper defaults to `f"{conn.Name}-server"`. The two suffixes don't match, so the fallback produces a hostname that has no corresponding Service in the cluster.

In typical flows the fallback isn't hit because the SDK reads `service_name` from the SparkConnect CR status, which the operator populates. But this path may be hit during initial reconciliation before status is populated. In that case, the fallback would construct `<name>-svc.<namespace>.svc.cluster.local` and that won't resolve. The resulting error would surface as a DNS resolution failure and be hard to attribute to this line.

The out-of-cluster `get_connect_url` path in `backend.py` already tries [`f"{info.name}-svc"`, `info.service_name`, `f"{info.name}-server"`] as [candidate names](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend.py#L338), so the codebase elsewhere acknowledges that `-server` is a valid name - the in-cluster path doesn't.

The fallback should construct the operator's default service name, `f"{info.name}-server"`, so the URL is resolvable on the race path. The alternative is to have the SDK wait for `Status.Server.ServiceName` to be populated before calling `build_service_url`, and remove the fallback entirely.

In addition, the `f"{info.name}-svc"` candidate in the out-of-cluster `get_connect_url` path in `backend.py` appears to be inconsistent with the operator’s default naming and should be removed.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

Kartikeya-trivedi:
*What this PR does / why we need it*:

The in-cluster fallback built `<name>-svc`, but the operator names the service `<name>-server`, so the URL didn't resolve when `service_name` wasn't populated yet (the reconciliation race). Switched the fallback to `-server`, and also dropped the stale `-svc` candidate from the out-of-cluster port-forward list in `backend.py`.

Fixes #515

### DIFF
--- a/kubeflow/spark/backends/kubernetes/backend.py
+++ b/kubeflow/spark/backends/kubernetes/backend.py
@@ -448,6 +448,11 @@ class KubernetesBackend(RuntimeBackend):
         Returns:
             (connect_url, port_forward_process or None). Caller may keep process reference;
             process exits when the Python process exits.
+
+        Raises:
+            RuntimeError: If the session reports no port-forward target, if
+                build_service_url cannot resolve an in-cluster host, or if
+                port-forward fails for every candidate.
         """
         if os.environ.get("KUBERNETES_SERVICE_HOST"):
             url = build_service_url(info)
@@ -457,19 +462,20 @@ class KubernetesBackend(RuntimeBackend):
         if port is None:
             port_str = os.environ.get("SPARK_CONNECT_LOCAL_PORT")
             port = int(port_str) if port_str else random.randint(15002, 16002)
-        # Prefer pod when available (bypasses Service/EndpointSlice); then try svc names
+        # Prefer pod when available (bypasses Service/EndpointSlice); then try svc name
         candidates: list[tuple[str, str]] = []
         if info.driver_pod_name:
             candidates.append(("pod", info.driver_pod_name))
-        for svc in [f"{info.name}-svc", info.service_name, f"{info.name}-server"]:
-            if svc and not any(c[0] == "svc" and c[1] == svc for c in candidates):
-                candidates.append(("svc", svc))
-        seen: set[str] = set()
+        if info.service_name:
+            candidates.append(("svc", info.service_name))
+        if not candidates:
+            raise RuntimeError(
+                f"No port-forward target for {info.namespace}/{info.name}: neither "
+                "status.server.podName nor status.server.serviceName is populated. "
+                "The session is not ready."
+            )
         for kind, target in candidates:
             key = f"{kind}/{target}"
-            if key in seen:
-                continue
-            seen.add(key)
             # Use 127.0.0.1 instead of localhost to force IPv4 (gRPC may prefer IPv6 which can fail)
             url = f"sc://127.0.0.1:{port}"
             cmd = [
--- a/kubeflow/spark/backends/kubernetes/backend_test.py
+++ b/kubeflow/spark/backends/kubernetes/backend_test.py
@@ -744,6 +744,20 @@ def test_get_connect_url(kubernetes_backend, test_case):
     print("test execution complete")
 
 
+def test_get_connect_url_raises_when_no_target(kubernetes_backend):
+    """get_connect_url raises when neither pod_name nor service_name is populated."""
+    info = SparkConnectInfo(
+        name="test-session",
+        namespace="default",
+        state=SparkConnectState.READY,
+    )
+    with (
+        patch.dict("os.environ", {"KUBERNETES_SERVICE_HOST": ""}, clear=False),
+        pytest.raises(RuntimeError, match="No port-forward target"),
+    ):
+        kubernetes_backend.get_connect_url(info)
+
+
 @pytest.mark.parametrize(
     "test_case",
     [
--- a/kubeflow/spark/backends/kubernetes/backend_test.py
+++ b/kubeflow/spark/backends/kubernetes/backend_test.py
@@ -698,16 +698,23 @@ def test_get_session_logs(kubernetes_backend, test_case):
             config={"in_cluster": False},
             expected_output={"url": "sc://127.0.0.1:15002", "proc_is_none": False},
         ),
+        TestCase(
+            name="out-of-cluster without pod or service name raises",
+            expected_status=FAILED,
+            config={"in_cluster": False, "service_name": None},
+            expected_error=RuntimeError,
+            expected_output="No port-forward target",
+        ),
     ],
 )
 def test_get_connect_url(kubernetes_backend, test_case):
-    """Test get_connect_url for in-cluster and port-forward scenarios."""
+    """Test get_connect_url for in-cluster, port-forward, and not-ready scenarios."""
     print("Executing test:", test_case.name)
     info = SparkConnectInfo(
         name="test-session",
         namespace="default",
         state=SparkConnectState.READY,
-        service_name="test-session-svc",
+        service_name=test_case.config.get("service_name", "test-session-server"),
     )
 
     if test_case.config["in_cluster"]:
@@ -729,6 +736,14 @@ def test_get_connect_url(kubernetes_backend, test_case):
             patch("kubeflow.spark.backends.kubernetes.backend.time.sleep"),
             patch.object(kubernetes_backend, "_wait_for_connect_port", return_value=True),
         ):
+            if test_case.expected_status == FAILED:
+                with pytest.raises(
+                    test_case.expected_error,
+                    match=test_case.expected_output,
+                ):
+                    kubernetes_backend.get_connect_url(info)
+                print("test execution complete")
+                return
             url, proc = kubernetes_backend.get_connect_url(info)
 
     if "url_contains" in test_case.expected_output:
@@ -742,20 +757,6 @@ def test_get_connect_url(kubernetes_backend, test_case):
         assert proc is mock_popen
 
     print("test execution complete")
-
-
-def test_get_connect_url_raises_when_no_target(kubernetes_backend):
-    """get_connect_url raises when neither pod_name nor service_name is populated."""
-    info = SparkConnectInfo(
-        name="test-session",
-        namespace="default",
-        state=SparkConnectState.READY,
-    )
-    with (
-        patch.dict("os.environ", {"KUBERNETES_SERVICE_HOST": ""}, clear=False),
-        pytest.raises(RuntimeError, match="No port-forward target"),
-    ):
-        kubernetes_backend.get_connect_url(info)
 
 
 @pytest.mark.parametrize(
--- a/kubeflow/spark/backends/kubernetes/utils.py
+++ b/kubeflow/spark/backends/kubernetes/utils.py
@@ -328,16 +328,37 @@ def validate_spark_connect_url(url: str) -> bool:
 
 
 def build_service_url(info: SparkConnectInfo) -> str:
-    """Build Spark Connect URL from session info.
+    """Build the in-cluster Spark Connect URL from session info.
+
+    Relies on the operator-reported Service name. The operator records the
+    actual Service name (default "<name>-server" or a user-customized name via
+    .spec.server.service) in Status.Server.ServiceName, so this is correct for
+    both default and customized Services. Callers are expected to have waited
+    for readiness (wait_for_ready), by which point the operator has populated
+    this field in the same atomic status write that set the Ready state.
 
     Args:
         info: SparkConnectInfo with service details.
 
     Returns:
-        Spark Connect URL (e.g., "sc://service-name:15002").
+        Spark Connect URL
+        (e.g., "sc://my-session-server.default.svc.cluster.local:15002").
+
+    Raises:
+        RuntimeError: If ``info.service_name`` is not populated. An empty value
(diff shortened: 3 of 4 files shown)


Kartikeya-trivedi:
Hi @tariq-hasan, gentle ping on this, it addresses issue #515 and CI is green. No rush, but a review whenever you have a moment would be appreciated.Thanks !!

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan: We should use the following based on the observation in the other comment.

```suggestion
        # Prefer pod when available (bypasses Service/EndpointSlice); then try svc name
        candidates: list[tuple[str, str]] = []
        if info.pod_name:
            candidates.append(("pod", info.pod_name))
        if info.service_name:
            candidates.append(("svc", info.service_name))
        if not candidates:
            raise RuntimeError(
                f"No port-forward target for {info.namespace}/{info.name}: neither "
                "status.server.podName nor status.server.serviceName is populated. "
                "The session is not ready."
            )
        for kind, target in candidates:
            key = f"{kind}/{target}"
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -80,7 +80,10 @@ def build_service_url(info: SparkConnectInfo) -> str:
     Returns:
         Spark Connect URL (e.g., "sc://service-name:15002").
     """
-    service = info.service_name or f"{info.name}-svc"
+    # Fallback matches the operator's default Service name (GetServerServiceName:
+    # "<name>-server") so the URL resolves even if Status.Server.ServiceName is not
+    # yet populated (e.g. a race during initial reconciliation).
+    service = info.service_name or f"{info.name}-server"
     return f"sc://{service}.{info.namespace}.svc.cluster.local:{constants.SPARK_CONNECT_PORT}"
```
tariq-hasan: Thinking through this a bit more the actual change should be

```suggestion
def build_service_url(info: SparkConnectInfo) -> str:
    """Build the in-cluster Spark Connect URL from session info.

    Relies on the operator-reported Service name. The operator records the
    actual Service name (default "<name>-server" or a user-customized name via
    .spec.server.service) in Status.Server.ServiceName, so this is correct for
    both default and customized Services. Callers are expected to have waited
    for readiness (wait_for_ready), by which point the operator has populated
    this field in the same atomic status write that set the Ready state.

   Args:
       info: SparkConnectInfo with service details.

    Returns:
        Spark Connect URL
        (e.g., "sc://my-session-server.default.svc.cluster.local:15002").

    Raises:
        RuntimeError: If ``info.service_name`` is not populated. An empty value
            means the session is not ready yet, not a name we should guess;
            callers should wait for readiness before building the URL.
    """
    if not info.service_name:
        raise RuntimeError(
            f"Cannot build Spark Connect URL for {info.namespace}/{info.name}: "
            "status.server.serviceName is not populated yet. The session is not "
            "ready to accept in-cluster connections."
        )
    return (
        f"sc://{info.service_name}.{info.namespace}.svc.cluster.local"
        f":{constants.SPARK_CONNECT_PORT}"
    )
```

Unlike what the [issue](https://github.com/kubeflow/sdk/issues/515#issue-4555954428) claims I don't think there's a race since a connection is only ever [attempted](https://github.com/Kartikeya-trivedi/sdk/blob/fix/spark-connect-service-name-fallback/kubeflow/spark/backends/kubernetes/backend.py#L580-L583) once the server status transitions to `READY`. And the control plane only [updates](https://github.com/kubeflow/spark-operator/blob/master/internal/controller/sparkconnect/reconciler.go#L226-L236) the status on the API server once a service with a definite name (either user-provided or default) is there already.

Since the API server is the source of truth and `info.service_name` [reads](https://github.com/Kartikeya-trivedi/sdk/blob/112b2ba250b4a8c6f37cc85de777cdc77806ed5c/kubeflow/spark/backends/kubernetes/utils.py#L295-L308) the persisted status from the API server then `info.service_name` already contains either the user-provided name or the `-server` default. The ` or f"{info.name}-server"` fallback on the SDK can never be reached.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -97,14 +97,15 @@ def test_build_from_session_info(self):
...
+        """Build URL when service_name is None falls back to the operator default."""
         info = SparkConnectInfo(
             name="my-session",
             namespace="default",
             state=SparkConnectState.READY,
         )
         url = build_service_url(info)
-        assert "my-session-svc" in url
+        # Fallback must match the operator's GetServerServiceName default ("<name>-server").
+        assert "my-session-server" in url
```
tariq-hasan, line 243: The tests need to be updated to reflect the code changes.

tariq-hasan (changes requested):
Hi @Kartikeya-trivedi! Thanks for raising the PR. I have left a few comments.

kubeflow/spark/backends/kubernetes/backend.py
Kartikeya-trivedi: Done, switched to pod/service_name only and added a raise + test for the no-target case

kubeflow/spark/backends/kubernetes/utils.py
Kartikeya-trivedi: Makes sense since no race since status flips to READY only after the service exists and a guess would break custom service names anyway. Removed the fallback, it now raises if it's empty.

kubeflow/spark/backends/kubernetes/utils_test.py
Kartikeya-trivedi: Updated the test to assert the raise and added one for the backend no-target path too.

Kartikeya-trivedi:
Hi @tariq-hasan I made change as requested ie removed the SDK-side name guessing entirely, rely on the operator-reported ServiceName, raise when it's absent, tests updated on top of that I also added one test for the backend no-target path too. Thanks for the review!!

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan, line 456:
```suggestion

       Returns:
           (connect_url, port_forward_process or None). Caller may keep process reference;
           process exits when the Python process exits.

        Raises:
            RuntimeError: If the session reports no port-forward target, if
                build_service_url cannot resolve an in-cluster host, or if
                port-forward fails for every candidate.
       """
```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -96,15 +96,15 @@ def test_build_from_session_info(self):
         url = build_service_url(info)
         assert url == "sc://my-session-svc.spark.svc.cluster.local:15002"
```
tariq-hasan: Let's have tests for both default and custom cases.

```suggestion
    def test_build_from_operator_default_service_name(self):
        """Build service URL from the operator's default Service name ("<name>-server")."""
        info = SparkConnectInfo(
            name="my-session",
            namespace="spark",
            state=SparkConnectState.READY,
            service_name="my-session-server",
        )
        url = build_service_url(info)
        assert url == "sc://my-session-server.spark.svc.cluster.local:15002"

    def test_build_from_custom_service_name(self):
        """Build service URL from a user-customized Service name."""
        info = SparkConnectInfo(
            name="my-session",
            namespace="spark",
            state=SparkConnectState.READY,
            service_name="custom-endpoint",
        )
        url = build_service_url(info)
        assert url == "sc://custom-endpoint.spark.svc.cluster.local:15002"

```

tariq-hasan (commented):
Hi @Kartikeya-trivedi! Apologies for the delay. I have added a few additional comments.

kubeflow/spark/backends/kubernetes/backend.py
Kartikeya-trivedi: Added the Raises: section and  verified that all three paths actually raise

kubeflow/spark/backends/kubernetes/utils_test.py
Kartikeya-trivedi: Done added separate tests for the operator-default (-server) and custom service names. Thanks!!

Kartikeya-trivedi:
thanks for the review, @tariq-hasan! Addressed both comments and did suggested changes, DCO's green and the spark tests pass locally. Ready for another look when you get a chance 🙏

tariq-hasan:
Hi @Kartikeya-trivedi! Please rebase the PR to address the merge conflicts. Changes look good to me otherwise.

/cc @andreyvelich @Shekharrajak @Goku2099

Kartikeya-trivedi:
Rebased onto latest main and squashed to a single commit, @tariq-hasan. While rebasing I also had to update the parametrized build_service_url test that came in with the KEP-107 changes,its "without service name" case still expected the old fallback URL, so it now asserts the RuntimeError instead, alongside the operator-default and custom service-name cases. Spark tests pass locally. Ready for another look, thanks!

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -744,6 +744,20 @@ def test_get_connect_url(kubernetes_backend, test_case):
...
+        namespace="default",
+        state=SparkConnectState.READY,
+    )
+    with (
+        patch.dict("os.environ", {"KUBERNETES_SERVICE_HOST": ""}, clear=False),
+        pytest.raises(RuntimeError, match="No port-forward target"),
+    ):
+        kubernetes_backend.get_connect_url(info)
+
+
```
tariq-hasan: Let's remove this.

```suggestion
```

kubeflow/spark/backends/kubernetes/backend_test.py
tariq-hasan, line 709: Let's add the test case here.

```suggestion
@pytest.mark.parametrize(
   "test_case",
   [
       TestCase(
           name="in-cluster returns svc URL and no process",
           expected_status=SUCCESS,
           config={"in_cluster": True},
           expected_output={"url_contains": "svc.cluster.local", "proc_is_none": True},
       ),
       TestCase(
           name="out-of-cluster starts port-forward and returns localhost URL",
           expected_status=SUCCESS,
           config={"in_cluster": False},
           expected_output={"url": "sc://127.0.0.1:15002", "proc_is_none": False},
       ),
        TestCase(
            name="out-of-cluster without pod or service name raises",
            expected_status=FAILED,
            config={"in_cluster": False, "service_name": None},
            expected_error=RuntimeError,
            expected_output="No port-forward target",
        ),
   ],
)
```

tariq-hasan:
```suggestion
    """Test get_connect_url for in-cluster, port-forward, and not-ready scenarios."""
    print("Executing test:", test_case.name)
    info = SparkConnectInfo(
        name="test-session",
        namespace="default",
        state=SparkConnectState.READY,
        service_name=test_case.config.get("service_name", "test-session-server"),
    )
```

tariq-hasan, line 748: Let's update the logic given the new test case.


```suggestion
            if test_case.expected_status == FAILED:
                with pytest.raises(
                    test_case.expected_error,
                    match=test_case.expected_output,
                ):
                    kubernetes_backend.get_connect_url(info)
                print("test execution complete")
                return

            url, proc = kubernetes_backend.get_connect_url(info)

```

tariq-hasan (commented):
Thanks @Kartikeya-trivedi! This test needs some updates to match the `pytest` test case parameterization pattern in the project.

Kartikeya-trivedi:
Thanks for hard carrying me @tariq-hasan, applied all four, the not-ready case is now part of the test_get_connect_url parameterization. Tests pass locally.

tariq-hasan:
Thanks @Kartikeya-trivedi!
/lgtm
/assign @andreyvelich

Goku2099:
Looks good overall 
One question: backend.py used to de-dupe candidates via a seen set, which is removed here. Since candidates can now only ever have one "pod" and one "svc" entry, I think that's safe, but wanted to confirm that removal was intentional rather than incidental.

Kartikeya-trivedi:
it was intentional  with at most one pod and one svc candidate, duplicates can't happen. Worth noting seen was actually already dead code before this change: the old svc loop deduped inline via the not any(...) guard, and keys are prefixed by kind, so pod/x and svc/x never collide. cc @Goku2099

andreyvelich (commented):
Thank you for this @Kartikeya-trivedi!
/lgtm
/approve

PR #521, opened 2026-06-03
feat: [KEP-107] Add Spark batch job submission and lifecycle APIs

Goku2099:
## Summary

This PR implements the Phase 1 batch job APIs from KEP-107 by adding Spark batch job submission and lifecycle management through the Spark Operator.

The implementation follows the final KEP design and introduces the FileJob-based submission model together with lifecycle APIs for monitoring and managing SparkApplication resources.

## Features Added

- Introduce `FileJob` as the Phase 1 batch job specification.
- Implement `SparkClient.submit_job()` for Spark batch job submission.
- Add lifecycle APIs:
  - `get_job()`
  - `list_jobs()`
  - `wait_for_job_status()`
  - `get_job_logs()`
  - `delete_job()`
- Add `SparkJob` model following KEP-107.
- Add Simplified SDK job states:
  - `CREATED`
  - `RUNNING`
  - `COMPLETED`
  - `FAILED`
- Map SparkApplication states to SDK lifecycle states.
- Validate FileJob inputs.
- Support remote application URIs and cluster-accessible local application paths.
- Add backend, utility, and SDK unit tests.

## Tested

Validated on a Kind cluster with the Spark Operator installed.

Verified:

- Batch job submission using remote application URIs.
- Job status retrieval (`get_job`).
- Job listing (`list_jobs`).
- Waiting for terminal job states (`wait_for_job_status`).
- Driver log retrieval (`get_job_logs`).
- Job deletion (`delete_job`).
- SparkApplication generation.
- Executor resource configuration.
- SparkApplication → SDK status mapping.
- Backend, utility, API, and SDK unit tests.

## Future Work

Future phases will add:

- Function-based batch jobs (`FuncJob`).
- `spark_conf` support.
- Options pattern for advanced SparkApplication configuration.

ref: #524


Goku2099:
 I have updated the PR to include the revised submit_job implementation and the lifecycle APIs. Everything is complete now and it's ready for review.

andreyvelich (commented):
Thanks for the great work on this, @Goku2099! 🙏

This review was performed using AI-assisted tooling. I've left inline comments on specific lines with suggestions — please take a look.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +328,218 @@ def get_spark_connect_info_from_cr(
...
+    if cr.spec and cr.spec.executor:
+        num_executors = cr.spec.executor.instances
+
+    if cr.status:
+        if cr.status.application_state:
+            raw_state = cr.status.application_state.state or ""
+
+            status = SPARK_APPLICATION_STATUS_MAPPING.get(
+                raw_state,
+                SparkJobStatus.CREATED,
```
andreyvelich: Unmapped SparkApplication states default to `CREATED`, so a job in a future/renamed operator state would look healthy and `wait_for_job_status` would hang until timeout. KEP-107 maps `UNKNOWN → FAILED`. Suggest defaulting unmapped states to `FAILED` and adding a `logger.warning(...)` so the mapping gap is visible. Keep the no-status-block case as `CREATED`.

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,130 @@ def delete_session(self, name: str) -> None:
...
+        """List Spark jobs."""
+        return self.backend.list_jobs(status=status)
+
+    def delete_job(self, name: str) -> None:
+        return self.backend.delete_job(name)
+
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: set[SparkJobStatus],
```
andreyvelich: `status` is required here, but KEP-107 defines `status: Set[SparkJobStatus] = {SparkJobStatus.COMPLETED}` and its examples call `wait_for_job_status(job_name)` without it → currently raises `TypeError`. `TrainerClient.wait_for_job_status` also defaults `status={TRAINJOB_COMPLETE}`. Please add a default (use a `None` sentinel or `frozenset`, not a mutable `{...}` default) here + in base.py + backend.py. Also consider exposing `polling_interval` on the public client, as TrainerClient does.

kubeflow/spark/types/types.py
```diff
@@ -113,3 +115,55 @@ class Executor:
     num_instances: int | None = None
     resources_per_executor: dict[str, str] | None = None
     java_options: str | None = None
+
+
+class SparkJobStatus(str, Enum):
```
andreyvelich, line 120: `SparkJobStatus`, `SparkJob`, `FileJob`, and `FuncJob` are public API (users need them for `submit_job`/`list_jobs`/`wait_for_job_status`) but aren't exported in `kubeflow/spark/__init__.py __all__` alongside `Driver`/`Executor`. Please add them so `from kubeflow.spark import FileJob` works.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +636,331 @@ def get_session_logs(
...
+
+            response = thread.get(common_constants.DEFAULT_TIMEOUT)
+
+            spark_application = models.SparkV1beta2SparkApplication.from_dict(response)
+
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(f"Timeout to get SparkApplication: {self.namespace}/{name}") from e
+
+        except Exception as e:
+            raise RuntimeError(f"Failed to get SparkApplication: {self.namespace}/{name}") from e
```
andreyvelich: Unlike its siblings `delete_job` and the existing `get_session`/`delete_session`, `get_job` doesn't special-case a 404 — a mistyped/deleted job name collapses into the same generic error as a real API/RBAC failure. Please add an `except client.ApiException` branch raising a clear "not found" message for `e.status == 404`, and consider appending `: {e}` (as `submit_job` does).

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,130 @@ def delete_session(self, name: str) -> None:
     def get_session_logs(self, name: str, follow: bool = False) -> Iterator[str]:
         """Get logs from a session."""
         return self.backend.get_session_logs(name, follow=follow)
+
+    def submit_job(
```
andreyvelich, line 174: The client-layer guards here (FuncJob→NotImplementedError, non-FileJob→TypeError, empty `file_source`→ValueError, `spark_conf`/`options`→NotImplementedError) and the `.name` extraction are real logic, not a pass-through, and aren't covered by any test (spark_client_test.py only tests connect). Please add tests for these branches — a silent regression dropping `spark_conf` would break the KEP contract.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +636,331 @@ def get_session_logs(
...
+            if now - start_time >= timeout:
+                raise TimeoutError(
+                    f"Timeout waiting for SparkApplication to reach "
+                    f"{status}: {self.namespace}/{name} "
+                    f"(timeout: {timeout}s)"
+                )
+
+            time.sleep(polling_interval)
+
+    def get_job_logs(
```
andreyvelich, line 1100: `get_job_logs` is untested. Please add cases for `follow=True` (stream), `follow=False` (split), the no-driver-pod `RuntimeError` (KEP's documented TTL-cleanup case), and the timeout/runtime error wrapping. Note it's a generator, so tests must consume the iterator to execute it.

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,130 @@ def delete_session(self, name: str) -> None:
...
+            Spark job information after reaching one of the target statuses.
+        """
+
+        return self.backend.wait_for_job_status(
+            name=name,
+            status=status,
+            timeout=timeout,
+        )
+
+    def get_job_logs(
```
andreyvelich, line 302: Missing return type — add `-> Iterator[str]` (matches `get_session_logs` and the backend/ABC counterparts). CLAUDE.md requires return types on all functions.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +636,331 @@ def get_session_logs(
             raise RuntimeError(
                 f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}"
             ) from e
+
+    # --------------------------
+    # Batch Job Methods
+    # --------------------------
+
+    def _validate_job(
```
andreyvelich, line 765: Add a `-> None` return type here and on `_validate_file_job` (line 659) — CLAUDE.md requires return types on all functions.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -22,7 +22,32 @@
...
+from kubeflow.spark.types.types import (
+    Driver,
+    Executor,
+    SparkConnectInfo,
+    SparkConnectState,
+    SparkJob,
+    SparkJobStatus,
+)
+
+REMOTE_URI_SCHEMES = ("s3://", "s3a://", "gs://", "gcs://", "http://", "https://", "hdfs://")
```
andreyvelich: `REMOTE_URI_SCHEMES` is defined but never used. Either wire it into `_validate_file_job`/`build_spark_application_cr` for local-vs-remote handling, or remove it (CLAUDE.md: remove unreachable code).

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,130 @@ def delete_session(self, name: str) -> None:
...
+
+            ValueError:
+                If required job fields are missing.
+
+            NotImplementedError:
+                If a Phase 2 feature is used.
+        """
+
+        if isinstance(job, FuncJob):
+            raise NotImplementedError("FuncJob support will be added in Phase 2..")
```
andreyvelich: Typo: double period — `Phase 2..` → `Phase 2.`

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,130 @@ def delete_session(self, name: str) -> None:
...
+            num_executors=num_executors,
+            resources_per_executor=resources_per_executor,
+        ).name
+
+    def get_job(self, name: str) -> SparkJob:
+        """Get a specific Spark job by name."""
+
+        return self.backend.get_job(name)
+
+    def list_jobs(
```
andreyvelich, line 245: Minor: KEP-107 writes `list_jobs(status: Optional[SparkJobStatus])` (single value), while this uses `set[SparkJobStatus]`. The `set` form is more consistent with `wait_for_job_status`, so it's arguably an improvement — worth reconciling the KEP text so the doc and code agree.

Goku2099:
@andreyvelich 
Thanks for the detailed review, I have addressed all the review comments in the latest commit, including the API adjustments, additional unit tests, error handling improvements, public exports, type annotations, and cleanup of unused code. Please take another look when you have time. Thanks

Shekharrajak (commented):
Lets have some e2e 

  - Submit FileJob (remote or built-in sample .py) → wait_for_job_status({COMPLETED}) succeeds
  - get_job returns driver_pod_name + RUNNING→COMPLETED transition
  - list_jobs() includes the submitted job; list_jobs(status={COMPLETED}) filters
  - get_job_logs(name) yields non-empty driver logs
  - Failing job (bad script) → wait_for_job_status({COMPLETED}) surfaces failure fast 
  - delete_job(name) removes the SparkApplication (get_job → not-found)

kubeflow/spark/api/spark_client.py
tariq-hasan: We can remove the KEP reference as it has gone through an update and the line reference is out of date.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +636,345 @@ def get_session_logs(
             raise RuntimeError(
                 f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}"
             ) from e
+
+    # --------------------------
+    # Batch Job Methods
+    # --------------------------
```
tariq-hasan: Since we are using comments as a marker to distinguish batch job methods from interactive connect methods can we have a separate header for spark connect methods in this module?

This is a schematic for what I have been intending:

```
class KubernetesBackend(RuntimeBackend):
    """Kubernetes backend for managing SparkConnect sessions."""

    def __init__(...):
        pass

    # ------------------------------------------------------------------
    # Spark Connect sessions
    # ------------------------------------------------------------------

    def _extract_name_option(...):
        pass

    def _create_session(...):
        pass

    def get_session(...):
        pass

    ....
    ....
    ....

    # ------------------------------------------------------------------
    # Spark batch jobs
    # ------------------------------------------------------------------

    def _validate_job(...):
        pass

    def _validate_file_job(...):
        pass

    def submit_job(...):
        pass

    ....
    ....
    ....

```

We can remove the `Job Lifecycle Methods` marker as they are all technically batch job methods.

kubeflow/spark/api/spark_client.py
tariq-hasan, line 61: Can we have a heading here to demarcate the Spark Connect methods?

```
    # ------------------------------------------------------------------
    # Spark Connect sessions
    # ------------------------------------------------------------------
```

We can have a similar heading for the Spark Application methods below.

```
    # ------------------------------------------------------------------
    # Spark batch jobs
    # ------------------------------------------------------------------
```

This would be in line with how we want to demarcate the two use cases in [`kubeflow/spark/kubernetes/backend.py`](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/backends/kubernetes/backend.py#L640-L642).

tariq-hasan: Do we want to retain the examples here? If we do then we would want to also add examples in the docstring for the `submit_job` method for consistency. Else we can remove these examples here since this would be covered by the docs already.

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,133 @@ def delete_session(self, name: str) -> None:
...
+            raise NotImplementedError("spark_conf support is not yet implemented.")
+
+        if options is not None:
+            raise NotImplementedError("options are not supported in Phase 1.")
+
+        return self.backend.submit_job(
+            job=job,
+            num_executors=num_executors,
+            resources_per_executor=resources_per_executor,
+        ).name
```
tariq-hasan, line 230: I am wondering if we can return the created `SparkJob` instance instead of just its name. This would align the return type for job submission (a dataclass encapsulating info about a job) more closely with what we return for interactive workflows (a dataclass encapsulating info about a session).

More importantly, from a functional perspective, a namespaced name is what makes a K8s resource unique - although it's true that the client is always configured against a single namespace.

I understand that the KEP has been merged but just wanted to point out. @andreyvelich @Shekharrajak

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,133 @@ def delete_session(self, name: str) -> None:
...
+
+    def list_jobs(
+        self,
+        status: set[SparkJobStatus] | None = None,
+    ) -> list[SparkJob]:
+        """List Spark jobs."""
+        return self.backend.list_jobs(status=status)
+
+    def delete_job(self, name: str) -> None:
+        return self.backend.delete_job(name)
```
tariq-hasan: Best to avoid a `return` since we are returning `None` and for consistency with how [`delete_session`](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/api/spark_client.py#L146-L148) implements it.

```suggestion
    def delete_job(self, name: str) -> None:
        self.backend.delete_job(name)
```

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +150,133 @@ def delete_session(self, name: str) -> None:
...
+
+    def wait_for_job_status(
+        self,
+        name: str,
+        status: set[SparkJobStatus] | None = None,
+        timeout: int = 600,
+        polling_interval: int = 2,
+    ) -> SparkJob:
```
tariq-hasan, line 275: Can we confirm if we want to expose `polling_interval` in the public `wait_for_job_status` API?

The KEP does not expose this parameter in its [specification](https://github.com/kubeflow/sdk/blob/main/proposals/107-spark-client/README.md?plain=1#L613-L619).

kubeflow/spark/types/types.py
```diff
@@ -113,3 +115,55 @@ class Executor:
...
+    CREATED = "Created"
+    RUNNING = "Running"
+    COMPLETED = "Completed"
+    FAILED = "Failed"
+
+
+@dataclass
+class SparkJob:
+    """Information about a Spark batch job."""
+
```
tariq-hasan, line 198: For consistency with [SparkConnectInfo](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/types/types.py#L34-L55) can we enhance the docstring here to describe what each arg means?

kubeflow/spark/types/types.py
tariq-hasan: Not directly related to the PR but can we remove the statements about GPUs? I believe the intended meaning here is  that the resources datatype is generic enough to support any accelerator under the extended resource model. That said it is easy to confuse this with thinking Spark Connect CRD supports GPUs. As it stands Spark Connect CRD does not support GPUs at this time. We have a pending [issue](https://github.com/kubeflow/spark-operator/issues/2958) in the spark operator backlog for GPU support.

Same for the executor dataclass below.

tariq-hasan: The KEP reference needs to be removed as the line number is out of date.

Same for the executor dataclass below.

kubeflow/spark/backends/kubernetes/constants.py
tariq-hasan: Can we move all the default configs to the bottom of the module - given these are shared between both Spark Connect and Spark Application?

That way this module would be split as Spark Connect first, Spark Application second and shared constants third. This will make it easy for anyone reading the code to easily find relevant constants.

kubeflow/spark/__init__.py
tariq-hasan, line 36:
```suggestion
from kubeflow.spark.types.types import (
    Driver,
    Executor,
    FileJob,
    FuncJob,
    SparkConnectInfo,
    SparkConnectState,
    SparkJob,
    SparkJobStatus,
)
```

kubeflow/spark/__init__.py
```diff
@@ -37,6 +37,10 @@
     # Types
     "Driver",
     "Executor",
+    "FileJob",
+    "FuncJob",
+    "SparkJob",
+    "SparkJobStatus",
     "SparkConnectInfo",
     "SparkConnectState",
```
tariq-hasan, line 47:
```suggestion
    # Types
    "Driver",
    "Executor",
    "FileJob",
    "FuncJob",
    "SparkConnectInfo",
    "SparkConnectState",
    "SparkJob",
    "SparkJobStatus",
```

kubeflow/spark/backends/base.py
```diff
@@ -17,7 +17,14 @@
...
 from collections.abc import Iterator
 
-from kubeflow.spark.types.types import Driver, Executor, SparkConnectInfo
+from kubeflow.spark.types.types import (
+    Driver,
+    Executor,
+    FileJob,
+    SparkConnectInfo,
+    SparkJob,
+    SparkJobStatus,
```
tariq-hasan, line 29:
```suggestion
    Driver,
    Executor,
    FileJob,
    FuncJob,
    SparkConnectInfo,
    SparkJob,
    SparkJobStatus,
```

kubeflow/spark/backends/base.py
tariq-hasan, line 37:
```suggestion
    """Abstract base class for Spark backends.

    All Spark backends must implement these methods to manage Spark Connect sessions and Spark batch jobs.
    """
```

tariq-hasan, line 38: We can group the methods here with

```
    # ------------------------------------------------------------------
    # Spark Connect sessions
    # ------------------------------------------------------------------
```

and

```
    # ------------------------------------------------------------------
    # Spark batch jobs
    # ------------------------------------------------------------------
```

as well for consistency with the other modules.

tariq-hasan: The spark connect controller at this time only ever oscillates between `Ready` and `NotReady`.

The `Provisioning` state is not leveraged in the operator.

```suggestion
       Returns:
           SparkConnectInfo with session details.
```

kubeflow/spark/backends/base.py
```diff
@@ -148,3 +155,108 @@ def get_session_logs(
...
         """
         raise NotImplementedError()
+
+    @abc.abstractmethod
+    def submit_job(
+        self,
+        job: FileJob,
+        executor: Executor | None = None,
+        num_executors: int | None = None,
+        resources_per_executor: dict[str, str] | None = None,
```
tariq-hasan, line 151:
```suggestion
        self,
        job: FileJob | FuncJob,
        executor: Executor | None = None,
        num_executors: int | None = None,
        resources_per_executor: dict[str, str] | None = None,
```

kubeflow/spark/backends/base.py
```diff
@@ -148,3 +155,108 @@ def get_session_logs(
...
+    ) -> SparkJob:
+        """Submit a Spark batch job.
+
+        Args:
+            job: File-based Spark application to execute.
```
tariq-hasan:
```suggestion
        Args:
            job: Spark application to execute.
```

kubeflow/spark/backends/base.py
```diff
@@ -148,3 +155,108 @@ def get_session_logs(
...
+            executor: Executor configuration.
+            num_executors: Number of executor instances.
+            resources_per_executor: Resource requirements for each executor.
+
+        Returns:
+            Submitted Spark job information.
+        """
```
tariq-hasan:
```suggestion
        Returns:
            Submitted Spark job information.

        Raises:
            TypeError: If ``job`` is not a supported job type.
            ValueError: If a required job field is missing.
            TimeoutError: If the submission request times out.
            RuntimeError: If submission fails.
        """
```

kubeflow/spark/backends/base.py
tariq-hasan, line 45: The signature in the kubernetes backend matches for the [`create_and_connect`](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/backends/kubernetes/backend.py#L539-L549) function. It does not match for the [`connect`](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/backends/kubernetes/backend.py#L415-L420) function.

```suggestion
    @abc.abstractmethod
    def create_and_connect(
        self,
```

tariq-hasan: There's also a mismatch with the arguments and return type in the kubernetes backend [implementation](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/backends/kubernetes/backend.py#L539-L549).

tariq-hasan: I'm thinking if we want to mark this internal with a leading underscore and place it on the abstract class given that this is a public interface that all backends are meant to implement. Perhaps we either remove the method or remove the leading underscore.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -22,7 +23,32 @@
...
+    "SUSPENDED": SparkJobStatus.RUNNING,
+    "RESUMING": SparkJobStatus.RUNNING,
+    "COMPLETED": SparkJobStatus.COMPLETED,
+    "FAILED": SparkJobStatus.FAILED,
+    "SUBMISSION_FAILED": SparkJobStatus.FAILED,
+    "FAILING": SparkJobStatus.FAILED,
+    "PENDING_RERUN": SparkJobStatus.FAILED,
+    "INVALIDATING": SparkJobStatus.FAILED,
+    "UNKNOWN": SparkJobStatus.FAILED,
+}
```
tariq-hasan: Can we make the mapping more idiomatic (in a Python context) as follows and move it to `types.py`?

```python
_SDK_STATE_BY_OPERATOR_STATE: dict[str, SparkJobStatus] = {
    state: sdk_status
    for sdk_status, operator_states in {
        SparkJobStatus.CREATED: ("SUBMITTED"),
        SparkJobStatus.RUNNING: ("RUNNING", "SUCCEEDING", "SUSPENDING", "SUSPENDED", "RESUMING"),
        SparkJobStatus.COMPLETED: ("COMPLETED",),
        SparkJobStatus.FAILED: (
            "FAILED", "SUBMISSION_FAILED", "FAILING",
            "PENDING_RERUN", "INVALIDATING", "UNKNOWN",
        ),
    }.items()
    for state in operator_states
}

class SparkJobStatus(str, Enum):
    CREATED = "Created"
    RUNNING = "Running"
    COMPLETED = "Completed"
    FAILED = "Failed"

    @classmethod
    def from_operator_state(cls, raw_state: str | None) -> "SparkJobStatus":
        """Map a SparkApplication ``applicationState.state`` to an SDK status.

        Unknown states (including future operator states) map to CREATED rather
        than FAILED, since the operator's status value set is open and may evolve.
        """
        return _SDK_STATE_BY_OPERATOR_STATE.get((raw_state or "").upper(), cls.CREATED)
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
...
+
+
+def generate_job_name() -> str:
+    """Generate a unique batch job name.
+
+    Returns:
+        Job name in format: spark-job-{uuid}.
+    """
+    short_uuid = str(uuid.uuid4())[:8]
+    return f"spark-job-{short_uuid}"
```
tariq-hasan: It may be better to align with [Spark Connect](https://github.com/Goku2099/sdk/blob/b6ef980a061f6f162c35f31616cb3a046585aebf/kubeflow/spark/backends/kubernetes/utils.py#L60-L61) and create a dedicated constant.

```suggestion
    short_uuid = str(uuid.uuid4())[:8]
    return f"{constants.JOB_NAME_PREFIX}-{short_uuid}"
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
...
+        volume_mounts: Optional volume mounts.
+
+    Returns:
+        SparkApplication DriverSpec model.
+    """
+    return models.SparkV1beta2DriverSpec(
+        cores=constants.DEFAULT_DRIVER_CPU,
+        memory=_memory_kubernetes_to_spark(constants.DEFAULT_DRIVER_MEMORY),
+        volume_mounts=volume_mounts,
+    )
```
tariq-hasan: Driver resources should be configurable through this function. The goal would be for `get_driver_spec` and `get_server_spec_from_driver` to perform the same logic but only differ in the return type. Can we move the common logic to a shared utility function and retain dedicated methods only for the return type?

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
...
+            cores = _validate_cpu_value(resource_dict["cpu"])
+
+        if "memory" in resource_dict:
+            memory = _memory_kubernetes_to_spark(resource_dict["memory"])
+
+    return models.SparkV1beta2ExecutorSpec(
+        cores=cores,
+        memory=memory,
+        instances=instances,
+    )
```
tariq-hasan: Same for executor spec.

`get_executor_spec` and `get_executor_spec_from_executor` are meant to perform the same logic but only differ in the return type. We should move the common logic to a shared utility function and retain dedicated methods only for the return type.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
         service_name=server_status.service_name if server_status else None,
         creation_timestamp=spark_connect_cr.metadata.creation_timestamp,
     )
+
+
+# --------------------------
+# Batch Job Utility Functions
+# --------------------------
```
tariq-hasan: It's ideal to group all the functions in the module into three distinct sections.

```
    # ------------------------------------------------------------------
    # Spark Connect session utility functions
    # ------------------------------------------------------------------
```

```
    # ------------------------------------------------------------------
    # Spark batch job utility functions
    # ------------------------------------------------------------------
```

```
    # ------------------------------------------------------------------
    # Shared utility functions
    # ------------------------------------------------------------------
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
...
+            arguments=arguments or None,
+            driver=get_driver_spec(),
+            executor=get_executor_spec(
+                executor=executor,
+                num_executors=num_executors,
+                resources_per_executor=resources_per_executor,
+            ),
+        ),
+    )
+
```
tariq-hasan, line 663: Is the expectation that we will only support Python jobs?

File type can be Java/Scala and we can pass in a jar file with  a class name to support Java/Scala jobs as well.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +329,222 @@ def get_spark_connect_info_from_cr(
...
+        raise ValueError(f"Invalid CPU type '{type(cpu)}'. Expected str or int.")
+
+    if cores <= 0:
+        raise ValueError("CPU cores must be greater than 0")
+
+    # safety upper bound
+    if cores > 1024:
+        raise ValueError("CPU cores value is unrealistically large")
+
+    return cores
```
tariq-hasan: CPU values should be configurable for milli CPUs and fractional values. Can we follow the review comments in https://github.com/kubeflow/sdk/pull/465 to make the changes and enable both Spark Connect and batch jobs to benefit from this feature?

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: Can we also add `"": SparkJobStatus.CREATED`? This is the [new](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L351) state for SparkApplication CRD.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +636,345 @@ def get_session_logs(
...
+        name: str,
+        status: set[SparkJobStatus] | None = None,
+        timeout: int = 600,
+        polling_interval: int = 2,
+    ) -> SparkJob:
+        """Wait for a Spark job to reach one of the target states."""
+
+        if status is None:
+            status = {SparkJobStatus.COMPLETED}
+
```
tariq-hasan, line 1052: Because of this defaulting the polling logic only checks for completed state. A failed state would therefore show as timeout error after the complete timeout. Can we have a fail-fast mechanism in place when we detect a failed state and return early?

kubeflow/spark/types/types.py
tariq-hasan, line 34: There is no `RUNNING` state in the operator. Do we still want to keep `RUNNING` here? Currently the semantics of `READY` in the operator match the semantics of both `READY` and `RUNNING` in the SDK so `RUNNING` here is redundant.

Add to that the operator makes no use of its `PROVISIONING` state at this time in its SparkConnect controller.

There is also a New state in the CRD but no support or mapping for it in the SDK.

Ref: https://github.com/kubeflow/spark-operator/blob/master/api/v1alpha1/sparkconnect_types.go#L170-L180

@andreyvelich @Shekharrajak

tariq-hasan (commented):
Hi @Goku2099! I have added comments on the PR. They are a mix of implementation aspects, design details, and style/consistency nits.

cc @RobuRishabh @nabuskey @vara-bonthu

kubeflow/spark/api/spark_client.py
Goku2099: cc @andreyvelich @Shekharrajak

kubeflow/spark/api/spark_client.py
Goku2099: https://github.com/kubeflow/sdk/pull/521#discussion_r3515590069

kubeflow/spark/types/types.py
Goku2099: cc @Shekharrajak @andreyvelich WYT

kubeflow/spark/backends/base.py
Goku2099: Good point. Since this is only an implementation detail used by create_and_connect() and isn't called by the client, I will remove it from the abstract interface.

kubeflow/spark/types/types.py
andreyvelich: Sure, let's remove it for now.

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: Good suggestion. I will refactor the shared resource resolution into common utility helpers and keep the DriverSpec/ExecutorSpec builders responsible only for constructing the respective return types.

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: I will refactor the shared resource resolution into common helper functions so both the Spark Connect and SparkApplication builders share the same logic and differ only in the returned API model.

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: Also probably better to refactor the function names - something like `get_spark_connect_driver_spec` and `get_spark_job_driver_spec` or similar.

Same for executors below.

Goku2099: Okay, It makes sense I will refactor the function names

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: The KEP exposes FileJob.main_class but doesn't specify how to determine the SparkApplication type. Python and R can be inferred from the file extension, but Java and Scala both use JARs. Should we default JARs to Java, or would you prefer adding an explicit application type to FileJob?

WYT @andreyvelich @tariq-hasan @Shekharrajak

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: I will update wait_for_job_status to fail fast when the job reaches FAILED (unless FAILED is one of the requested target states) and add unit tests for this behavior.

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: I would think use of the extension would be more of a fallback mechanism and a user would need to explicitly specify what language they are using.

I suppose most use cases would be for Python but a data scientist might also just want to run some third-party pre-compiled jar file using the Python SDK - although that may be more of a less common case.

Maybe we could expose the type directly on the FileJob dataclass as that is [supported](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L41-L43) on the CRD but not sure what others think.

Goku2099: Sounds good. Let's wait for feedback from @andreyvelich and @Shekharrajak before deciding on the API.

kubeflow/spark/api/spark_client.py
andreyvelich: Maybe we can discuss this as a followup?

In the `train()` API we also return [only TrainJob name](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L116), since we need to make additional `get_job()` call to get actual info about the TrainJob after creation.
IMO, the UX is slightly different for `connect()`, since users leverage Session object itself to interact with Spark session. Thus, we return the object after `connect()` call.

kubeflow/spark/api/spark_client.py
andreyvelich: @Goku2099 Let's remove the examples for now. We can directly add more examples here: https://sdk.kubeflow.org/en/latest/examples.html#spark-examples

kubeflow/spark/types/types.py
andreyvelich: We should not have `READY` condition. @Goku2099 As we discussed to keep states consistent w/ TrainJob, we only should have these states for the phase 1: https://github.com/kubeflow/sdk/tree/main/proposals/107-spark-client#application-states

SDK Status | SparkApplication States
-- | --
CREATED | SUBMITTED
RUNNING | RUNNING, SUCCEEDING, SUSPENDING, SUSPENDED, RESUMING
COMPLETED | COMPLETED
FAILED | FAILED, SUBMISSION_FAILED, FAILING, PENDING_RERUN, INVALIDATING, UNKNOWN

andreyvelich: Oh, sorry, @tariq-hasan you mentioned about SparkConnect states, not SparkJob.
Can we create tracking issue to resolve this in the followup PR? 

We can simplify those statuses.
cc @Shekharrajak

kubeflow/spark/backends/kubernetes/utils.py
andreyvelich: @tariq-hasan Let's discuss this as a followup? I believe, we should support Python initially, to make `FileJob` and `FuncJob` compatible with each other.
In the future, we can allow users to submit `FileJob` using Scala files with ScalaSpark and different image.

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan:
```suggestion
class KubernetesBackend(RuntimeBackend):
        """Kubernetes backend for managing SparkConnect sessions and Spark batch jobs."""
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+                f"{constants.SPARK_APPLICATION_KIND}: "
+                f"{self.namespace}/{name}"
+            ) from e
+
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to get logs for "
+                f"{constants.SPARK_APPLICATION_KIND}: "
+                f"{self.namespace}/{name}"
+            ) from e
```
tariq-hasan: `get_job_logs` and `get_session_logs` share the same logic. We can reduce code duplication by moving the shared logic to a common helper such as `_get_pod_logs` under a `Shared helpers` banner like we do in the case of [`utils.py`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/backends/kubernetes/utils.py#L38-L40). Or, better yet, we can move the shared helper to `utils.py`.

kubeflow/spark/backends/base.py
```diff
@@ -101,50 +119,131 @@ def delete_session(self, name: str) -> None:
...
-        Polls the session status until it reaches READY state or times out.
+    # ------------------------------------------------------------------
+    # Spark batch jobs
+    # ------------------------------------------------------------------
+
+    @abc.abstractmethod
+    def submit_job(
+        self,
+        job: FileJob | FuncJob,
+        executor: Executor | None = None,
```
tariq-hasan: As per [this conversation](https://github.com/kubeflow/sdk/pull/524#pullrequestreview-4532714703) the agreement was to not have `Driver` or `Executor` on the `submit_job` API.

The [KEP](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/proposals/107-spark-client/README.md?plain=1#L540-L547) and the [public submit_job API](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/api/spark_client.py#L173-L180) remain faithful to this alignment.

`Executor` (but not `Driver`) does appear on both the base backend and K8s backend [signature](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/backends/kubernetes/backend.py#L698-L704).

I am wondering what the expectation is regarding the scope of `Driver` and `Executor` in the spark job context. Are they completely out of scope or do they have a place in the implementation?

kubeflow/spark/types/types.py
tariq-hasan: We remove reference to GPU support.

Same for executor.

```suggestion
   Args:
       image: Custom container image for the driver.
       resources: Resource requirements as dict (e.g., {"cpu": "2", "memory": "4Gi"}).
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+        spark_application_list = models.SparkV1beta2SparkApplicationList.from_dict(
+            response,
+        )
+
+        jobs = [get_spark_application_info_from_cr(app) for app in spark_application_list.items]
+
+        if status:
+            jobs = [job for job in jobs if job.status in status]
+
+        return jobs
```
tariq-hasan, line 977: Do we want `list_jobs` to return all jobs in the namespace - regardless of whether they were created by the client or not? Or do we want to return only those that were created by the client or SDK?

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan, line 435: We probably just want to reference the constant so we do not have to keep updating the version here.

```suggestion
   Args:
       name: Session name.
       namespace: Kubernetes namespace.
       spark_version: Spark version (default: `constants.DEFAULT_SPARK_VERSION`).
       num_executors: Number of executor instances (simple mode).
```

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +165,137 @@ def delete_session(self, name: str) -> None:
...
+        resources_per_executor: dict[str, str] | None = None,
+        spark_conf: dict[str, str] | None = None,
+        options: list[Any] | None = None,
+    ) -> str:
+        """Submit a batch Spark job.
+
+        This method supports two job types:
+
+        - **FileJob**: Submit an existing Spark application referenced by a local
+        or remote file source.
```
tariq-hasan:
```suggestion
        - **FileJob**: Submit a Spark application referenced by a local
        or remote file source.
```

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +165,137 @@ def delete_session(self, name: str) -> None:
...
+        """
+
+        if isinstance(job, FuncJob):
+            raise NotImplementedError("FuncJob support will be added in Phase 2.")
+
+        if not isinstance(job, FileJob):
+            raise TypeError("job must be an instance of FileJob or FuncJob.")
+
+        if not job.file_source:
+            raise ValueError("FileJob.file_source is required.")
```
tariq-hasan: We have validation here as well as in [`backend.py`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/backends/kubernetes/backend.py#L648-L696). If we do validate then probably best to do it all in one place i.e., in the K8s backend. That way, the spark client module stays lightweight similar to what we have for [trainer client](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/trainer/api/trainer_client.py).

kubeflow/spark/types/types.py
```diff
@@ -106,10 +111,127 @@ class Executor:
...
+            remote URIs such as s3a://, gs://, hdfs:// and https://.
+        args: Optional command-line arguments passed to the application.
+        main_class: Optional main class for JVM applications.
+            Ignored for Python applications.
+    """
+
+    file_source: str
+    args: list[str] | None = None
+    main_class: str | None = None
+
```
tariq-hasan, line 220: Should we update the [KEP](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/proposals/107-spark-client/README.md?plain=1#L152) to reflect that FileJob is imported from `kubeflow.spark.types.types`?

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +447,166 @@ def get_spark_connect_info_from_cr(
...
+        ValueError:
+            If the default driver resource configuration is invalid.
+    """
+    cores, memory = _resolve_driver_resources()
+
+    return models.SparkV1beta2DriverSpec(
+        cores=cores,
+        memory=memory,
+        volume_mounts=volume_mounts,
+    )
```
tariq-hasan, line 585: `get_spark_job_driver_spec` is an internal helper and the only place it is called is in `build_spark_application_cr` where we do not pass in any volume mount. Do we need volume mount for any use case?

This also ties in with the question of how to support configurable tuning of drivers and executors other than using just the default [constants](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/backends/kubernetes/constants.py#L41-L51) - given that `Driver` and `Executor` do not appear on the public [submit_job](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/api/spark_client.py#L173-L180) API.

As an example, how should we pass in `ServiceAccount` that a driver needs to create executor pods? The SparkConnect implementation relies on the `Driver` dataclass to configure this `ServiceAccount`.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+
+        except multiprocessing.TimeoutError as e:
+            raise TimeoutError(
+                f"Timeout creating SparkApplication: {self.namespace}/{job_name}"
+            ) from e
+
+        except Exception as e:
+            raise RuntimeError(
+                f"Failed to create SparkApplication: {self.namespace}/{job_name}: {e}"
+            ) from e
```
tariq-hasan, line 878: Do we want to expose CRD semantics on user-facing messages i.e., should `SparkApplication` be renamed to Spark job?

kubeflow/spark/types/types.py
```diff
@@ -106,10 +111,127 @@ class Executor:
...
+        ),
+        SparkJobStatus.COMPLETED: ("COMPLETED",),
+        SparkJobStatus.FAILED: (
+            "FAILED",
+            "SUBMISSION_FAILED",
+            "FAILING",
+            "PENDING_RERUN",
+            "INVALIDATING",
+            "UNKNOWN",
+        ),
```
tariq-hasan, line 178: I'm wondering if it would be feasible to have typed operator states here (instead of bare strings) by pulling in enums directly from the generated Python models in the spark-operator repo.

@andreyvelich @Shekharrajak

kubeflow/spark/api/spark_client.py
tariq-hasan, line 132: I'm noticing that [`list_jobs`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/api/spark_client.py#L249-L254) supports a `status` argument but `list_sessions` does not. Just wondering if there's any reason for the difference.

If this turns out to be a gap on `list_sessions` perhaps we can make the correction as a separate issue but just wanted to point out the asymmetry.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+            raise ValueError("`job.file_source` must be a non-empty string.")
+
+        if job.args is not None:
+            if not isinstance(job.args, list):
+                raise ValueError("`job.args` must be a list of strings.")
+
+            if not all(isinstance(arg, str) for arg in job.args):
+                raise ValueError("All `job.args` must be strings.")
+        if job.main_class is not None and not isinstance(job.main_class, str):
+            raise ValueError("`job.main_class` must be a string.")
```
tariq-hasan:
```suggestion
        if job.args is not None:
            if not isinstance(job.args, list):
                raise ValueError("`job.args` must be a list of strings.")

            if not all(isinstance(arg, str) for arg in job.args):
                raise ValueError("All `job.args` must be strings.")

        if job.main_class is not None and not isinstance(job.main_class, str):
            raise ValueError("`job.main_class` must be a string.")
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +447,166 @@ def get_spark_connect_info_from_cr(
...
+            driver=get_spark_job_driver_spec(),
+            executor=get_spark_job_executor_spec(
+                executor=executor,
+                num_executors=num_executors,
+                resources_per_executor=resources_per_executor,
+            ),
+        ),
+    )
+
+
```
tariq-hasan, line 664: Just a short note for reference here that the `submit_job` API currently accepts [`FileJob`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/api/spark_client.py#L175) with the provision to input a [`main_class`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/types/types.py#L209-L224) parameter. This would break the code as it exists if a user inputs a jar file with a main class that would then be typed as a Python application by `build_spark_application_cr`.

This is an implementation detail and can be resolved once we have finalized the discussion.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+            RuntimeError: If the SparkApplication reaches the FAILED state before reaching
+                one of the target statuses.
+            TimeoutError: If the target status is not reached within the timeout.
+        """
+
+        if status is None:
+            status = {SparkJobStatus.COMPLETED}
+
+        if polling_interval < 0:
+            raise ValueError("polling_interval must be non-negative.")
```
tariq-hasan: We do not want to have `polling_interval` equal 0.

```suggestion
        if polling_interval <= 0:
            raise ValueError("polling_interval must be positive.")
```

We should have a similar validation for `timeout` as well.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -624,3 +640,424 @@ def get_session_logs(
...
+
+        start_time = time.time()
```
tariq-hasan: There's a mix of `time.time()` and `time.monotonic()` throughout the backend module without a clear consistency argument. Should we standardize on `time.monotonic()` and apply this consistently in all places?

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -71,6 +144,92 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
...
+    if not math.isfinite(cores) or cores <= 0:
+        raise ValueError(f"Invalid CPU value: {cpu!r}")
+
+    cores = math.ceil(cores)
+
+    if cores > 1024:
+        raise ValueError("CPU cores value is unrealistically large")
+
+    return cores
+
```
tariq-hasan, line 291: The logic in `_validate_cpu_value` assumes that the CRDs use the number of cores as CPU. This is indeed the case for [`SparkConnect`](https://github.com/kubeflow/spark-operator/blob/master/api/v1alpha1/sparkconnect_types.go#L108-L111) but not for [`SparkApplication`](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L544-L547).

There's an [issue](https://github.com/kubeflow/spark-operator/issues/2957) in the spark-operator repo to address this gap. Can we create an issue in the sdk repo as well to close this gap once the necessary support is added on the `SparkConnect` CRD?

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -303,3 +447,166 @@ def get_spark_connect_info_from_cr(
...
+        if cr.status.application_state:
+            status = SparkJobStatus.from_operator_state(
+                cr.status.application_state.state,
+            )
+
+        if cr.status.driver_info:
+            driver_pod_name = cr.status.driver_info.pod_name
+
+    return SparkJob(
+        name=cr.metadata.name,
```
tariq-hasan, line 697: `cr.metadata.name` will fail if it falsy. Better to directly replicate the logic in [`get_spark_connect_info_from_cr`](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/backends/kubernetes/utils.py#L411-L449) for both implementation accuracy and consistency (with tweaks wherever necessary to adapt it for the `SparkApplication` case).

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -123,6 +130,141 @@ def get_spark_connect(
...
+                memory="1g",
+            ),
+            executor=models.SparkV1beta2ExecutorSpec(
+                cores=1,
+                memory="1g",
+                instances=1,
+            ),
+        ),
+    )
+
```
tariq-hasan, line 173: Can we use a shared helper to avoid code duplication among `get_spark_application` and the two related functions?

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -138,11 +280,18 @@ def mock_get_response(name: str) -> dict:
         return get_spark_connect(name=name, state="Provisioning").to_dict()
     elif name == SPARK_CONNECT_FAILED:
         return get_spark_connect(name=name, state="Failed").to_dict()
+    elif name.startswith("spark-job-"):
+        return get_spark_application(name=name).to_dict()
```
tariq-hasan, line 210: Let's define what we need in the shared test utilities [module](https://github.com/Goku2099/sdk/blob/0c2ce987e81338ad49b720d303bfe58a7d60c766/kubeflow/spark/test/common.py) to ensure consistency with the SparkConnect implementation.

tariq-hasan (commented):
Hi @Goku2099! I have made additional comments. Thanks again for all the work on the PR and addressing prior comments.

kubeflow/spark/backends/base.py
Goku2099: I will remove the stale executor reference from the base backend and keep the public API limited to job, num_executors, and resources_per_executor for this phase. If we later expose richer driver/executor configuration, we can extend the API in a follow-up KEP.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: returning all SparkApplications in the namespace provides a simpler and more predictable user experience. Restricting it to SDK-created jobs would introduce extra tracking/filtering logic and may hide jobs that users still expect to manage through the SDK.

kubeflow/spark/api/spark_client.py
Goku2099: Agreed. I will remove the duplicate validation from spark_client.py and keep the Phase 1 feature gating there.

kubeflow/spark/types/types.py
Goku2099: sure

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: I think it makes sense to keep the implementation aligned with the current API and remove unused internal parameters like volume_mounts for now. We can add them back when configurable driver/executor tuning is introduced in a later phase.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Renaming makes sense, I will change the user- facing messages.

kubeflow/spark/api/spark_client.py
Goku2099: `list_jobs(status=...)` is primarily intended for lifecycle management (e.g. monitoring running, completed, or failed batch jobs), whereas `list_sessions()` was designed as a discovery API to enumerate available Spark Connect sessions before connecting. That said, I agree the asymmetry is worth discussing, and if status-based filtering is useful for sessions as well, I did be happy to make the APIs consistent.

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: Agreed. This depends on how we finalize application type support for FileJob. Once that design is settled, I will update build_spark_application_cr() accordingly.

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: Sounds good. I will create a follow-up SDK issue to track this once the SparkConnect CRD supports the required CPU representation.

kubeflow/spark/backends/kubernetes/backend_test.py
Goku2099: I will Merge get_spark_application, get_spark_application_with_state, and get_spark_application_without_status into a single shared get_spark_application helper to remove the duplication.

kubeflow/spark/backends/kubernetes/backend_test.py
Goku2099: Sounds good. I will move the common SparkApplication test helpers into the shared test utilities module to keep them consistent with the SparkConnect tests.

kubeflow/spark/backends/base.py
Goku2099: cc @andreyvelich @Shekharrajak

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: cc @andreyvelich @Shekharrajak

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: cc @andreyvelich @Shekharrajak

kubeflow/spark/api/spark_client.py
Goku2099: cc @andreyvelich @Shekharrajak

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: Issue has been created: https://github.com/kubeflow/sdk/issues/578.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -596,31 +613,419 @@ def get_session_logs(
...
+        if job.main_class is not None and not isinstance(job.main_class, str):
+            raise ValueError("`job.main_class` must be a string.")
+
+    def submit_job(
+        self,
+        job: FileJob | FuncJob,
+        executor: Executor | None = None,
+        num_executors: int | None = None,
+        resources_per_executor: dict[str, str] | None = None,
+    ) -> SparkJob:
```
tariq-hasan, line 817: The executor here is a dead field because it is not called by the public `submit_job` API.

```suggestion
    def submit_job(
        self,
        job: FileJob | FuncJob,
        num_executors: int | None = None,
        resources_per_executor: dict[str, str] | None = None,
    ) -> SparkJob:
```

In addition to this change we need to remove the references to the `executor` field in the docstring and the call to `build_spark_application_cr` below.

We also need to remove all references to the `executor` field in the function signatures, docstrings and function body for `build_spark_application_cr` and `get_spark_job_executor_spec`.

`_resolve_executor_spec` will continue to need the `executor` field so long as the `connect` API supports the `executor` field since this is a shared helper.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -596,31 +613,419 @@ def get_session_logs(
...
+                    self.namespace,
+                    name,
+                    job.status,
+                    time.monotonic() - start_time,
+                )
+                return job
+            if job.status == SparkJobStatus.FAILED and SparkJobStatus.FAILED not in status:
+                raise RuntimeError(
+                    f"Spark job reached failed state: {self.namespace}/{name}(status={job.status})"
+                )
```
tariq-hasan, line 1077: Since we [subsume](https://github.com/Goku2099/sdk/blob/c52a67003b470add84c2da18046bf6abb52504bb/kubeflow/spark/types/types.py#L171-L178) the operator restart semantics under the `FAILED` state the fail-fast mechanism will fail a `SparkJob` where the corresponding `SparkApplication` is in the process of being restarted.

I believe this may not be important since the [default](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/defaults.go#L52-L54) restart policy in the operator is set to `Never` and SDK users would not configure restart policies but just wanted to point out either way.

cc @andreyvelich @Shekharrajak

examples/spark/batch_job_lifecycle.py
```diff
@@ -0,0 +1,253 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import os
+
+from kubeflow.common.types import KubernetesBackendConfig
+from kubeflow.spark import FileJob, SparkClient, SparkJobStatus
+
+REMOTE_PI = "https://raw.githubusercontent.com/apache/spark/master/examples/src/main/python/pi.py"
```
tariq-hasan, line 22: As per the offline discussion, it is better to have an example module sitting in the Kubeflow SDK project and use that here instead of a remote URL that might change - breaking the E2Es.

Also, can we have additional examples for filesystem URIs as well - `s3a://`, `gs://`, `hdfs://`, etc.?

kubeflow/spark/backends/kubernetes/constants.py
```diff
@@ -20,6 +20,24 @@
...
+
+# SparkApplication CRD
+SPARK_APPLICATION_GROUP = "sparkoperator.k8s.io"
+SPARK_APPLICATION_VERSION = "v1beta2"
+SPARK_APPLICATION_PLURAL = "sparkapplications"
+SPARK_APPLICATION_KIND = "SparkApplication"
+
 # Default values; keep major.minor aligned with pyspark-connect in pyproject.toml
 DEFAULT_SPARK_VERSION = "4.0.1"
 DEFAULT_SPARK_IMAGE = "apache/spark:4.0.1"
```
tariq-hasan, line 43: We may need to add connector packages here e.g., `hadoop-aws`, `gcs-connector`, etc. here and in [`pyproject.toml`](https://github.com/Goku2099/sdk/blob/c52a67003b470add84c2da18046bf6abb52504bb/pyproject.toml) - for filesystem URI support.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -596,31 +613,419 @@ def get_session_logs(
...
+                f"{constants.SPARK_APPLICATION_KIND}: "
+                f"{self.namespace}/{name}"
+            ) from e
+
+        except RuntimeError as e:
+            raise RuntimeError(
+                f"Failed to get logs for "
+                f"{constants.SPARK_APPLICATION_KIND}: "
+                f"{self.namespace}/{name}"
+            ) from e
```
tariq-hasan: This would ensure that any runtime error (from `job.driver_pod_name` being falsy) gets raised as soon as `get_job_logs` is called - since `get_job_logs` is no longer the generator function.

```suggestion
        def _stream() -> Iterator[str]:
            try:
                yield from read_pod_logs(
                    core_api=self.core_api,
                    namespace=self.namespace,
                    pod_name=job.driver_pod_name,
                    follow=follow,
                )

            except TimeoutError as e:
                raise TimeoutError(
                    f"Timeout to get logs for "
                    f"{constants.SPARK_APPLICATION_KIND}: "
                    f"{self.namespace}/{name}"
                ) from e

            except RuntimeError as e:
                raise RuntimeError(
                    f"Failed to get logs for "
                    f"{constants.SPARK_APPLICATION_KIND}: "
                    f"{self.namespace}/{name}"
                ) from e

        return _stream()
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -596,31 +613,419 @@ def get_session_logs(
...
             raise TimeoutError(
                 f"Timeout to get logs for {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}"
             ) from e
-        except Exception as e:
+
+        except RuntimeError as e:
             raise RuntimeError(
                 f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}"
             ) from e
+
```
tariq-hasan, line 760: Same pattern as for `get_job_logs`.

```suggestion
        def _stream() -> Iterator[str]:
            try:
                yield from read_pod_logs(
                    core_api=self.core_api,
                    namespace=self.namespace,
                    pod_name=info.driver_pod_name,
                    follow=follow,
                )

            except TimeoutError as e:
                raise TimeoutError(
                    f"Timeout to get logs for {constants.SPARK_CONNECT_KIND}: "
                    f"{self.namespace}/{name}"
                ) from e

            except RuntimeError as e:
                raise RuntimeError(
                    f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: "
                    f"{self.namespace}/{name}"
                ) from e

        return _stream()
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -71,6 +202,92 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
     return num + spark_suffix
```
tariq-hasan: Technically the memory here gets passed to `spark.{driver,executor}.memory` and Spark natively converts it to JVM `-Xmx` values which cannot be fractional. In other words we cannot support fractional memory and we need to run the conversion logic in this function to convert fractional memory to an absolute number.

```suggestion
    if not memory or not memory[-1].isalpha():
        return memory

    m = re.match(r"^(\d+(?:\.\d+)?)\s*([KMGTPE]i?|[kmgtp]b?)?$", memory, re.IGNORECASE)
    if not m:
        return memory

    coefficient, suffix = m.group(1), (m.group(2) or "").lower()

    exponent_by_suffix = {
        "ki": 10, "k": 10, "kb": 10,
        "mi": 20, "m": 20, "mb": 20,
        "gi": 30, "g": 30, "gb": 30,
        "ti": 40, "t": 40, "tb": 40,
        "pi": 50, "p": 50, "pb": 50,
        "ei": 60,
    }
    if suffix not in exponent_by_suffix:
        return memory

    exponent = exponent_by_suffix[suffix]

    spark_suffix = {10: "k", 20: "m", 30: "g", 40: "t", 50: "p"}.get(exponent)
    if "." not in coefficient and spark_suffix is not None:
        return coefficient + spark_suffix

    total_bytes = math.ceil(float(coefficient) * (2 ** exponent))
    return f"{math.ceil(total_bytes / (2 ** 20))}m"
```

hack/e2e-setup-cluster.sh
```diff
@@ -275,8 +275,11 @@ metadata:
   namespace: $NAMESPACE
 rules:
   - apiGroups: [""]
-    resources: ["pods", "services", "configmaps"]
-    verbs: ["get", "list", "watch", "create", "delete", "patch", "update"]
+    resources: ["pods", "services", "configmaps", "persistentvolumeclaims"]
+    verbs: ["get", "list", "watch", "create", "delete", "deletecollection", "patch", "update"]
```
tariq-hasan: Why would we need `persistentvolumeclaims` and `deletecollection` in the rules?

examples/spark/batch_job_lifecycle.py
```diff
@@ -0,0 +1,253 @@
...
+
+    client.delete_job(JOB_NAME)
+
+    print("Job deleted.")
+
+    try:
+        client.get_job(JOB_NAME)
+        raise RuntimeError("Job still exists after deletion.")
+    except RuntimeError:
+        print("Verified job has been deleted.")
```
tariq-hasan: Catching the exception and printing `"Verified job has been deleted."` would make the test self-defeating.

Same in `batch_failed_job.py`.

tariq-hasan (commented):
Hi @Goku2099! Thanks for adding in the E2E tests. I have included a few additional comments.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Good point. The current implementation relies on the operator's default restart policy (Never), so a FAILED state is terminal for jobs created by the SDK today. Supporting restart policies such as OnFailure would require distinguishing transient failures from terminal failures. I can address that in a follow-up if/when restart policies are exposed through the SDK.

hack/e2e-setup-cluster.sh
Goku2099: I added those while debugging, but I don't think they're actually required anymore. The RBAC permissions we needed were for sparkapplications and pods/log. I'll remove the unnecessary permissions.

examples/spark/batch_job_lifecycle.py
Goku2099: Regarding the filesystem URI examples, I think we can implement the local file E2E once the image/options configuration is available. For this PR, I'll keep the existing HTTPS example but replace the Apache Spark GitHub raw URL with a file that lives under examples/spark in the Kubeflow SDK repository. That way the example no longer depends on an external repository, and we can add the local://, s3://, gs://, and hdfs:// examples in a follow-up PR.

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: Added [issue](https://github.com/kubeflow/sdk/issues/593) to track this item of work.

kubeflow/spark/types/types.py
tariq-hasan: Added [issue](https://github.com/kubeflow/sdk/issues/594) for this item.

examples/spark/batch_job_lifecycle.py
tariq-hasan: Added [issue](https://github.com/kubeflow/sdk/issues/595) for filesystem support.

kubeflow/spark/backends/kubernetes/constants.py
tariq-hasan: This will be covered if necessary as part of https://github.com/kubeflow/sdk/issues/595.

examples/spark/batch_failed_job.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Kubeflow Authors.
```
andreyvelich: Can you remove the YEAR from all licence headers please?

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +138,124 @@ def delete_session(self, name: str) -> None:
...
+    # Spark batch jobs
+    # ------------------------------------------------------------------
+
+    def submit_job(
+        self,
+        job: FileJob | FuncJob,
+        num_executors: int | None = None,
+        resources_per_executor: dict[str, str] | None = None,
+        spark_conf: dict[str, str] | None = None,
+        options: list[Any] | None = None,
```
andreyvelich:
```suggestion
        options: list | None = None,
```

kubeflow/spark/api/spark_client.py
```diff
@@ -141,3 +138,124 @@ def delete_session(self, name: str) -> None:
...
+        Args:
+            name: Job name.
+            status: Target job state(s).
+            timeout: Maximum wait time in seconds. Default 600.
+            polling_interval: Time in seconds between status checks.
+
+        Returns:
+            Spark job information after reaching one of the target statuses.
+        """
+
```
andreyvelich, line 294: Since this PR is merged: https://github.com/kubeflow/sdk/pull/554, please can you add validation for Spark too?

kubeflow/spark/types/types_test.py
```diff
@@ -172,3 +179,160 @@ def test_kep107_example(self):
         assert executor.resources_per_executor["cpu"] == "8"
         assert executor.resources_per_executor["memory"] == "32Gi"
         assert executor.resources_per_executor["nvidia.com/gpu"] == "2"
+
+
+class TestSparkJobStatus:
+    """Tests for SparkJobStatus enum."""
+
+    def test_enum_values(self):
+        """Verify SparkJobStatus enum has expected values."""
```
andreyvelich: @Goku2099 @tariq-hasan Can we ask LLM to refactor this unit tests to follow the TestCase model like in Trainer?
https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/types/types_test.py

andreyvelich (commented):
Thanks for the update @Goku2099, overall lgtm, left a few comments.

kubeflow/spark/types/types_test.py
andreyvelich: @Goku2099 Did you address this comment?

Goku2099: Yes, I refactored the type tests to follow the Trainer TestCase pattern for the dataclass/object creation tests (Driver, Executor, SparkJob, FileJob, FuncJob, etc.). The enum tests (SparkJobStatus and SparkConnectState) remain parameterized since they don't involve object construction. Is there a specific part you think should be refactored further?

andreyvelich:
Thanks for the update @Goku2099!
/lgtm
/assign @tariq-hasan @Shekharrajak @kubeflow/kubeflow-sdk-team @Fiona-Waters @abhijeet-dhumal

tariq-hasan:
No further questions from my side. Thanks @Goku2099!
/lgtm

PR #524, opened 2026-06-07
feat: [KEP-107] Add long-running job submission proposal for SparkClient

Goku2099:
## Summary

This PR adds a new proposal document for long-running Spark job submission in SparkClient.

The proposal extends the SparkClient vision introduced in KEP-107 and focuses on batch-oriented Spark workloads executed through SparkApplication CRDs and the Spark Operator.

## Proposed Features

- `submit_job()` API for Spark job submission
- File-based job submission
- Function-based job submission (future scope)
- Lifecycle management APIs
  - `list_jobs()`
  - `get_job()`
  - `wait_for_job_status()`
  - `get_job_logs()`
  - `delete_job()`
- Job status tracking
- Observability and monitoring integration
- Testing strategy
- Implementation phases
- Examples and notebooks

## Motivation

Running Spark batch workloads on Kubernetes currently requires users to directly manage SparkApplication resources and operational workflows.

This proposal aims to provide a simple Python API that abstracts these operational details while remaining Kubernetes-native and aligned with existing Kubeflow SDK patterns.

## Related

- KEP-107: Spark Client SDK for Kubeflow

### DIFF
--- a/proposals/107-spark-client/README.md
+++ b/proposals/107-spark-client/README.md
@@ -318,7 +318,7 @@ class Executor:
 
 @dataclass
 class FileJob:
-    """File-based Spark application."""
+    """Spark application referenced by a local or remote file source."""
 
     file_source: str
     arguments: Optional[List[str]] = None
@@ -451,56 +451,70 @@ class SparkClient:
     def builder(cls) -> SparkClientBuilder:
         """Create a builder for advanced configuration."""
 
+
     def connect(
         self,
         base_url: Optional[str] = None,
         token: Optional[str] = None,
-        name: Optional[str] = None,
-        app_name: Optional[str] = None,
         num_executors: Optional[int] = None,
         resources_per_executor: Optional[Dict[str, str]] = None,
         spark_conf: Optional[Dict[str, str]] = None,
         driver: Optional[Driver] = None,
         executor: Optional[Executor] = None,
         options: Optional[List] = None,
+        timeout: int = 300,
+        connect_timeout: int = 120,
     ) -> SparkSession:
         """Connect to Spark - unified API for both existing servers and new sessions.
 
         This method supports two modes based on parameters:
-        - **Connect mode**: When `base_url` is provided, connects to an existing Spark Connect server
-        - **Create mode**: When `base_url` is not provided, creates a new Spark Connect session
+
+        - **Connect mode**: When `base_url` is provided, connects to an existing Spark Connect server.
+        - **Create mode**: When `base_url` is not provided, creates a new Spark Connect session.
 
         Args:
-            base_url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
-                 If provided, connects to existing server. If None, creates new session.
-            token: Optional authentication token for existing server.
-            name: Optional session name. Auto-generated if not provided (create mode only).
-            app_name: Optional Spark application name (create mode only).
+            base_url: Optional URL to an existing Spark Connect server
+                (e.g. ``sc://server:15002``). If provided, connects to the
+                existing server. Otherwise, creates a new session.
+            token: Optional authentication token for an existing server.
             num_executors: Number of executor instances (create mode only).
-            resources_per_executor: Resource requirements per executor as dict.
-                Format: `{"cpu": "5", "memory": "10Gi", "cpu_limit": "8", "memory_limit": "12Gi"}` (create mode only).
-            spark_conf: Spark configuration dictionary (create mode only).
-            driver: Driver configuration object (create mode only).
-            executor: Executor configuration object (create mode only).
-            options: List of configuration options (create mode only).
+            resources_per_executor: Resource requirements per executor.
+                Format: ``{"cpu": "5", "memory": "10Gi"}``.
+            spark_conf: Spark configuration dictionary.
+            driver: Driver configuration object.
+            executor: Executor configuration object.
+            options: List of advanced Spark configuration options.
+            timeout: Timeout in seconds to wait for the Spark Connect session
+                to become ready.
+            connect_timeout: Timeout in seconds for establishing the
+                SparkSession connection.
 
         Returns:
-            SparkSession connected to Spark (self-managing).
+            SparkSession connected to Spark.
 
         Examples:
-            spark = client.connect(base_url="sc://server:15002")
 
+            # Connect to an existing Spark Connect server
+            spark = client.connect(
+                base_url="sc://server:15002",
+                token="team-token",
+            )
+
+            # Create a new Spark Connect session
             spark = client.connect(
                 num_executors=5,
                 resources_per_executor={
                     "cpu": "5",
-                    "memory": "10Gi"
+                    "memory": "10Gi",
                 },
-                spark_conf={"spark.sql.adaptive.enabled": "true"}
+                spark_conf={
+                    "spark.sql.adaptive.enabled": "true",
+                },
             )
 
             spark = client.connect()
         """
+
 
     def list_sessions(self) -> List[SparkConnectInfo]:
         """List active Spark Connect sessions."""
@@ -535,7 +549,7 @@ class SparkClient:
 
         This method supports two job types:
 
-        - **FileJob**: Submit an existing Python or JAR application.
+        - **FileJob**:  Submit an existing Spark application referenced by a local or remote file source.
         - **FuncJob**: Submit a Python function as a Spark batch job.
 
         Args:
@@ -554,7 +568,7 @@ class SparkClient:
 
             client.submit_job(
                 job=FileJob(
-                    file_source="./etl.py",
+                    file_source="local:///opt/spark/app/etl.py",
                     arguments=["--date", "2026-06-18"],
                 )
             )
@@ -638,7 +652,7 @@ Spark batch workloads share many characteristics with other long-running Kuberne
 | job=FileJob(...) | File mode | Existing scripts, CI/CD pipelines |
 | job=FuncJob(...) | Function mode | Inline transformations, notebooks |
 
-### Example: File Mode (`file_source`)
+### Example: File Mode (`job=FileJob(...)`)
 
 ```python
 from kubeflow.spark import SparkClient
@@ -658,27 +672,27 @@ client.wait_for_job_status(job_name)
 
 ### File Mode Implementation
 
-File mode supports both local Python files and remote application URIs.
+File mode supports both local and remote application sources.
 
 SparkClient implements long-running batch job execution by translating user-provided job specifications into SparkApplication resources managed by the Spark Operator.
 
-When a local Python file is provided, SparkClient packages the file into a Kubernetes ConfigMap and mounts it into the Spark driver pod at `/opt/spark/app` before creating the SparkApplication resource.
+When `file_source` uses a remote URI (for example `s3a://`, `gs://`, `hdfs://`, or `https://`), SparkApplication references the remote application directly.
+
+When `file_source` uses a local URI (for example `local:///opt/spark/app/etl.py`), the application file is expected to already be available to the SparkApplication, such as through a mounted PersistentVolumeClaim (PVC) or a pre-built container image. SparkClient does not package or upload local files automatically.
 
 ```python
 client.submit_job(
     job=FileJob(
-        file_source="./etl.py",
+        file_source="local:///opt/spark/app/etl.py",
     )
 )
 ```
-
-The submitted SparkApplication references the mounted file through a Spark-compatible local URI:
 
 ```yaml
 mainApplicationFile: local:///opt/spark/app/etl.py
 ```
 
-When a remote URI is provided, SparkClient references the remote resource directly without creating a ConfigMap.
+Remote application sources can be referenced directly:
 
 ```python
 client.submit_job(
@@ -688,21 +702,12 @@ client.submit_job(
 )
 ```
 
-Because Kubernetes ConfigMaps have size limits, local file packaging is intended for small single-file applications. If the local file exceeds the supported ConfigMap size limit (approximately 1 MB), SparkClient will reject the submission and require the application to be provided through a remote URI such as `s3a://`, `gs://`, `hdfs://`, or `https://`.
+SparkClient relies on the Spark Operator's native application submission and dependency management mechanisms. For local application sources, users are responsible for ensuring the referenced file is available to the SparkApplication. This can be achieved through mechanisms such as mounted PVCs or pre-built container images.
 
-> **Limitations (Phase 1 Local File Submission)**
->
-> Local file submission is intended as a lightweight convenience feature and has several limitations:
->
-> - ConfigMap size limit (~1 MB)
-> - Single-file applications only
-> - No automatic dependency bundling
->
-> For larger applications or workloads with additional Python packages, JARs, archives, or dependency requirements, users should provide the application through a remote URI and use the Spark Operator's native dependency management mechanisms (`deps.files`, `deps.pyFiles`, `deps.jars`, etc.).
+Additional dependencies (Python packages, JARs, archives, and other resources) continue to be managed using the Spark Operator's native dependency mechanisms (`deps.files`, `deps.pyFiles`, `deps.jars`, etc.).
 
-This proposal does not replace the Spark Operator's native dependency management features. SparkClient local-file submission only packages the primary application file. Additional dependencies continue to be handled through the SparkApplication dependency mechanisms already supported by the Spark Operator.
 
-### Example: Function Mode (`func`)
+### Example: Function Mode (`job=FuncJob(...)`)
 
 ```python
 from kubeflow.spark import SparkClient
@@ -745,7 +750,7 @@ SparkClient follows a similar approach to TrainerClient for function-based execu
 
 1. The SDK extracts the user function source code and invocation arguments.
 2. The SDK embeds the extracted function source and invocation directly into the `initContainer` command.
-3. When the `initContainer` starts, it reconstructs a standalone Python application (for example `spark_job.py`) in a shared volume mounted by both the `initContainer` and the Spark driver.
+3. When the `initContainer` starts, it reconstructs a standalone Python application (for example `spark_job.py`) in an `emptyDir` volume shared between the `initContainer` and the Spark driver.
 4. The SparkApplication references the generated application through a local Spark URI:
 
 ```yaml
@@ -758,27 +763,26 @@ A simplified SparkApplication illustrates this execution flow:
 
 ```yaml
 spec:
+  volumes:
+    - name: spark-job
+      emptyDir: {}
+
   initContainers:
     - name: prepare-job
+      volumeMounts:
+        - name: spark-job
+          mountPath: /opt/spark/app
       command:
         - bash
         - -c
         - |
-          read -r -d '' SCRIPT << 'EOF'
-          def etl_pipeline(date: str, output_path: str):
-              from pyspark.sql import SparkSession
-
-              spark = SparkSession.builder.getOrCreate()
-              # User Spark transformations...
-              ...
-
-          etl_pipeline(
-              date="2024-01-15",
-              output_path="s3a://data/processed/",
-          )
-          EOF
-
+          ...
           printf "%s" "$SCRIPT" > /opt/spark/app/spark_job.py
+
+  driver:
+    volumeMounts:
+      - name: spark-job
+        mountPath: /opt/spark/app
 
   mainApplicationFile: local:///opt/spark/app/spark_job.py
 ```
@@ -826,7 +830,7 @@ class SparkJobStatus(str, Enum):
 
     CREATED = "Created"
     RUNNING = "Running"
-    COMPLETE = "Complete"
+    COMPLETED = "Completed"
     FAILED = "Failed"
 ```
 
@@ -836,7 +840,7 @@ SparkApplication-specific states are mapped into these SDK-level states to provi
 |------------|-------------------------|
 | CREATED | SUBMITTED |
 | RUNNING | RUNNING, SUCCEEDING, SUSPENDING, SUSPENDED, RESUMING |
-| COMPLETE | COMPLETED |
+| COMPLETED | COMPLETED |
 | FAILED | FAILED, SUBMISSION_FAILED, FAILING, PENDING_RERUN, INVALIDATING, UNKNOWN |
 
 The SDK intentionally exposes a simplified status model in Phase 1. Additional status categories may be introduced in future releases based on user feedback and operational requirements.
--- a/proposals/107-spark-client/README.md
+++ b/proposals/107-spark-client/README.md
@@ -13,7 +13,7 @@ A simple Python SDK to run Spark on Kubernetes. The SDK provides `SparkClient`:
 
 - **`connect()` API** - Creates new Spark Connect sessions or connects to existing servers
 - **Auto-provisions** Spark Connect servers when configuration is provided
-- **Connects** to existing Spark Connect servers when URL is provided
+- **Connects** to existing Spark Connect servers when base URL is provided
 - **Auto-cleans up** resources on exit
 - **Batch job support** - submit and manage SparkApplication jobs via `submit_job()`
 
@@ -66,7 +66,7 @@ The SparkClient SDK is designed for different user personas with varying needs:
 
 ---
 
-## Use Cases
+## User Stories
 
 ### 1. Data Scientist: Quick Data Exploration
 
@@ -102,7 +102,6 @@ spark = client.connect(
         "memory_limit": "40Gi"
     },
     spark_conf={"spark.sql.adaptive.enabled": "true"},
-    app_name="feature-engineering"
 )
 
 raw_data = spark.read.parquet("s3a://data/events/")
@@ -124,7 +123,7 @@ from kubeflow.spark import SparkClient
 
 client = SparkClient()
 spark = client.connect(
-    url="sc://spark-cluster.spark-system.svc:15002",
+    base_url="sc://spark-cluster.spark-system.svc:15002",
     token="team-token",
 )
 
@@ -150,14 +149,25 @@ spark.stop()
 
 ```python
 from kubeflow.spark import SparkClient
+from kubeflow.spark.types.jobs import FileJob
 from kubeflow.common.types import KubernetesBackendConfig
 
 client = SparkClient(backend_config=KubernetesBackendConfig(namespace="etl-jobs"))
 
 job_name = client.submit_job(
-    main_file="s3a://bucket/etl/daily_pipeline.py",
-    arguments=["--date", "2024-01-15", "--output", "s3a://bucket/output/"],
-    name="daily-etl-2024-01-15",
+    job=FileJob(
+        file_source="s3a://bucket/etl/daily_pipeline.py",
+        arguments=[
+            "--date", "2024-01-15",
+            "--output", "s3a://bucket/output/",
+        ],
+    ),
+    spark_conf={
+        "spark.sql.adaptive.enabled": "true",
+    },
+    options=[
+        Name("daily-etl-2024-01-15"),
+    ],
 )
 
 job = client.get_job(job_name)
@@ -220,7 +230,7 @@ spark.stop()
 from kubeflow.spark import SparkClient
 
 client = SparkClient()
-spark = client.connect(url="sc://spark-server:15002")
+spark = client.connect(base_url="sc://spark-server:15002")
 df = spark.read.parquet("s3a://bucket/data/")
 df.show()
 spark.stop()
@@ -251,8 +261,8 @@ spark.stop()
 
 The `connect()` method provides a unified interface for both creating new Spark Connect sessions and connecting to existing servers. The method automatically determines the mode based on the parameters provided:
 
-- **Create Mode**: When `url` is not provided, creates a new Spark Connect session with the specified configuration
-- **Connect Mode**: When `url` is provided, connects to an existing Spark Connect server
+- **Create Mode**: When `base_url` is not provided, creates a new Spark Connect session with the specified configuration
+- **Connect Mode**: When `base_url` is provided, connects to an existing Spark Connect server
 
 This simplification reduces API surface area and makes the SDK easier to use:
 
@@ -266,7 +276,7 @@ spark = client.connect(
     spark_conf={"spark.sql.adaptive.enabled": "true"}
 )
 
-spark = client.connect(url="sc://server:15002")
+spark = client.connect(base_url="sc://server:15002")
 
 spark = client.connect()
 ```
@@ -310,7 +320,7 @@ class Executor:
 class FileJob:
     """File-based Spark application."""
 
-    main_file: str
+    file_source: str
     arguments: Optional[List[str]] = None
     main_class: Optional[str] = None
 
@@ -320,7 +330,7 @@ class FuncJob:
     """Function-based Spark application."""
 
     func: Callable
-    func_args: Optional[Dict[str, Any]] = None
+    func_args: dict | None = None
 ```
 
 ### Options Pattern
@@ -443,7 +453,7 @@ class SparkClient:
 
     def connect(
         self,
-        url: Optional[str] = None,
+        base_url: Optional[str] = None,
         token: Optional[str] = None,
         name: Optional[str] = None,
         app_name: Optional[str] = None,
@@ -457,11 +467,11 @@ class SparkClient:
         """Connect to Spark - unified API for both existing servers and new sessions.
 
         This method supports two modes based on parameters:
-        - **Connect mode**: When `url` is provided, connects to an existing Spark Connect server
-        - **Create mode**: When `url` is not provided, creates a new Spark Connect session
+        - **Connect mode**: When `base_url` is provided, connects to an existing Spark Connect server
+        - **Create mode**: When `base_url` is not provided, creates a new Spark Connect session
 
         Args:
-            url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
+            base_url: Optional URL to existing Spark Connect server (e.g., "sc://server:15002").
                  If provided, connects to existing server. If None, creates new session.
             token: Optional authentication token for existing server.
             name: Optional session name. Auto-generated if not provided (create mode only).
@@ -478,7 +488,7 @@ class SparkClient:
             SparkSession connected to Spark (self-managing).
 
         Examples:
-            spark = client.connect(url="sc://server:15002")
+            spark = client.connect(base_url="sc://server:15002")
 
             spark = client.connect(
                 num_executors=5,
@@ -544,14 +554,14 @@ class SparkClient:
 
             client.submit_job(
                 job=FileJob(
-                    main_file="./etl.py",
+                    file_source="./etl.py",
                     arguments=["--date", "2026-06-18"],
                 )
             )
 
             client.submit_job(
                 job=FileJob(
-                    main_file="s3a://jobs/etl.py",
+                    file_source="s3a://jobs/etl.py",
                     arguments=["--date", "2026-06-18"],
                 )
             )
@@ -576,7 +586,7 @@ class SparkClient:
     def get_job_logs(
         self,
         name: str,
-        container: str = "driver",
+        step: str = "driver",
         follow: bool = False,
     ) -> Iterator[str]:
         """Get logs from a Spark job (driver or executor).
@@ -628,7 +638,7 @@ Spark batch workloads share many characteristics with other long-running Kuberne
 | job=FileJob(...) | File mode | Existing scripts, CI/CD pipelines |
 | job=FuncJob(...) | Function mode | Inline transformations, notebooks |
 
-### Example: File Mode (`main_file`)
+### Example: File Mode (`file_source`)
 
 ```python
 from kubeflow.spark import SparkClient
@@ -638,7 +648,7 @@ client = SparkClient(backend_config=KubernetesBackendConfig(namespace="etl-jobs"
 
 job_name = client.submit_job(
     job=FileJob(
-        main_file="s3a://bucket/etl/daily_pipeline.py",
+        file_source="s3a://bucket/etl/daily_pipeline.py",
         arguments=["--date", "2024-01-15"],
     )
 )
@@ -657,7 +667,7 @@ When a local Python file is provided, SparkClient packages the file into a Kuber
 ```python
 client.submit_job(
     job=FileJob(
-        main_file="./etl.py",
+        file_source="./etl.py",
     )
 )
 ```
@@ -673,7 +683,7 @@ When a remote URI is provided, SparkClient references the remote resource direct
 ```python
 client.submit_job(
     job=FileJob(
-        main_file="s3a://bucket/etl.py",
+        file_source="s3a://bucket/etl.py",
     )
 )
 ```
@@ -700,8 +710,9 @@ from kubeflow.common.types import KubernetesBackendConfig
 from pyspark.sql import SparkSession
 import pyspark.sql.functions as F
 
-def etl_pipeline(spark: SparkSession, date: str, output_path: str):
+def etl_pipeline(date: str, output_path: str):
     """ETL logic with Spark transformations."""
+    spark = SparkSession.builder.getOrCreate()
     df = spark.read.parquet(f"s3a://data/raw/{date}/")
 
     result = (
@@ -732,37 +743,51 @@ Function mode is intended for Python-first workflows where users submit Python c
 
 SparkClient follows a similar approach to TrainerClient for function-based execution.
 
-1. The SDK prepares a Python application from the user function and invocation arguments.
-2. The SDK generates a standalone Python application file (for example `spark_job.py`) containing the function definition and execution logic.
-3. An initContainer writes the generated application file into a shared volume mounted by both the initContainer and Spark driver.
-4. The SparkApplication references the generated file through a local Spark URI:
+1. The SDK extracts the user function source code and invocation arguments.
+2. The SDK embeds the extracted function source and invocation directly into the `initContainer` command.
+3. When the `initContainer` starts, it reconstructs a standalone Python application (for example `spark_job.py`) in a shared volume mounted by both the `initContainer` and the Spark driver.
+4. The SparkApplication references the generated application through a local Spark URI:
 
 ```yaml
 mainApplicationFile: local:///opt/spark/app/spark_job.py
 ```
 
-5. The Spark driver executes the generated application using the normal SparkApplication lifecycle.
+5. The Spark driver executes the generated application using the standard SparkApplication lifecycle.
 
-Conceptually, the generated file contains the user function definition and invocation logic:
+A simplified SparkApplication illustrates this execution flow:
 
-```python
-def etl_pipeline(spark, date, output_path):
-    ...
+```yaml
+spec:
+  initContainers:
+    - name: prepare-job
+      command:
+        - bash
+        - -c
+        - |
+          read -r -d '' SCRIPT << 'EOF'
+          def etl_pipeline(date: str, output_path: str):
+              from pyspark.sql import SparkSession
 
-etl_pipeline(
-    spark,
-    date="2024-01-15",
-    output_path="s3a://data/processed/",
-)
+              spark = SparkSession.builder.getOrCreate()
+              # User Spark transformations...
+              ...
+
+          etl_pipeline(
+              date="2024-01-15",
+              output_path="s3a://data/processed/",
+          )
+          EOF
+
+          printf "%s" "$SCRIPT" > /opt/spark/app/spark_job.py
+
+  mainApplicationFile: local:///opt/spark/app/spark_job.py
 ```
 
-For Phase 2, SparkClient follows the same high-level pattern used by TrainerClient: the generated Python source is written by an initContainer into a shared volume, and the Spark driver executes the generated application from a fixed location such as `/opt/spark/app/spark_job.py`.
+This approach allows users to submit Python functions without manually creating or packaging Spark application files, while remaining aligned with the SparkApplication execution model and the existing TrainerClient implementation pattern.
 
-This approach avoids requiring users to manually package application files while remaining aligned with the SparkApplication execution model and the existing TrainerClient implementation pattern.
+> **Note:** The exact command template and generated wrapper may evolve during implementation, but the overall execution flow remains the same.
 
-> **Note:** The exact code-generation mechanism, packaging workflow, and generated application structure may evolve during implementation, but the overall execution flow remains the same.
-
-Once prepared, the SparkApplication executes the generated application using the same lifecycle management APIs as file-based submissions.
+Once the generated application has been prepared by the `initContainer`, the Spark driver executes it using the same lifecycle management APIs as file-based submissions.
 
 > **Note:** Function mode (`job=FuncJob(...)`) will be available in Phase 2.
 
@@ -773,10 +798,14 @@ Once prepared, the SparkApplication executes the generated application using the
 ```python
 @dataclass
 class SparkJob:
+    """Information about a Spark batch job."""
+
     name: str
-    status: SparkJobStatus
-    application_id: str | None = None
-    error_message: str | None = None
+    namespace: str
+    status: SparkJobStatus | None = None
+    creation_timestamp: datetime | None = None
+    num_executors: int | None = None
+    driver_pod_name: str | None = None
 ```
 
 ---
@@ -820,7 +849,7 @@ The SDK intentionally exposes a simplified status model in Phase 1. Additional s
 |---------|-------------|
 | **Unified `connect()` API** | Single method for both creating sessions and connecting to existing servers |
 | **Auto-provisioning** | Creates Spark Connect server when configuration is provided to `connect()` |
-| **Connect mode** | Connect to existing servers via `connect(url=...)` |
+| **Connect mode** | Connect to existing servers via `connect(base_url=...)` |
 | **Self-managing sessions** | `connect()` returns SparkSession that manages itself |
 | **Full PySpark API** | Returns standard `SparkSession` |
 | **Simplified configuration** | Direct parameters like `num_executors`, `resources_per_executor`, `spark_conf` |
@@ -841,7 +870,7 @@ SparkClient (Stateless)
     │   │
     │   ├── connect() ──► Unified API
     │   │   │
-    │   │   ├── connect(url=...) ──► sc://server:15002 ──► SparkSession
+    │   │   ├── connect(base_url=...) ──► sc://server:15002 ──► SparkSession
     │   │   │
     │   │   └── connect(num_executors=...) ──► SparkConnect CRD ──► Server Pod + Executors
     │   │                                                   │
@@ -927,7 +956,7 @@ client = SparkClient.builder().backend(
 | Phase | Feature | Description |
 |-------|---------|-------------|
 | **Phase 1** | `connect()` (unified) | Unified API: connect to existing servers or create new sessions |
-| **Phase 1** | `connect(url=...)` | Connect to existing Spark Connect servers |
+| **Phase 1** | `connect(base_url=...)` | Connect to existing Spark Connect servers |
 | **Phase 1** | `connect(num_executors=...)` | Auto-provision Spark Connect servers with configuration |
 | **Phase 1** | `submit_job(job=FileJob(...))` |  SparkApplication-based batch job submission and lifecycle APIs |
 | **Phase 2** | `submit_job(job=FuncJob(...))` | Function-based batch job submission |
--- a/proposals/107-spark-client/README.md
+++ b/proposals/107-spark-client/README.md
@@ -3,6 +3,7 @@
 ## Authors
 
 - Shekhar Rajak - [@shekharrajak](https://github.com/shekharrajak)
+- Sameer Yadav - [@Goku2099](https://github.com/Goku2099)
 
 Ref: https://github.com/kubeflow/sdk/issues/107
 
@@ -31,15 +32,144 @@ Running Spark on Kubernetes requires managing complex infrastructure. Users want
 2. Auto-provision Spark Connect servers
 3. Support connecting to existing servers
 4. Full PySpark compatibility
-5. Extensible architecture for future batch job support
-6. Submit and manage batch Spark jobs via SparkApplication CRD
-7. Kubeflow ecosystem integration (Pipelines, Trainer, Spark History MCP Server)
+5. Submit and manage batch Spark jobs via SparkApplication CRD
+6. Provide lifecycle management APIs for Spark jobs
 
 ## Non-Goals
 
 - Supporting Spark outside Kubernetes (local mode, standalone clusters)
 - Managing Spark Operator installation
 - Replacing the Spark Operator
+
+---
+
+
+## User Personas
+
+The SparkClient SDK is designed for different user personas with varying needs:
+
+```
++------------------+     +------------------+     +-------------------+
+|  Data Engineer   |     |  Data Scientist  |     |   ML Engineer     |
++------------------+     +------------------+     +-------------------+
+|                  |     |                  |     |                   |
+| - Batch ETL jobs |     | - Interactive    |     | - Feature eng.    |
+| - Job scheduling |     |   exploration    |     | - Training data   |
+| - Log monitoring |     | - Notebooks      |     | - Batch processing|
+| - Queue routing  |     | - Ad-hoc queries |     | - ETL workflows   |
+|                  |     |                  |     |                   |
++--------+---------+     +--------+---------+     +--------+----------+
+         |                        |                        |
+         v                        v                        v
+    submit_job()            connect()                  Both modes
+```
+
+---
+
+## Use Cases
+
+### 1. Data Scientist: Quick Data Exploration
+
+```python
+from kubeflow.spark import SparkClient
+
+client = SparkClient()
+spark = client.connect()
+
+df = spark.read.parquet("s3a://data/sales/")
+df.printSchema()
+df.describe().show()
+
+result = df.groupBy("product").sum("revenue").orderBy("sum(revenue)", ascending=False)
+result.show(10)
+
+spark.stop()
+```
+
+### 2. ML Engineer: Feature Engineering
+
+```python
+from kubeflow.spark import SparkClient
+from kubeflow.common.types import KubernetesBackendConfig
+
+client = SparkClient(backend_config=KubernetesBackendConfig(namespace="ml-jobs"))
+spark = client.connect(
+    num_executors=20,
+    resources_per_executor={
+        "cpu": "4",
+        "cpu_limit": "8",
+        "memory": "32Gi",
+        "memory_limit": "40Gi"
+    },
+    spark_conf={"spark.sql.adaptive.enabled": "true"},
+    app_name="feature-engineering"
+)
+
+raw_data = spark.read.parquet("s3a://data/events/")
+features = raw_data.select(
+    "user_id",
+    "event_type",
+    F.hour("timestamp").alias("hour"),
+    F.dayofweek("timestamp").alias("day_of_week"),
+)
+features.write.parquet("s3a://data/features/")
+
+spark.stop()
+```
+
+### 3. Platform Engineer: Connect to Shared Cluster
+
+```python
+from kubeflow.spark import SparkClient
+
+client = SparkClient()
+spark = client.connect(
+    url="sc://spark-cluster.spark-system.svc:15002",
+    token="team-token",
+)
+
+spark.sql("SELECT * FROM shared_database.table").show()
+spark.stop()
+```
+
+### 4. Notebook Workflow
+
+```python
+from kubeflow.spark import SparkClient
+
+client = SparkClient()
+spark = client.connect()
+
+df = spark.read.json("s3a://logs/")
+df.groupBy("status_code").count().show()
+
+spark.stop()
+```
+
+### 5. Data Engineer: Batch ETL Job
+
+```python
+from kubeflow.spark import SparkClient
+from kubeflow.common.types import KubernetesBackendConfig
+
+client = SparkClient(backend_config=KubernetesBackendConfig(namespace="etl-jobs"))
+
+job_name = client.submit_job(
+    main_file="s3a://bucket/etl/daily_pipeline.py",
+    arguments=["--date", "2024-01-15", "--output", "s3a://bucket/output/"],
+    name="daily-etl-2024-01-15",
+)
+
+job = client.get_job(job_name)
+print(f"Job submitted: {job_name}")
+print(f"Status: {job.status}")
+
+completed_job = client.wait_for_job_status(job_name, timeout=3600)
+print(f"Final status: {completed_job.status}")
+
+for line in client.get_job_logs(job_name, container="driver"):
+    print(line)
+```
 
 ---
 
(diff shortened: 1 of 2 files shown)


Shekharrajak (commented):
Looks fine for the early stage implementation. 
Lets link this file into original KEP and update about submit job implementation (e.g. this line https://github.com/kubeflow/sdk/blob/main/docs/proposals/107-spark-client/README.md?plain=1#L833)

Shekharrajak (approved):

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+
+Spark jobs transition through a set of SDK-level states that are derived from the underlying SparkApplication status.
+
+| SparkApplication State | SDK State |
+|-----------------------|-----------|
+| Submitted | Submitted |
+| Running | Running |
+| Completed | Completed |
+| Failed | Failed |
+
```
tariq-hasan: The SparkApplication state machine is far richer than what the SDK supports: https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L338-L357.

Do we want the SDK to be a simpler, opinionated interface for job submission (and internally manage a complex SDK-to-control-plane state mapping) or do we want the SDK to accurately reflect the control plane states for the sake of observability?

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+- Spark and Kubeflow integration scenarios
+
+Testing coverage will expand as new SparkClient capabilities are introduced across implementation phases.
+
+### Version Compatibility
+
+Testing should cover supported Spark versions, including:
+
+- Spark 3.5.x
+- Spark 4.x
```
tariq-hasan: We do not have a CI matrix in the SDK workflows for testing different Spark versions at this stage. The default spark operator image is built on a Spark 4.0.1 base image: https://github.com/kubeflow/spark-operator/blob/master/Dockerfile#L17. And the SDK e2e tests build the spark operator image by checking out the operator code from the master branch and building the referenced Dockerfile: https://github.com/kubeflow/sdk/blob/main/.github/workflows/test-spark-examples.yaml#L51-L61.

Same for the SDK unit tests - they are all using the default 4.0.1 image: https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/constants.py#L25.

If we want to test the support for multiple Spark versions then this will need to be a separate work of item in terms of e2e testing - applicable for both Spark interactive connect and job submission workflows.

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+
+| TrainerClient | SparkClient |
+|---------------|-------------|
+| `train()` | `submit_job()` |
+| `list_jobs()` | `list_jobs()` |
+| `get_job()` | `get_job()` |
+| `get_job_logs()` | `get_job_logs()` |
+| `wait_for_job_status()` | `wait_for_job_status()` |
+| `delete_job()` | `delete_job()` |
+
```
tariq-hasan: Other than parity with TrainerClient do we want to also consider parity between interactive connect and job submission workflows?

The following is the signature for the `connect()` function:

```python
    def connect(
        self,
        base_url: str | None = None,
        token: str | None = None,
        spark_conf: dict[str, str] | None = None,
        driver: Driver | None = None,
        options: list | None = None,
        timeout: int = 300,
        connect_timeout: int = 120,
    ) -> SparkSession:
        pass
```

The following is the signature for the proposed `submit_job()`  function in KEP-107:

```python
    def submit_job(
        self,
        func: Optional[Callable[[SparkSession], Any]] = None,
        func_args: Optional[Dict[str, Any]] = None,
        main_file: Optional[str] = None,
        main_class: Optional[str] = None,
        arguments: Optional[List[str]] = None,
        name: Optional[str] = None,
    ) -> str:
        pass
```

A few questions:
- The `num_executors`, `resources_per_executor`, `spark_conf`, `driver`, `executor`, `options`, `timeout` and `connect_timeout` parameters are all unsupported on `submit_job()`. Is this to be expected?
- Instead of directly exposing `func`, `func_args`, `main_file`, `main_class` and `arguments` do we want to expose the submission modes (func or file) more idiomatically by defining dataclasses for the submission modes and exposing a single typed submission mode parameter (similar to how options is structured) in the interest of a better user experience?
- As it stands we are supporting multiple CRDs under a single client. Do we want to have a clean abstraction such that we maintain separate interfaces for the SparkConnect and SparkApplication capabilities with their separate function signatures while still supporting all the methods under the single client?

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+Potential job metadata includes:
+
+- Job name
+- Current status
+- Spark application identifier
+- Creation timestamp
+- Start time
+- Completion time
+- Spark UI URL
+
```
tariq-hasan: I am noticing that this is the current support for `SparkConnectInfo` and `SparkConnectState` in the SDK: https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/types/types.py#L22-L53.

That said the SparkApplication CRD offers a much richer set of info: [SparkApplicationStatus](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L150-L177), [SparkApplicationState](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L338-L363), [DriverState](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L365-L375), [ExecutorState](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L377-L387) and [DriverInfo](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L618-L628).

Since the SDK is meant to opinionated in terms of the capabilities it exposes I am wondering what specific info here would be relevant to expose and not expose. For example, would it be useful to expose [`status.applicationState.errorMessage`](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L362) to enable better observability on the SDK?

tariq-hasan: In addition, the Trainer client provides a first-class [Events API](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L214-L231). Do we want to have similar support in the Spark SDK?

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+### Log Retrieval
+
+SparkClient provides access to Spark driver logs through the get_job_logs() API.
+
+The implementation retrieves logs from the Kubernetes resources associated with the SparkApplication, allowing users to inspect job execution, diagnose failures, and monitor progress without interacting directly with Kubernetes APIs.
+
```
tariq-hasan: The current implementation for [`get_session_logs`](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend.py#L585-L626) uses `read_namespaced_pod_log` for reading the pod logs.

That said, batch jobs eventually run to completion and the resources would then be garbage-collected as per the [TTL setting](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L129-L134).

Since reading logs after completion is a valid use case how should we handle the case when the pod log approach does not return anything since the pods will be garbage-collected once its parent SparkApplication CR is removed?

In other words, do we want to support durable log access for terminal jobs (through Spark History Server or object store) or explicitly scope `get_job_logs` to "while pods exist" and document the limitation?

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+The `wait_for_job_status()` API enables applications and pipelines to block until a job reaches a desired state.
+
+```python
+completed_job = client.wait_for_job_status(
+    job_name,
+    timeout=3600,
+    polling_interval=10,
+)
+```
```
tariq-hasan: Even though it is an example, maybe we should align with both [KEP-107](https://github.com/kubeflow/sdk/blob/main/docs/proposals/107-spark-client/README.md?plain=1#L417-L422) and [TrainerClient](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L233-L240) and use the defaults. What would you think?

```python
completed_job = client.wait_for_job_status(
    name,
    status: SparkJobStatus.COMPLETED,
    timeout: int = 600,
    polling_interval=2
)
```

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+The resulting SparkApplication references the mounted file using a Spark-compatible local path:
+
+```yaml
+mainApplicationFile: local:///opt/spark/app/etl.py
+```
+
+This approach allows Spark workloads running inside Kubernetes to execute application code that originates from a user's local environment.
+
```
tariq-hasan: The operator already has extensive support for [dependencies](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L389-L417).

What reasons would there be to not leverage this capability and instead route the dependency through a configmap?

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,294 @@
...
+### Job Information
+
+The `get_job()` API returns metadata and execution information associated with a Spark job.
+
+Potential job metadata includes:
+
+- Job name
+- Current status
+- Spark application identifier
```
tariq-hasan: The Spark application identifier is eventual metadata and is empty until the driver starts.

Submission ID would be immediate metadata while the job submission is in progress.

Maybe we need to make this explicit.

Ref: https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L155-L158

tariq-hasan (commented):
I have added a few questions for clarification.

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,499 @@
...
+| `main_file=...` | File mode | Local Python files, remote application URIs, CI/CD pipelines |
+| `func=...` | Function mode | Inline transformations, notebooks |
+
+### File Mode (`main_file`)
+
+File mode supports both local Python files and remote application URIs.
+
+When a local Python file is provided, SparkClient automatically packages the file using a Kubernetes ConfigMap before creating the SparkApplication resource. This allows application code from the user's local environment to be executed inside Kubernetes-managed Spark workloads.
+
+When a remote URI is provided, SparkClient directly references the remote resource without creating a ConfigMap.
```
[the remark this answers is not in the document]
tariq-hasan: I believe this needs to be addressed.

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,499 @@
...
+#### Local File Packaging Workflow
+
+When a local Python file is provided, SparkClient automatically packages the file using a Kubernetes ConfigMap before creating the SparkApplication resource.
+
+```python
+client.submit_job(
+    main_file="./etl.py",
+)
+```
+The resulting SparkApplication references the mounted file using a Spark-compatible local path:
```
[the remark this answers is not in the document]
tariq-hasan: I believe this needs to be addressed.

docs/proposals/107-spark-client/long-running-job-submission.md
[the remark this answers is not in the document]
Goku2099: The current implementation already enforces ConfigMap size limits and recommends remote URIs for larger applications. I will update the design doc to explicitly document the packaging constraints and recommended alternatives.

docs/proposals/107-spark-client/long-running-job-submission.md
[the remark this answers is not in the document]
Goku2099: The current implementation creates a ConfigMap-backed volume, mounts it at `/opt/spark/app` in the driver pod, and rewrites the local file path to a Spark-compatible `local:///opt/spark/app/<script>` URI.

I will update the design doc to explicitly document the volume mount, path mapping, and how the packaged file is made available to Spark.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: That's a good point. I simplified the state model in the draft primarily to illustrate the lifecycle at a high level, but I agree the SparkApplication state machine is significantly richer.

I don't have a strong preference here and would like guidance on the desired SDK behavior. Should SparkClient expose a simplified, opinionated state model and internally map SparkApplication states, or should it surface the underlying SparkApplication states directly for observability and debugging purposes?

Once we align on that direction, I will update the proposal accordingly.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: Makes sense. I added the version compatibility note as a testing goal, but you're right that the current SDK test infrastructure is built around the default Spark 4.0.1 image and does not maintain a multi-version CI matrix today.

I will remove this section from the proposal and keep multi-version Spark testing as separate future work rather than part of this design.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: These are good questions.

My original assumption was that `submit_job()` would initially expose only the parameters required for SparkApplication-based batch execution, while keeping the existing `connect()` API focused on Spark Connect sessions. That's why the current proposal doesn't carry over parameters such as `num_executors`, `resources_per_executor`, `options`, and the connection-related settings.

For the submission mode question, I don't have a strong preference between the current overloaded API (`main_file`, `func`, etc.) and a typed submission-mode abstraction. My goal was mainly to align with the direction already outlined in KEP-107, but I'm open to whichever approach provides a cleaner long-term API.

Similarly, regarding separate interfaces for Spark Connect and SparkApplication workflows, I think there are tradeoffs between maintaining a single unified client and introducing more explicit abstractions around the underlying execution models. I did like to hear thoughts from the broader group before locking that into the proposal.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: My intention was not to expose the full SparkApplication status model but rather a simplified SDK-level job information object, similar to `SparkConnectInfo`.

I agree we should be more explicit about which fields are part of the SDK contract. In addition to the basic metadata and timestamps, exposing an error message for failed jobs seems valuable for observability and debugging.

For driver and executor specific details, I was initially thinking of keeping them as implementation details and exposing a higher-level SDK abstraction. However, I can see the value of surfacing additional SparkApplication information for observability use cases. I will update the proposal to clarify the intended SDK-level metadata model, and we can further discuss how much of the underlying SparkApplication state should be exposed.

Goku2099: That's a good point. I hadn't included an Events API in the initial proposal and was primarily focusing on status tracking and log retrieval for observability.

I do think there is value in exposing Kubernetes and SparkApplication events, especially for troubleshooting scenarios where logs alone are insufficient. My preference would be to keep it as a future enhancement rather than part of the initial implementation scope, but I would be interested in adding support for a `get_job_events()` API as the SparkClient observability capabilities evolve.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: The current proposal assumes `get_job_logs()` retrieves logs from the underlying driver pod while the SparkApplication resources still exist.

I had not included durable log storage as part of the initial design. My expectation was that long-term log retention would be handled by existing Spark ecosystem components such as Spark History Server or external storage systems rather than by SparkClient itself.

I will update the proposal to clarify the scope of `get_job_logs()` and document that durable log retention is outside the initial implementation scope. We can consider deeper integration with History Server or external log backends as a future enhancement.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: My motivation for the ConfigMap approach was primarily to support a simple local-file workflow where users can submit a Python script directly from their development environment:

```python
client.submit_job(main_file="./etl.py")
```

without requiring the file to be uploaded to object storage or otherwise made accessible to the cluster beforehand.

The intention was not to replace the Spark Operator dependency mechanisms. Remote URIs and existing Spark dependency features would continue to be supported and remain the recommended approach for larger applications. The ConfigMap packaging path was intended as a convenience feature for smaller workloads and notebook-driven development workflows.

I amhappy to revisit this if leveraging the existing dependency support provides a cleaner design.

docs/proposals/107-spark-client/long-running-job-submission.md
Goku2099: Good catch. I was treating the Spark application identifier as generic job metadata, but you're right that it is only populated after the application starts running.

I will update the proposal to distinguish between the submission identifier, which is available immediately after submission, and the Spark application identifier, which becomes available later in the job lifecycle.

docs/proposals/107-spark-client/long-running-job-submission.md
Shekharrajak: True, our main goal must be running spark job smoothly and parsing result/error as raw as possible from the spark-operator. 

Kubeflow SDK SparkClient must honour all the features of spark-operator and if we need something very specific to kubeflow that must be out of the box available (for now we do not have much wrapper so you can skip - example - when we run sdk in kubeflow pipeline/notebook we might have to make sure sparkclient already have all the available dependencies so that user do not have to worry about kubeflow specific stuff).

Goku2099: Thanks, that makes sense.
I will update the proposal to reflect that direction and avoid over-simplifying the SparkApplication state and metadata model.

tariq-hasan:
We would appreciate any further feedback to finalize the KEP.

/assign @andreyvelich @alimaredia @RobuRishabh @vara-bonthu @nabuskey @astefanutti @kramaranya

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,360 @@
...
+)
+```
+
+### Job Information
+
+The `get_job()` API returns a SDK-level job information object derived from the underlying SparkApplication resource.
+
+The SDK exposes a simplified set of metadata intended for job lifecycle management and observability, while abstracting Kubernetes and SparkApplication implementation details.
+
+Potential fields include:
```
Shekharrajak: We must have all the spark app spec as per https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L30

Shekharrajak:
```
spark_conf     -> spark-submit --conf key=value
  main_file      -> spark-submit final application resource
  arguments      -> argv passed to user Spark application
  driver/executor -> modeled Spark/Kubernetes resource config
  deps           -> jars, files, pyFiles, packages, archives
```

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,360 @@
...
+
+This document focuses on those implementation details and describes how long-running Spark jobs are realized within the SparkClient architecture on Kubernetes.
+
+## Relationship to Existing Kubeflow SDKs
+
+### TrainerClient API Alignment
+
+| TrainerClient | SparkClient |
+|---------------|-------------|
+| `train()` | `submit_job()` |
```
Shekharrajak: Can you write each function's signature and definition ? 

I am thinking like : 


```
submit_job(spec) -> SparkJob
  get_job(name) -> SparkJobStatus
  list_jobs(...) -> list[SparkJob]
  delete_job(name)

```
So something like : 

```
client = SparkClient(
      backend=KubernetesSparkBackend(
          operator="kubeflow",   # kubeflow | apache
          context="prod-us-west-2",
          namespace="tenant-a-spark",
      )
  )

  job = client.submit_job(
      SparkJobSpec(
          name="daily-etl",
          main_file="s3a://jobs/etl.py",
          arguments=["--date", "2026-06-16"],
          spark_conf={
              "spark.sql.adaptive.enabled": "true",
              "spark.jars.packages": "org.apache.hadoop:hadoop-aws:3.4.2",
              "spark.hadoop.fs.s3a.endpoint": "...",
          },
          runtime=SparkRuntime(spark_version="4.1.2", image="apache/spark:4.1.2-scala"),
          driver=SparkPod(service_account="spark-tenant-a", cpu="2", memory="4g"),
          executor=SparkExecutor(instances=20, cpu="4", memory="16g"),
          dependencies=SparkDependencies(
              jars=[],
              py_files=[],
              files=[],
              archives=[],
              packages=["org.apache.iceberg:iceberg-spark-runtime-4.0_2.13:1.11.0"],
          ),
          placement=SparkPlacement(
              tenant="tenant-a",
              region="us-west-2",
              cluster_id="spark-prod-us-west-2",
              namespace="tenant-a-spark",
              queue="tenant-a-prod",
          ),
      )
  )
```

But explore and check if we have all optional, required fields in above definition .

Shekharrajak: Mapping with spark-operator  spark-operator/api/v1beta2/sparkapplication_types.go : 

```
 @dataclass(frozen=True)
  class SparkRuntime:
      spark_version: str
      image: str | None = None
      image_pull_policy: str | None = None
      image_pull_secrets: tuple[str, ...] = ()
      python_version: str | None = None
      scala_version: str | None = None
      jdk_version: str | None = None
```
---

```
 @dataclass(frozen=True)
  class SparkJobSpec:
      name: str | None
      application_type: Literal["Java", "Scala", "Python", "R"]
      main_file: str
      main_class: str | None = None
      arguments: tuple[str, ...] = ()

      runtime: SparkRuntime
      driver: SparkDriver = field(default_factory=SparkDriver)
      executor: SparkExecutor = field(default_factory=SparkExecutor)
      dependencies: SparkDependencies = field(default_factory=SparkDependencies)

      spark_conf: Mapping[str, str] = field(default_factory=dict)
      hadoop_conf: Mapping[str, str] = field(default_factory=dict)

      placement: SparkPlacement = field(default_factory=SparkPlacement)
      scheduling: SparkScheduling = field(default_factory=SparkScheduling)

      restart_policy: Mapping[str, Any] | None = None
      dynamic_allocation: Mapping[str, Any] | None = None
      monitoring: Mapping[str, Any] | None = None
      ttl_seconds: int | None = None

      labels: Mapping[str, str] = field(default_factory=dict)
      annotations: Mapping[str, str] = field(default_factory=dict)
      backend_overrides: Mapping[str, Any] = field(default_factory=dict)

```
---

```
@dataclass(frozen=True)
  class SparkPod:
      cores: int | None = None
      core_request: str | None = None
      core_limit: str | None = None
      memory: str | None = None
      memory_limit: str | None = None
      memory_overhead: str | None = None
      image: str | None = None
      service_account: str | None = None
      java_options: str | None = None

      labels: Mapping[str, str] = field(default_factory=dict)
      annotations: Mapping[str, str] = field(default_factory=dict)
      env: Sequence[Mapping[str, Any]] = ()
      env_from: Sequence[Mapping[str, Any]] = ()

      node_selector: Mapping[str, str] = field(default_factory=dict)
      affinity: Mapping[str, Any] | None = None
      tolerations: Sequence[Mapping[str, Any]] = ()
      priority_class_name: str | None = None
      scheduler_name: str | None = None

      volumes: Sequence[Mapping[str, Any]] = ()
      volume_mounts: Sequence[Mapping[str, Any]] = ()
      config_maps: Sequence[Mapping[str, str]] = ()
      secrets: Sequence[Mapping[str, str]] = ()

      pod_template: Mapping[str, Any] | None = None
      init_containers: Sequence[Mapping[str, Any]] = ()
      sidecars: Sequence[Mapping[str, Any]] = ()
      security_context: Mapping[str, Any] | None = None
      pod_security_context: Mapping[str, Any] | None = None
      termination_grace_period_seconds: int | None = None
```

docs/proposals/107-spark-client/long-running-job-submission.md
```diff
@@ -0,0 +1,360 @@
...
+
+The high-level workflow is:
+```
+                            SparkClient
+                                |
+                                v
+                            submit_job()
+                                |
+                                v
+                           Input Validation
```
Shekharrajak: Pulled some of the validations from other available framework : 
   
  - runtime.spark_version required.
  - main_file required for Kubeflow operator.
  - application_type must be Java, Scala, Python, or R.
  - main_class required for Java/Scala.
  - driver.cores and executor.cores must be >= 1 if set.
  - executor.instances must be >= 1 if set.
  - reject conflicts between typed fields and spark_conf, especially namespace, image, service account, executor instances, driver/executor labels, annotations, scheduler name.

Shekharrajak (changes requested):
The vision for this feature:

  - SparkClient should let users submit, observe, debug, and delete long-running Spark jobs without writing raw CRD YAML.
  - The public API should model Spark concepts, not one operator’s CRD schema.
  - It must preserve full Spark power through spark_conf, dependencies, pod templates, labels, annotations, service accounts, and scheduler options. (start with simple but should be part of design always)
  - It should be multi-tenant and multi-region ready through explicit placement fields: namespace, kube context, tenant, region, cluster, queue. (do not think much for initial phases but should keep in mind)
  - It should be future-proof by using backend adapters for Kubeflow Spark Operator, (future:  Batch Processing Gateway, Livy).
  - It should make simple jobs easy while keeping advanced jobs possible **without SDK redesign.**

docs/proposals/107-spark-client/README.md
```diff
@@ -369,32 +370,65 @@ class SparkClient:
 
     def submit_job(
         self,
-        func: Optional[Callable[[SparkSession], Any]] = None,
-        func_args: Optional[Dict[str, Any]] = None,
-        main_file: Optional[str] = None,
-        main_class: Optional[str] = None,
-        arguments: Optional[List[str]] = None,
-        name: Optional[str] = None,
+        name: Optional[str] = None,
```
Shekharrajak: We are going with primitive datatypes to make it easier for data scientists and ML engineers.

Shekharrajak (approved):
It is looking good for the initial implementation .

docs/proposals/107-spark-client/kep.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+latest-milestone: "TBD"
+
+milestone:
+  alpha:
+    kubeflow-sdk: "TBD"
+  beta:
+    kubeflow-sdk: "TBD"
+  stable:
+    kubeflow-sdk: "TBD"
```
andreyvelich: We don't really have alpha, beta, stable features in Kubeflow SDK.
Let's remove stage, and just update latest-milestone to 0.5.0 or 0.6.0

proposals/107-spark-client/README.md
```diff
@@ -3,6 +3,7 @@
 ## Authors
 
 - Shekhar Rajak - [@shekharrajak](https://github.com/shekharrajak)
+- Sameer Yadav - [@Goku2099](https://github.com/Goku2099)
```
andreyvelich, line 6: Could you also update Goals, since after this change we will support SparkJob submissions: https://github.com/Goku2099/sdk/blob/3a24f651246191f77507f46b27c2bfaac8e6e9f9/docs/proposals/107-spark-client/README.md#goals

Also, you can remove: Kubeflow ecosystem integration (Pipelines, Trainer, Spark History MCP Server), since this is not part of this KEP.

proposals/107-spark-client/README.md
```diff
@@ -686,7 +839,7 @@ SparkClient (Stateless)
     │   ├── get_session_logs(name) ──► Iterator[str]
     │   └── delete_session(name)
     │
-    └── Batch Jobs (managed by name)
```
andreyvelich, line 689: Please move [User Stories](https://github.com/Goku2099/sdk/blob/3a24f651246191f77507f46b27c2bfaac8e6e9f9/docs/proposals/107-spark-client/README.md#use-cases) to the top of this document, where we will share how the API can be used.
See example from KFP: https://github.com/Goku2099/sdk/tree/3a24f651246191f77507f46b27c2bfaac8e6e9f9/docs/proposals/125-kfp-client#user-stories

docs/proposals/107-spark-client/README.md
andreyvelich: @Shekharrajak Can we simplify API as much as possible in the first version?
Similar to connect API and `train()` API, can we just have this:
```py
    def submit_job(
        self,
        func,
        func_args,
        spark_conf: dict[str, str] | None = None,
        driver: Driver | None = None,
        options: list | None = None,
    ) 
```
All other parameters, we can move to options.

If we want to allow submit job from image, we might want to design multiple options to configure input assets, like in [TrainerClient](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/api/trainer_client.py#L110-L113)

Maybe like this:
```py
def submit_job(
    self,
    job: FuncJob | FileJob,
):
FuncJob # to define Spark transformations in the function
FileJob # to define Spark transformations in the file (e.g. s3a://)
```

Also, see: https://github.com/kubeflow/sdk/pull/308
cc @tariq-hasan @szaher

docs/proposals/107-spark-client/README.md
```diff
@@ -642,11 +724,82 @@ job_name = client.submit_job(
 
 client.wait_for_job_status(job_name)
 ```
+### Function Mode Implementation
+
+Function mode is intended for Python-first workflows where users submit Python callables instead of application files.
+
+In this mode, SparkClient packages the user function and its dependencies into an executable artifact before SparkApplication creation. An initContainer may be used to prepare and stage the artifact before application startup, while pod templates can be used to customize the Spark driver and executor pods.
+
+Once prepared, the SparkApplication references the generated artifact and executes it using the same lifecycle management APIs as file-based submissions.
```
andreyvelich: Please explain how initContainer approach will work, and how we will create localFile for driver to read the file. Ref: https://github.com/kubeflow/sdk/pull/163#discussion_r2561008012

docs/proposals/107-spark-client/README.md
```diff
@@ -642,11 +724,82 @@ job_name = client.submit_job(
...
+    FAILED = "Failed"
+    SUBMISSION_FAILED = "SubmissionFailed"
+    PENDING_RERUN = "PendingRerun"
+    INVALIDATING = "Invalidating"
+    SUCCEEDING = "Succeeding"
+    FAILING = "Failing"
+    SUSPENDING = "Suspending"
+    SUSPENDED = "Suspended"
+    RESUMING = "Resuming"
+    UNKNOWN = "Unknown"
```
andreyvelich: Do we need all of those statues?
Can we define just this in a phase 1:
```
Created
Running
Complete
Failed
```
To be consistent with TrainJobs: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/constants/constants.py#L42

docs/proposals/107-spark-client/README.md
```diff
@@ -642,11 +724,82 @@ job_name = client.submit_job(
...
+
+- Spark application identifier
+- Spark UI URL
+- History Server URL
+
+### Log Retrieval
+
+SparkClient provides access to Spark driver logs through the `get_job_logs()` API.
+
+Logs are retrieved from the Kubernetes resources associated with the SparkApplication, allowing users to inspect job execution and diagnose failures without interacting directly with Kubernetes APIs.
```
andreyvelich: Let's remove since this is out of scope?
@XploY04 is working on OTel integration as part of this KEP: https://github.com/kubeflow/sdk/pull/382
We can come back to observability discussions there.

andreyvelich (commented):
Thanks for this work @Goku2099! I left few thoughts
/cc @vara-bonthu @nabuskey @franciscojavierarceo @roburishabh 
Appreciate your review too.

Shekharrajak:
> @Shekharrajak Can we simplify API as much as possible in the first version?
Similar to connect API and train() API, can we just have this:

Our thought was to make sure public API is modular enough to take basic to advance spark specs, confs and any upgrade in spark-operator will be available in sdk sparkclient on the fly: 

```
for connect():

  spark = client.connect(
      num_executors=5,
      resources_per_executor={"cpu": "4", "memory": "16Gi"},
      spark_conf={"spark.sql.adaptive.enabled": "true"},
      driver={"cores": 1, "memory": "2Gi"},
      executor={"cores": 4, "memory": "16Gi"},
      options={"name": "interactive-session"},
  )

  For submit_job():

  job_name = client.submit_job(
      job={
          "type": "Python",
          "main_file": "s3a://bucket/jobs/etl.py",
          "arguments": ["--date", "2026-06-20"],
          "image": "docker.io/library/spark:4.0.1",
          "spark_version": "4.0.1",
      },
      num_executors=5,
      resources_per_executor={"cpu": "4", "memory": "16Gi"},
      spark_conf={"spark.sql.adaptive.enabled": "true"},
      driver={"cores": 1, "memory": "2Gi"},
      executor={"cores": 4, "memory": "16Gi"},
      options={"name": "daily-etl"},
  )

```


cc @andreyvelich @tariq-hasan @sameerdattav

proposals/107-spark-client/README.md
```diff
@@ -642,11 +693,50 @@ job_name = client.submit_job(
...
+SparkClient provides lifecycle management APIs for monitoring and tracking Spark jobs submitted through the Spark Operator.
+
+Job state information is derived from the underlying SparkApplication resource while exposing a consistent Python interface.
+
+### Application States
+
+SparkClient derives job state information from SparkApplication status.
+
+```python
+class SparkJobStatus(str, Enum):
```
RobuRishabh, line 827: The SparkApplication CRD has 13 states (SUBMITTED, RUNNING, COMPLETED, FAILED, SUBMISSION_FAILED, PENDING_RERUN, INVALIDATING, SUCCEEDING, FAILING, SUSPENDING, SUSPENDED, RESUMING, UNKNOWN). Can we add an explicit mapping table showing which SparkApplication states map to each of these 4 SDK states? For example, does SUBMISSION_FAILED map to Failed? Where does SUSPENDED go? This mapping is critical for implementers and users debugging state transitions. Ref: [spark-operator types](https://github.com/kubeflow/spark-operator/blob/master/api/v1beta2/sparkapplication_types.go#L338-L357)"

proposals/107-spark-client/README.md
```diff
@@ -642,11 +693,50 @@ job_name = client.submit_job(
...
+
+An initContainer writes the generated application into a shared volume before Spark driver startup. The Spark driver container mounts the same volume and executes the generated application through a local Spark URI:
+
+```yaml
+mainApplicationFile: local:///opt/spark/app/spark_job.py
+```
+
+This approach allows users to submit Python functions without rebuilding container images or manually creating application files. Pod templates may be used to inject the initContainer and shared volume configuration into the SparkApplication driver and executor pods.
+
+Once prepared, the SparkApplication executes the generated application using the same lifecycle management APIs as file-based submissions.
```
RobuRishabh: Should we here elaborate on the serialization mechanism here? Specifically:
1. How is the user function serialized -- cloudpickle, dill, or a custom approach? 
2. How are the function's closure variables and dependencies captured? 
3. What does the generated spark_job.py wrapper look like? 
Even though this is Phase 2 (as noted [here](https://github.com/kubeflow/sdk/pull/524/files#diff-4715d58efaf92eebc6656b58201fc98804a1ed81919a77f821c23eddb715045bR714)), andreyvelich [asked about the initContainer details](https://github.com/kubeflow/sdk/pull/524#discussion_r3443011489) in his review (also ref: [#163 discussion](https://github.com/kubeflow/sdk/pull/163#discussion_r2561008012)) and having a concrete design would help validate feasibility.

proposals/107-spark-client/README.md
```diff
@@ -541,78 +628,42 @@ client = SparkClient(backend_config=KubernetesBackendConfig(namespace="etl-jobs"
...
-    arguments=["--date", "2024-01-15"],
+client.submit_job(
+    main_file="s3a://bucket/etl.py",
 )
-
-client.wait_for_job_status(job_name)
 ```
 
+Because Kubernetes ConfigMaps have size limits, local file packaging is intended for small single-file applications. If the local file exceeds the supported ConfigMap size limit (approximately 1 MB), SparkClient will reject the submission and require the application to be provided through a remote URI such as `s3a://`, `gs://`, `hdfs://`, or `https://`.
+
```
RobuRishabh: Nit: Consider calling out these limitations more prominently (e.g., a dedicated 'Limitations' subsection or a note block): 
(1) ~1MB size limit, 
(2) single-file only, 
(3) no dependency bundling. 
Also, could you clarify the relationship to the operator's native dependency support (deps.files, deps.pyFiles, etc.)?

docs/proposals/107-spark-client/README.md
RobuRishabh: Should the docstring explicitly note that logs are only available while the driver pod exists? After TTL-based garbage collection, get_job_logs() would return empty or raise an error.

proposals/107-spark-client/README.md
```diff
@@ -642,11 +693,50 @@ job_name = client.submit_job(
...
+This approach allows users to submit Python functions without rebuilding container images or manually creating application files. Pod templates may be used to inject the initContainer and shared volume configuration into the SparkApplication driver and executor pods.
+
+Once prepared, the SparkApplication executes the generated application using the same lifecycle management APIs as file-based submissions.
 
 > **Note**: Function mode (`func=...`) will be available in Phase 2.
 
 ---
 
+## Status Model
```
RobuRishabh, line 816: Can we defining the job info model that get_job() returns. At minimum, it would be helpful to include an error_message field that surfaces SparkApplicationStatus.applicationState.errorMessage, this would be critical for debugging failed jobs. exposing submissionID (immediate metadata) vs sparkApplicationId (eventual, only available after driver starts). Can we add a small dataclass or field listing here?

proposals/107-spark-client/README.md
Goku2099: At the moment the Function Mode section is intended to illustrate the expected execution flow rather than define the final implementation details.

The serialization mechanism, dependency handling, and generated wrapper structure have not been finalized yet and would likely need a dedicated design discussion before implementation. Since Function Mode is currently planned as a later phase, I was trying to avoid over-specifying details that may change.

Would it make more sense to keep this section at a high level and explicitly call out that the serialization and packaging strategy will be defined as part of the Function Mode design, or would you prefer a more concrete proposed approach in the KEP?

proposals/107-spark-client/README.md
Goku2099: Good point. The intent was for Phase 1 to provide a lightweight local-file submission mechanism rather than a full packaging solution. I will add a dedicated limitations note calling out the ConfigMap size constraint, single-file scope, and lack of dependency bundling. 
I will also clarify that this proposal does not replace the Spark Operator's native dependency management features (deps.files, deps.pyFiles, deps.jars, etc.), which remain available for remote submissions and advanced dependency handling.

docs/proposals/107-spark-client/README.md
Goku2099: Good point I will update the docstring to make that behavior explicit.

proposals/107-spark-client/README.md
Goku2099: The original proposal exposed the raw SparkApplication states directly, but after the recent feedback we moved toward a simplified status model. While thinking through the mapping table, I noticed that some of the SparkApplication states do not map naturally into Created / Running / Complete / Failed. For example, PENDING_RERUN is not really a terminal failure, SUSPENDED is not actively running, and UNKNOWN is different from a failed state.

@RobuRishabh @andreyvelich @Shekharrajak @tariq-hasan  Would a middle-ground model make more sense here? For example:

CREATED       ---> NEW, SUBMITTED
RUNNING       ---> RUNNING, SUCCEEDING
SUSPENDED     ---> SUSPENDING, SUSPENDED, RESUMING
PENDING_RERUN ---> PENDING_RERUN
COMPLETE      ---> COMPLETED
FAILED        ---> FAILED, SUBMISSION_FAILED, FAILING, INVALIDATING
UNKNOWN       ---> UNKNOWN

This is still significantly simpler than exposing all SparkApplication states directly, while preserving some of the operational states that seem important for debugging and monitoring. Curious what you think.

proposals/107-spark-client/README.md
Goku2099: Good point 
The current KEP defines the lifecycle APIs but does not define the shape of the SparkJob object returned by get_job() and list_jobs(). Adding a small dataclass or field listing would make the contract much clearer. I agree that fields such as error_message and application_id are important for debugging and operational visibility. I will add a minimal SparkJob model to the KEP and document the key fields returned by the lifecycle APIs.

docs/proposals/107-spark-client/README.md
Goku2099: I reviewed the proposed `FileJob` / `FuncJob` approach against the SparkApplicationSpec fields currently supported by the Spark Operator.

From my analysis, the simplified API can still support the existing SparkApplication feature set. Core submission inputs can be represented through the job definition, while advanced SparkApplication configuration can be exposed through the existing Options framework. I did not find any SparkApplication capabilities that become fundamentally inaccessible with this approach.

The main tradeoff is that advanced features (dependencies, monitoring, dynamic allocation, scheduling, image configuration, restart policies, etc.) would require either dedicated Option implementations or a more generic SparkApplication override mechanism.

@andreyvelich  Based on this analysis, are you comfortable with us adopting the simplified `FileJob` / `FuncJob` API direction and updating the KEP accordingly? We'd like to get alignment on the public API before proceeding further with the implementation.

Goku2099: I took another pass at the proposed `FileJob` / `FuncJob` API and tried mapping the current SparkApplicationSpec fields to understand where each configuration would live.

```python
job = client.submit_job(
    job=FileJob(
        main_file="s3a://bucket/jobs/etl.py",
        arguments=["--date", "2026-06-20"],
        # image?
        # spark_version?
        # python_version?
        # main_class?
    ),
    num_executors=5,
    resources_per_executor={
        "cpu": "4",
        "memory": "16Gi",
    },
    options={
        "spark_conf": {...},
        "hadoop_conf": {...},
        "deps": {...},
        "driver": {...},
        "executor": {...},
        "restart_policy": {...},
        "dynamic_allocation": {...},
        "monitoring": {...},
        "image_pull_secrets": [...],
        "batch_scheduler": "...",
        "spark_ui_options": {...},
        "time_to_live_seconds": 3600,
        "volumes": [...],
    },
)
```
The remaining design question for me is where we want fields such as image, spark_version, python_version, and main_class to live:

- inside FileJob / FuncJob
- or inside options

@andreyvelich Is this aligned with your vision for the simplified API?

proposals/107-spark-client/README.md
andreyvelich: I would start with less conditions in the phase 1, collect users feedback, and extend it in the future if needed.

docs/proposals/107-spark-client/README.md
andreyvelich: Sure, that looks good to me. We probably want to move spark_conf outside of options, like we did for [`connect()` API](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/api/spark_client.py#L49).

Can we describe min set of parameters to allow users submit Spark jobs via Function or File @Goku2099?

proposals/107-spark-client/README.md
RobuRishabh: I think a middle ground would work well here. Since @andreyvelich https://github.com/kubeflow/sdk/pull/524#discussion_r3443011489 asked for details on the initContainer mechanism and how the localFile is created, having at least a proposed approach would help validate feasibility without locking in implementation details.

For example: naming the serialization library (cloudpickle is the natural choice since PySpark already uses it for UDFs), a short pseudocode sketch of the generated wrapper, and a paragraph describing the initContainer -> shared emptyDir volume -> driver mount flow. This can still be marked as "proposed" with a note that final details will be refined during Phase 2 implementation.

@andreyvelich any suggestions?

andreyvelich: For phase 1, we can just wrap the users' function in initContainer args, like we do in Trainer: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/utils.py#L365-L366
So we don't need to deal with serialization.


Here is the template: https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/constants/constants.py#L138

1. We write users code into tmp file
2. Execute this file using `torchrun` or `mpirun`.

We can do similar flow for Spark, but instead of saving file to a tmp location, we always going to store it at the same place.

docs/proposals/107-spark-client/README.md
Goku2099: @andreyvelich  Based on the discussion so far, I think the minimum API surface could be:

For file-based jobs:
```python
client.submit_job(
    job=FileJob(
        main_file="s3a://bucket/jobs/etl.py",
        arguments=["--date", "2026-06-20"],
    ),
    num_executors=5,
    resources_per_executor={
        "cpu": "4",
        "memory": "16Gi",
    },
    spark_conf={...},
    options=[...],
)
```
For function-based jobs:
```python
client.submit_job(
    job=FuncJob(
        func=etl_pipeline,
        func_args=["2026-06-20"],
    ),
    num_executors=5,
    resources_per_executor={
        "cpu": "4",
        "memory": "16Gi",
    },
    spark_conf={...},
    options=[...],
)
```
Where:
- job defines the workload (Function or File).
- num_executors and resources_per_executor define compute requirements.
- spark_conf contains common Spark configuration.
- options provides extensibility for advanced SparkApplication-specific configuration.

@andreyvelich Is this closer to the minimal public API surface you had in mind for Phase 1?

proposals/107-spark-client/README.md
Goku2099: I will keep the simplified status model and add an explicit mapping table showing how SparkApplication states map to the SDK statuses.

Goku2099:
@andreyvelich 
I pushed a revised KEP addressing the recent review comments and API discussion. When you have a moment, could you take another look and let me know if we're aligned on the direction so we can proceed with implementation?

RobuRishabh (approved):

proposals/107-spark-client/README.md
```diff
@@ -43,6 +43,136 @@ Running Spark on Kubernetes requires managing complex infrastructure. Users want
...
+|                  |     |                  |     |                   |
++--------+---------+     +--------+---------+     +--------+----------+
+         |                        |                        |
+         v                        v                        v
+    submit_job()            connect()                  Both modes
+```
+
+---
+
+## Use Cases
```
andreyvelich:
```suggestion
## User Stories
```

proposals/107-spark-client/README.md
```diff
@@ -43,6 +43,136 @@ Running Spark on Kubernetes requires managing complex infrastructure. Users want
...
+spark = client.connect(
+    num_executors=20,
+    resources_per_executor={
+        "cpu": "4",
+        "cpu_limit": "8",
+        "memory": "32Gi",
+        "memory_limit": "40Gi"
+    },
+    spark_conf={"spark.sql.adaptive.enabled": "true"},
+    app_name="feature-engineering"
```
andreyvelich: This doesn't exist

```suggestion
```

proposals/107-spark-client/README.md
```diff
@@ -43,6 +43,136 @@ Running Spark on Kubernetes requires managing complex infrastructure. Users want
...
+### 3. Platform Engineer: Connect to Shared Cluster
+
+```python
+from kubeflow.spark import SparkClient
+
+client = SparkClient()
+spark = client.connect(
+    url="sc://spark-cluster.spark-system.svc:15002",
+    token="team-token",
+)
```
andreyvelich, line 128: @Goku2099 Could you update these user stories to follow the current design of connect() API?

proposals/107-spark-client/README.md
```diff
@@ -43,6 +43,136 @@ Running Spark on Kubernetes requires managing complex infrastructure. Users want
...
+from kubeflow.spark import SparkClient
+from kubeflow.common.types import KubernetesBackendConfig
+
+client = SparkClient(backend_config=KubernetesBackendConfig(namespace="etl-jobs"))
+
+job_name = client.submit_job(
+    main_file="s3a://bucket/etl/daily_pipeline.py",
+    arguments=["--date", "2024-01-15", "--output", "s3a://bucket/output/"],
+    name="daily-etl-2024-01-15",
+)
```
andreyvelich, line 171: Same suggestion here.

proposals/107-spark-client/README.md
```diff
@@ -175,6 +305,22 @@ class Executor:
     num_instances: Optional[int] = None
     resources_per_executor: Optional[Dict[str, str]] = None
     java_options: Optional[str] = None
+
+@dataclass
+class FileJob:
+    """File-based Spark application."""
+
+    main_file: str
```
andreyvelich: Maybe we should use file_source here, so we can support multiple sources in the future like S3.

proposals/107-spark-client/README.md
```diff
@@ -175,6 +305,22 @@ class Executor:
...
+    arguments: Optional[List[str]] = None
+    main_class: Optional[str] = None
+
+
+@dataclass
+class FuncJob:
+    """Function-based Spark application."""
+
+    func: Callable
+    func_args: Optional[Dict[str, Any]] = None
```
andreyvelich:
```suggestion
    func_args: dict | None = None
```

proposals/107-spark-client/README.md
```diff
@@ -412,7 +579,13 @@ class SparkClient:
         container: str = "driver",
```
andreyvelich: Any thoughts on using `step` here similar to Trainer? So this can be either `driver` or `executor`.
I feel like container might be unknown term for Data Scientists.

```suggestion
        step: str = "driver",
```

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-from kubeflow.spark import SparkClient
+File mode supports both local Python files and remote application URIs.
 
-client = SparkClient()
-spark = client.connect(
-    url="sc://spark-cluster.spark-system.svc:15002",
-    token="team-token",
+SparkClient implements long-running batch job execution by translating user-provided job specifications into SparkApplication resources managed by the Spark Operator.
+
+When a local Python file is provided, SparkClient packages the file into a Kubernetes ConfigMap and mounts it into the Spark driver pod at `/opt/spark/app` before creating the SparkApplication resource.
```
andreyvelich: What is the user value to support this if we have FuncJob for such use-cases?

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-### 5. Data Engineer: Batch ETL Job
+### Example: Function Mode (`func`)
 
 ```python
 from kubeflow.spark import SparkClient
 from kubeflow.common.types import KubernetesBackendConfig
+from pyspark.sql import SparkSession
+import pyspark.sql.functions as F
+
+def etl_pipeline(spark: SparkSession, date: str, output_path: str):
```
andreyvelich: Why SparkSession is the input argument for the function?
I thought that users have to start SparkSession in their function before write Spark Transformations.

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-for line in client.get_job_logs(job_name, container="driver"):
-    print(line)
-```
+SparkClient follows a similar approach to TrainerClient for function-based execution.
 
-### 6. ML Engineer: Feature Pipeline with Kubeflow Trainer
+1. The SDK prepares a Python application from the user function and invocation arguments.
+2. The SDK generates a standalone Python application file (for example `spark_job.py`) containing the function definition and execution logic.
+3. An initContainer writes the generated application file into a shared volume mounted by both the initContainer and Spark driver.
+4. The SparkApplication references the generated file through a local Spark URI:
```
andreyvelich: Actually, this is now correct, since SDK just extracts the users' function code and write it into `container` command: https://github.com/Goku2099/sdk/blob/b675cc7a95754750b7003ed948506ef24af76e66/kubeflow/trainer/backends/kubernetes/utils.py#L324

@Goku2099 Could you modify those steps, and add simple SparkApplication YAML example of how we embed the users' function into `initContainer` command?

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
+> **Note:** The exact code-generation mechanism, packaging workflow, and generated application structure may evolve during implementation, but the overall execution flow remains the same.
 
-| Parameter | Mode | Use Case |
-|-----------|------|----------|
-| `main_file=...` | File mode | Existing scripts, CI/CD pipelines |
-| `func=...` | Function mode | Inline transformations, notebooks |
+Once prepared, the SparkApplication executes the generated application using the same lifecycle management APIs as file-based submissions.
 
-### Example: File Mode (`main_file`)
+> **Note:** Function mode (`job=FuncJob(...)`) will be available in Phase 2.
```
andreyvelich, line 795: Any objections to implement FuncJob first @Goku2099 @tariq-hasan @Shekharrajak? I feel like it is much simpler experience for users to try out the SDK? Similar to how we give them simple experience to scale PyTorch on Kubernetes via CustomTrainer.

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-)
-
-client.wait_for_job_status(job_name)
+```python
+@dataclass
+class SparkJob:
+    name: str
+    status: SparkJobStatus
+    application_id: str | None = None
+    error_message: str | None = None
```
andreyvelich: Do we need this ?

```suggestion
```

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-    arguments=["--date", "2024-01-15"],
-)
-
-client.wait_for_job_status(job_name)
+```python
+@dataclass
+class SparkJob:
+    name: str
+    status: SparkJobStatus
+    application_id: str | None = None
```
andreyvelich: What is application_id here?

proposals/107-spark-client/README.md
```diff
@@ -425,225 +598,219 @@ class SparkClient:
...
-    main_file="s3a://bucket/etl/daily_pipeline.py",
-    arguments=["--date", "2024-01-15"],
-)
-
-client.wait_for_job_status(job_name)
+```python
+@dataclass
+class SparkJob:
+    name: str
+    status: SparkJobStatus
```
andreyvelich: We can also add:
```
creation_timestamp
num_executors
```

andreyvelich (commented):
Thanks for the updates @Goku2099, i left a few thoughts.

proposals/107-spark-client/README.md
Goku2099: The motivation is to support users who already have existing Spark applications as Python files and want to submit them without refactoring them into `FuncJob`s.

Supporting local files also provides a smooth development workflow. Users can iterate on a local Spark application and submit it directly during development, then use the same `FileJob` abstraction with a remote source (for example `s3a://`, `gs://`, or `hdfs://`) in production. This avoids requiring users to upload the application to remote storage during every development iteration while keeping the submission API consistent across environments.

proposals/107-spark-client/README.md
Goku2099: I agree that having users create the `SparkSession` inside the function is more aligned with the standard PySpark programming model and keeps the function self-contained. I will update the example accordingly. Thanks!

proposals/107-spark-client/README.md
Goku2099: I will update the implementation steps to reflect the actual flow where the SDK embeds the extracted function into the initContainer command, and I will add a simplified SparkApplication YAML snippet showing how the application file is reconstructed before the driver starts.

proposals/107-spark-client/README.md
Goku2099: I don't have any objections. Would it be okay to keep the implementation order as currently planned? The lifecycle APIs and most of the FileJob work are already in place, and FuncJob is planned immediately after that . If that sounds reasonable, I did implement FuncJob right after the lifecycle APIs.

proposals/107-spark-client/README.md
Goku2099: I think both fields are useful for monitoring and lifecycle management. I will add creation_timestamp and num_executors to the SparkJob model.

proposals/107-spark-client/README.md
Goku2099: I intended this to represent the Spark application ID reported by the underlying Spark application, but I agree the name is ambiguous. Since it isn't essential for the initial API, I'll remove it from the SparkJob model for now and revisit it later if we identify a concrete use case.

proposals/107-spark-client/README.md
Goku2099: I agree it's not essential for the initial API. I will remove it and keep the SparkJob model minimal for Phase 1.

proposals/107-spark-client/README.md
Goku2099: done

proposals/107-spark-client/README.md
Goku2099: done

proposals/107-spark-client/README.md
andreyvelich: @Goku2099 But in the KEP you are talking about ConfigMap approach, why it is needed?
If we want to support `local://` as the file_source in `FuncJob`, users should be responsible that the appropriate file will be mounted to the SparkApplication via PVC or pre-build Image.

proposals/107-spark-client/README.md
```diff
@@ -425,225 +608,238 @@ class SparkClient:
...
-print(f"Final status: {completed_job.status}")
+Function mode is intended for Python-first workflows where users submit Python callables instead of application files.
 
-for line in client.get_job_logs(job_name, container="driver"):
-    print(line)
+SparkClient follows a similar approach to TrainerClient for function-based execution.
+
+1. The SDK extracts the user function source code and invocation arguments.
+2. The SDK embeds the extracted function source and invocation directly into the `initContainer` command.
+3. When the `initContainer` starts, it reconstructs a standalone Python application (for example `spark_job.py`) in a shared volume mounted by both the `initContainer` and the Spark driver.
```
andreyvelich: We should use EmptyDir for shared volume between InitContainer and main Container, right?
https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

proposals/107-spark-client/README.md
```diff
@@ -425,225 +608,238 @@ class SparkClient:
...
+@dataclass
+class SparkJob:
+    """Information about a Spark batch job."""
+
+    name: str
+    namespace: str
+    status: SparkJobStatus | None = None
+    creation_timestamp: datetime | None = None
+    num_executors: int | None = None
+    driver_pod_name: str | None = None
```
andreyvelich, line 811: @Shekharrajak @Goku2099 Do we also want to change `pod_name` in the `SparkConnectInfo` object?
https://github.com/Goku2099/sdk/blob/2d95c1b41f426c852aa257bd5f4b457102e13f6c/kubeflow/spark/types/types.py#L41

Shekharrajak: Are you asking if we should be having `pod_name` argument or not ? or asking to use the same argument key for consistency ? 
If it is about consistency then lets use driver_pod_name in both places.

proposals/107-spark-client/README.md
Goku2099: I see what you mean. My intention behind the ConfigMap approach was to simplify the development workflow for local Python files, but I agree that it adds extra implementation complexity to the SDK. I am happy to simplify the proposal by treating local:/// as a file already available to the SparkApplication, with users responsible for mounting it via a PVC or including it in a pre-built image. I'll update the KEP accordingly.

proposals/107-spark-client/README.md
Goku2099: I will update the KEP to explicitly mention emptyDir and reflect it in the example SparkApplication YAML.

proposals/107-spark-client/README.md
Goku2099: I agree with Shekhar that driver_pod_name is more consistent with the SparkJob model. I will update SparkConnectInfo to use the same naming.

andreyvelich: Yes, let's use `driver_pod_name` everywhere.

andreyvelich: @Goku2099 Please could you open separate PR to change the name for SparkConnectInfo too?

proposals/107-spark-client/README.md
andreyvelich: Yes, let's start with that, and extend the `FileJob` with ConfigMap approach later, if users really need it.

proposals/107-spark-client/README.md
```diff
@@ -175,6 +315,22 @@ class Executor:
     num_instances: Optional[int] = None
     resources_per_executor: Optional[Dict[str, str]] = None
     java_options: Optional[str] = None
+
+@dataclass
+class FileJob:
+    """Spark application referenced by a local or remote file source."""
+
+    file_source: str
+    arguments: Optional[List[str]] = None
```
andreyvelich: let's use just args, to be consistent with other APIs?

```suggestion
    args: Optional[List[str]] = None
```

proposals/107-spark-client/README.md
```diff
@@ -425,225 +622,228 @@ class SparkClient:
...
+        - bash
+        - -c
+        - |
+          ...
+          printf "%s" "$SCRIPT" > /opt/spark/app/spark_job.py
+
+  driver:
+    volumeMounts:
+      - name: spark-job
+        mountPath: /opt/spark/app
```
andreyvelich: WDYT about these names @Goku2099 @tariq-hasan @Shekharrajak ?


```suggestion
    - name: spark-app-source
      emptyDir: {}

  initContainers:
    - name: prepare-spark-app
      volumeMounts:
        - name: spark-app-source
          mountPath: /opt/spark/app
      command:
        - bash
        - -c
        - |
          ...
          printf "%s" "$SCRIPT" > /opt/spark/app/main.py

  driver:
    volumeMounts:
      - name: spark-app-source
        mountPath: /opt/spark/app
```

proposals/107-spark-client/README.md
```diff
@@ -409,10 +600,16 @@ class SparkClient:
     def get_job_logs(
         self,
         name: str,
-        container: str = "driver",
+        step: str = "driver",
```
andreyvelich: Actually, I've been thinking about this a bit more. Maybe step isn't the right term here, since Spark has a well-established Driver/Executor model.
Let's remove it entirely and, for Phase 1, only expose logs from the Driver?
We can extend the API in later phases to support Executor logs as well.

WDYT @Goku2099 ?
```suggestion
```

andreyvelich (commented):
Thanks for the updates @Goku2099!
LGTM, just left a few small comments.

proposals/107-spark-client/README.md
Goku2099: I agree. Exposing only the driver logs in Phase 1 keeps the API simpler. We can extend get_job_logs() in a future phase with explicit executor log support once we define the appropriate API.

proposals/107-spark-client/README.md
Goku2099: Looks good to me.I think these names are clearer. I will update the KEP accordingly.

andreyvelich:
Thanks for this great work @Goku2099! Excited to see this moving forward.
/lgtm
/assign @Shekharrajak @tariq-hasan @RobuRishabh @vara-bonthu @nabuskey 
in case you have any followup comments.

andreyvelich:
Since we got +1 from other reviewers in Slack, we can move this forward.
/lgtm
/approve

PR #540, opened 2026-06-25
[WIP] Add kfp kubeflow client

### LINKED ISSUE #125: Integrate the Kubeflow Pipelines with Kubeflow SDK

MStokluska:

Implements the SDK side of KEP-125 — adds `kubeflow.pipelines` package that re-exports
`PipelinesClient` from `kfp.kubeflow_client`, along with DSL modules (`dsl`, `compiler`,
`components`, `kubernetes`), types, and constants.

This enables the full pipeline workflow from a single kubeflow import:

```python
from kubeflow.pipelines import PipelinesClient, dsl

client = PipelinesClient()
client.upload_pipeline(my_pipeline, name="training")
run = client.run("training", params={"epochs": 10})
client.wait_for_run_status(run, timeout=300)
```

The client implementation lives in the KFP repo (merged to pipelines master);
this PR adds the re-export layer, [pipelines] extra dependency, tests, and documentation.

Fixes https://github.com/kubeflow/sdk/issues/125

**Checklist:**


docs/source/conf.py
```diff
@@ -85,6 +85,50 @@
...
+#
+# Neither `autodoc_type_aliases` nor `typehints_formatter` can fix this reliably
+# because sphinx_autodoc_typehints evaluates type aliases to their underlying
+# runtime objects (where __name__ is V2beta1*), bypassing both mechanisms.
+#
+# The doctree-resolved hook below post-processes rendered text nodes to replace
+# V2beta1* names with their clean aliases. It only runs on pipelines/ pages to
+# avoid unintended side effects on other SDK docs.
+
+_KFP_TYPE_DISPLAY_NAMES = {
```
MStokluska, line 102: The `kfp.kubeflow_client` defines type aliases like `Pipeline = kfp_server_api.V2beta1Pipeline` using plain assignment. Sphinx resolves these to the underlying class names, so method signatures render as → `V2beta1PipelineVersion` instead of → `PipelineVersion`. These are internal server API type names that users should never see in docs IMO.

This doctree hook is a workaround — it post-processes the rendered text to replace `V2beta1*` names with the clean aliases, scoped only to pipelines pages.

The proper fix is in the kfp repo: change the plain aliases to TypeAlias annotations (`Pipeline: TypeAlias = V2beta1Pipeline`), which tells Sphinx to preserve the alias name. Once that lands, this hook can be removed

kubeflow/pipelines/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+            raise ImportError(_KFP_INSTALL_MSG) from e
+
+    if name in _KFP_ATTRS:
+        try:
+            from kfp import kubeflow_client
+
+            return getattr(kubeflow_client, name)
+        except ImportError as e:
+            raise ImportError(_KFP_INSTALL_MSG) from e
+
```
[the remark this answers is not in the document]
MStokluska: This mirrors the pattern used by other SDK clients. If kfp is installed but has an internal import failure, the user has a broken installation and the "please install" message is still correct IMO. Adding logic to distinguish "missing" vs "broken" would add complexity for an edge case that doesn't occur in practice.

pyproject.toml
```diff
@@ -44,6 +44,9 @@ spark = [
   "kubeflow-spark-api>=2.4.0",
 ]
 hub = ["model-registry>=0.3.6"]
+# TODO: Replace with a released kfp version (e.g. kfp[kubernetes]>=2.17.0) once
+# a kfp release including kfp.kubeflow_client is published to PyPI.
+pipelines = ["kfp[kubernetes]@git+https://github.com/kubeflow/pipelines.git@master#subdirectory=sdk/python"]
```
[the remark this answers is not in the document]
MStokluska: This is intentional and temporary - the TODO comment directly above this line explains it. The kfp release containing kfp.kubeflow_client hasn't been published to PyPI yet (the PR was just merged to master). Once a kfp release ships with this module, we'll replace the git reference with a pinned version

pyproject.toml
```diff
@@ -55,13 +58,14 @@ docs = [
...
 dev = [
-  "kubeflow[hub,spark]",
+  "kubeflow[hub,pipelines,spark]",
   "pytest>=7.0",
   "pytest-mock>=3.10",
   "coverage>=7.0",
   "kubeflow_trainer_api@git+https://github.com/kubeflow/trainer.git@master#subdirectory=api/python_api",
   "kubeflow_katib_api@git+https://github.com/kubeflow/katib.git@master#subdirectory=api/python_api",
   "kubeflow_spark_api@git+https://github.com/kubeflow/spark-operator.git@master#subdirectory=api/python_api",
+  "kfp[kubernetes]@git+https://github.com/kubeflow/pipelines.git@master#subdirectory=sdk/python",
```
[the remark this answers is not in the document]
MStokluska: All git-based dev dependencies in this file (kubeflow_trainer_api, kubeflow_katib_api, kubeflow_spark_api) point to @master, not pinned commits. This is the established pattern for dev - you always want the latest upstream code during development. Pinning to a specific commit would require constant manual bumps and diverge from the convention already in place.

docs/source/pipelines/index.rst
```diff
@@ -0,0 +1,107 @@
...
+
+How It Works
+------------
+
+1. **Define** - Author pipelines using the KFP DSL (``dsl``, ``compiler``, ``components``, ``kubernetes``)
+2. **Upload** - Register pipelines on the server with ``upload_pipeline``
+3. **Run** - Execute pipelines by name, callable, or YAML file
+4. **Monitor** - Wait for specific run states with ``wait_for_run_status``
+
+The ``PipelinesClient`` uses a name-first API: all methods work with display names
```
MStokluska: To improve, this sentence isn't 100% accurate.

kubeflow/pipelines/api/pipelines_client_test.py
```diff
@@ -0,0 +1,322 @@
...
+        for mod_name in list(sys.modules):
+            if mod_name == "kfp" or mod_name.startswith("kfp."):
+                del sys.modules[mod_name]
+        monkeypatch.setattr("builtins.__import__", _block_kfp_import)
+
+        mod = importlib.import_module("kubeflow.pipelines")
+
+        try:
+            getattr(mod, test_case.config["attr"])
+            assert test_case.expected_status == SUCCESS
```
andreyvelich: This try/except idiom is inverted and fragile, and diverges from the house style in `backend_test.py` which uses `pytest.raises`. Suggest:
```python
with pytest.raises(ImportError, match=r"pip install 'kubeflow\[pipelines\]'"):
    getattr(mod, test_case.config["attr"])
```

kubeflow/pipelines/api/pipelines_client_test.py
```diff
@@ -0,0 +1,322 @@
...
+                expected_error=AttributeError,
+            ),
+        ],
+    )
+    def test_unknown_attr(self, test_case):
+        import kubeflow.pipelines as kp
+
+        try:
+            getattr(kp, test_case.config["attr"])
+            assert test_case.expected_status == SUCCESS
```
andreyvelich: Prefer `with pytest.raises(AttributeError, match="has no attribute"):`. The current form also doesn't assert the message, so a regression to a generic error wouldn't be caught.

kubeflow/pipelines/api/pipelines_client_test.py
```diff
@@ -0,0 +1,322 @@
...
+    def test_type_reexport(self, test_case):
+        from kfp import kubeflow_client as kfp_kc
+
+        import kubeflow.pipelines as kp
+
+        type_name = test_case.config["type_name"]
+        reexported = getattr(kp, type_name)
+        original = getattr(kfp_kc, type_name)
+        assert reexported is original
+        assert test_case.expected_status == SUCCESS
```
andreyvelich: `assert test_case.expected_status == SUCCESS` compares a hardcoded constant to itself — it's a no-op. The real check here is `reexported is original`. Consider dropping the `TestCase` wrapper for a plain `@pytest.mark.parametrize("type_name", [...])`, or use `TestCase` as intended with varying success/error cases.

README.md
```diff
@@ -214,6 +214,36 @@ ModelRegistryClient("https://example.org:456")        # Port parsed from base_ur
...
+
+client = PipelinesClient()
+
+# Upload and run
+client.upload_pipeline(hello_pipeline, name="hello-pipeline")
+run = client.run("hello-pipeline", params={"greeting": "Kubeflow"})
+
+# Wait for completion
+completed = client.wait_for_run_status(run, timeout=300)
+print(f"Run finished: {completed.state}")
```
andreyvelich, line 244: `Run.state` returns the raw server enum in **UPPERCASE** (`'SUCCEEDED'`), but the documented constants are lowercase (`RUN_SUCCEEDED == 'succeeded'`). A reader who writes `completed.state == RUN_SUCCEEDED` will silently get `False`. Worth a note clarifying the casing.

docs/source/pipelines/index.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   client = PipelinesClient()
+
+   # Upload and run
+   client.upload_pipeline(hello_pipeline, name="hello-pipeline")
+   run = client.run("hello-pipeline", params={"greeting": "Kubeflow"})
+
+   # Wait for completion
+   completed = client.wait_for_run_status(run, timeout=300)
+   print(f"Run finished: {completed.state}")
```
andreyvelich, line 52: Same `.state` casing caveat as the README: `Run.state` is uppercase (`'SUCCEEDED'`) while `constants.RUN_SUCCEEDED` is lowercase. Consider clarifying here too.

docs/source/pipelines/index.rst
```diff
@@ -0,0 +1,107 @@
...
+Overview
+--------
+
+The Kubeflow SDK provides a ``PipelinesClient`` that gives users the full
+author, compile, upload, run, and monitor pipeline workflow from a single
+``kubeflow`` import. Using the Pipelines SDK, you can:
+
+- **Define pipelines** - Use the KFP DSL to author pipelines and components
+- **Upload and version** - Register pipelines on the server with automatic versioning
+- **Run pipelines** - Execute by name, from a function, or from a compiled YAML file
```
andreyvelich, line 15: YAML-file run mode is listed here (and at line 59) but never shown — all examples use run-by-name/callable. Add a one-line `client.run("pipeline.yaml", ...)` example or drop the claim.

docs/source/conf.py
```diff
@@ -85,6 +85,50 @@
...
+}
+
+
+def _replace_v2beta1_in_doctree(_app, doctree, docname):
+    if not docname.startswith("pipelines/"):
+        return
+
+    from docutils import nodes
+
+    for node in doctree.traverse(nodes.Text):
```
andreyvelich: `doctree.traverse()` is deprecated since docutils 0.18 (the docs group resolves 0.22.4). Use `doctree.findall(nodes.Text)` to avoid a `DeprecationWarning`.

docs/source/pipelines/api.rst
```diff
@@ -0,0 +1,82 @@
...
+- ``ListPipelineVersionsResponse`` — paginated list of pipeline versions
+- ``ListRunsResponse`` — paginated list of runs
+- ``ListExperimentsResponse`` — paginated list of experiments
+
+Constants
+---------
+
+Available via ``from kubeflow.pipelines import constants``.
+
+.. data:: kubeflow.pipelines.constants.RUN_SUCCEEDED
```
andreyvelich, line 44: These constant values are hardcoded in RST and can silently drift from `kfp.kubeflow_client.constants`. Prefer `.. autodata:: kubeflow.pipelines.constants.RUN_SUCCEEDED` so they track the source.

README.md
```diff
@@ -245,7 +275,7 @@ job_id = client.train(trainer=CustomTrainer(func=train_fn))
 | **Kubeflow Katib**          | ✅ **Available** | v0.19.0+        | Hyperparameter optimization                                           |
 | **Kubeflow Model Registry** | ✅ **Available** | v0.3.0+         | Manage model artifacts, versions and ML artifacts metadata            |
 | **Kubeflow Spark Operator** | ✅ **Available** | v2.5.0+         | Manage Spark applications for data processing and feature engineering |
-| **Kubeflow Pipelines**      | 🚧 Planned       | TBD             | Build, run, and track AI workflows                                    |
+| **Kubeflow Pipelines**      | ✅ **Available** | latest          | Build, run, and track AI workflows                                    |
```
andreyvelich: Other rows pin a concrete minimum (e.g. `v2.5.0+`); `latest` is vague. Consider the eventual pinned version, or note `git/master (pre-release)` until the kfp release lands.

kubeflow/pipelines/__init__.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            from kubeflow.pipelines.api.pipelines_client import PipelinesClient
+
+            return PipelinesClient
+        except ImportError as e:
+            raise ImportError(_KFP_INSTALL_MSG) from e
+
+    if name in _DSL_MODULES:
+        try:
+            import importlib
```
andreyvelich: Nit: `import importlib` runs on every DSL-module access and has no kfp dependency — moving it to module top-level would be marginally cleaner.

andreyvelich (commented):
Thanks for the work on this, @MStokluska! 🙌

This review was performed with AI-assisted tooling. I verified all re-exported names, types, constants, and the documented method signatures against the upstream `kfp.kubeflow_client` (kubeflow/pipelines#13405) — everything matches. I've left a few small, actionable inline comments below.

README.md
andreyvelich: @MStokluska Could you also update diagram with the up-to-date KFP logo: https://github.com/kubeflow/sdk#kubeflow-sdk-benefits

README.md
MStokluska: Good catch. The README example only prints completed.state (no comparison), so it works as-is. I've added a casing note to the Sphinx docs under Key Concepts instead — that's where users go for details beyond the quick-start.

docs/source/pipelines/index.rst
MStokluska: Good catch — added a YAML file example under Common Patterns.

docs/source/pipelines/api.rst
MStokluska: Tried autodata with :no-value:, :annotation:, and both combined — all render the str class docstring (str(object='') -> str ...) because these are plain string values, not typed attributes. Sphinx autodoc always pulls the class docstring for the object's type. The manual .. data:: directives seems to be the cleanest option here.

HarshPopat23:
Hii @andreyvelich and @MStokluska ,

One small improvement suggestion for the `PipelinesClient` integration: since the wrapper exposes constants like `RUN_SUCCEEDED` in lowercase, but KFP run objects may return raw server states like `SUCCEEDED` in uppercase, could we normalize statuses inside `wait_for_run_status`?

For example, `wait_for_run_status` could accept both SDK constants (`succeeded`, `failed`) and raw KFP states (`SUCCEEDED`, `FAILED`) by normalizing both sides before comparison. This would avoid users accidentally writing checks that silently fail due to casing differences.

This seems aligned with the goal of making the Kubeflow SDK wrapper more user-friendly while still keeping it thin over `kfp.Client`.

andreyvelich:
This was merged into this PR: https://github.com/kubeflow/sdk/pull/591

/close

PR #547, opened 2026-06-30
fix: don't silently fall back to current context namespace

anxkhn:

`get_default_target_namespace()` in `kubeflow/common/utils.py` is used by both the
Trainer and Optimizer Kubernetes backends to choose the namespace when a caller
passes none. When the caller does pass an explicit `context=`, the function did
not honor it: if that context was not found, or existed but had no `namespace`
set, it silently returned the *current* kube context's namespace instead. The
missing-namespace case raised `KeyError`, which the broad `except` swallowed,
also landing on the current/default context. Either way the user's explicit
`context=` intent was ignored, so a job could be created in the wrong namespace.

This resolves the requested context's namespace safely and falls back to
`constants.DEFAULT_NAMESPACE` when that context is absent or has no namespace,
instead of using the current context. The no-context path is unchanged. It also
strips the trailing newline from the in-cluster service-account namespace read,
matching the canonical Kubernetes pattern. Adds the first unit tests for
`kubeflow/common/utils.py` (mocks only, no network).

Fixes #

(No tracking issue - self-identified correctness fix. Happy to file one first if
preferred.)

**Checklist:**

(Not user-facing: internal default-namespace resolution only; no API or docs
change.)

### DIFF
--- a/kubeflow/common/utils.py
+++ b/kubeflow/common/utils.py
@@ -26,17 +26,24 @@ def get_default_target_namespace(context: str | None = None) -> str:
     if not is_running_in_k8s():
         try:
             all_contexts, current_context = config.list_kube_config_contexts()
-            # If context is set, we should get namespace from it.
-            if context:
+            # If a context is explicitly requested, honor only that context. Fall
+            # back to the default namespace rather than silently using the current
+            # context's namespace when the requested context is missing or has no
+            # namespace set. Use "is not None" so an explicit empty string is still
+            # treated as an explicit request instead of falling through to the
+            # current context.
+            if context is not None:
                 for c in all_contexts:
                     if isinstance(c, dict) and c.get("name") == context:
-                        return c["context"]["namespace"]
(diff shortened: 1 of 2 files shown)


anxkhn:
Good catch. Switched the guard from `if context:` to `if context is not None:`, so an explicitly provided `context=""` is now treated as an explicit request rather than falling through to the current context's namespace. An empty string matches no kubeconfig context, so it resolves to `DEFAULT_NAMESPACE`, the same non-silent behavior as an unknown context (kept consistent with the function's existing fall-back-to-default handling rather than adding a new raise path, which the broad `except` would swallow anyway). Added a unit test covering `context=""` alongside the existing `None` and explicit-context cases.

anxkhn:
@kramaranya rebased onto current main and resolved the utils test conflict while preserving the upstream validation tests. Relevant tests and lint pass. Please take another look when you have a chance.

PR #549, opened 2026-07-01
feat: add runtime kind support for TrainingRuntime and ClusterTrainingRuntime

Sumitha007:

This PR adds runtime kind support to the Kubernetes backend by preserving whether a runtime is a `TrainingRuntime` or `ClusterTrainingRuntime`.

Fixes: https://github.com/kubeflow/sdk/pull/549

### Changes
- Add a `kind` field to the `Runtime` type.
- Populate the runtime kind when retrieving runtimes from Kubernetes.
- Update runtime-related unit tests.
- Update TrainJob tests to validate the runtime kind.


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -593,8 +593,10 @@ def __get_runtime_from_cr(
             )
 
         return types.Runtime(
```
andreyvelich, line 604: Indentation is broken here (args under-indented, plus a stray blank line inside `get_runtime_trainer(`). `ruff format --check` fails on this, which breaks `make verify`/CI. Run `uv run ruff format kubeflow` before pushing.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -593,8 +593,10 @@ def __get_runtime_from_cr(
...
-            name=runtime_cr.metadata.name,
-            trainer=utils.get_runtime_trainer(
+        name=runtime_cr.metadata.name,
+        kind=runtime_cr.kind,
```
andreyvelich: `runtime_cr.kind` is likely `None` in the `list_runtimes` path: Kubernetes list responses don't populate `kind` on individual items (only the `...List` envelope does). That would make `runtimeRef.kind` `None` and defeat this PR's goal. The tests don't catch it — `normalize_model` sets `kind` on each mock item, which real list items lack. Please verify against a real cluster; if confirmed, derive kind from the list scope (namespaced vs cluster) instead.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -916,18 +926,22 @@ def test_get_runtime(kubernetes_backend, test_case):
...
-            config={
-                "namespace_error": client.ApiException(status=404),
-                "name": LIST_RUNTIMES,
-            },
-            expected_output=[
-                create_runtime_type(name="runtime-1"),
-                create_runtime_type(name="runtime-2"),
-                create_runtime_type(name="runtime-3"),
-            ],
+    name="namespace 404 but cluster succeeds",
```
andreyvelich: This `TestCase(...)` block is dedented to the wrong level — `ruff format --check` rejects it (CI failure). `uv run ruff format` fixes it.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -690,11 +691,19 @@ def create_runtime_type(
...
+            if name in {TORCH_RUNTIME, "runtime-1", "ns-runtime-2"}
+            else "ClusterTrainingRuntime"
+        )
 
+    return types.Runtime(
+        name=name,
+        trainer=trainer,
+        kind=kind,
+    )
 def get_train_job_data_type(
```
andreyvelich, line 715: Missing two blank lines before this function (`create_runtime_type` runs straight into it). `ruff format` will add them.

kubeflow/trainer/types/types.py
```diff
@@ -269,6 +269,7 @@ def set_command(self, command: tuple[str, ...]):
 class Runtime:
     name: str
     trainer: RuntimeTrainer
+    kind: str | None = None
```
andreyvelich: Consider `Literal["TrainingRuntime", "ClusterTrainingRuntime"] | None` (or an enum / the existing `constants.*_KIND`) instead of a bare `str` so only the two valid kinds are representable.

andreyvelich (commented):
Thanks for the contribution, @Sumitha007! 🙌 This review was performed with the help of AI tooling — comments are inline on specific lines.

Sumitha007:
Thanks for the review! I'm working through the suggested changes and will push an update shortly.

kubeflow/trainer/types/types.py
Sumitha007: Thanks! Updated Runtime.kind to use Literal["TrainingRuntime", "ClusterTrainingRuntime"] | None and pushed the changes.

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -569,6 +584,7 @@ def __get_runtime_from_cr(
         self,
         runtime_cr: models.TrainerV1alpha1ClusterTrainingRuntime
         | models.TrainerV1alpha1TrainingRuntime,
+        kind: str,
```
andreyvelich: Instead of passing kind to this function, you can directly access it via: `runtime_cr.kind`

```suggestion
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -569,6 +584,7 @@ def __get_runtime_from_cr(
         self,
```
andreyvelich, line 578: Can you also add another unit tests to the `train()` API to ensure we create TrainJob with the appropriate spec, like here: https://github.com/Sumitha007/sdk/blob/b5114f20a1b1629dc6d1722939acc343c85961b8/kubeflow/trainer/backends/kubernetes/backend_test.py#L1344C13-L1344C28

kubeflow/trainer/backends/kubernetes/backend.py
Sumitha007: Thanks for the suggestion! I have one question before I update this. Earlier you mentioned that runtime_cr.kind could be None in the list_runtimes() path. Should I also revert the changes in list_runtimes() to use runtime_cr.kind, or should I keep deriving the kind from the list scope there?

andreyvelich: Good point.
I think, k8s list API should populate the Kind parameter in the object (e.g. ClusterTrainingRuntime).
Could you verify locally if you just use `runtime_cr.kind`, will that work correctly?

Sumitha007: Thanks! I tried to verify this locally, but I don't have a running Kubernetes/Kubeflow cluster available. kubectl cannot connect to a cluster, so I was only able to run the unit tests. Is there a recommended way to reproduce this locally, or another way you'd suggest verifying whether runtime_cr.kind is populated in list responses?

andreyvelich: Yeah, could you try to just use `Kind` to deploy cluster locally and deploy Trainer control plane?
You can follow this getting started guide: https://trainer.kubeflow.org/en/latest/operator-guides/installation.html#prerequisites

Sumitha007: Thanks! I'll set up a local Kind cluster following the installation guide and verify whether runtime_cr.kind is populated in the list API. I'll update the PR with my findings.

Sumitha007: Thanks! I verified this on a local Kind cluster. ClusterTrainingRuntime objects returned by the Kubernetes API include the kind field (for example, kind: ClusterTrainingRuntime in the YAML output). I've updated the implementation to use runtime_cr.kind directly and pushed the changes.

kubeflow/trainer/backends/kubernetes/backend.py
Sumitha007: I can definitely add the unit test. Just to confirm, do you want a train() test that verifies TrainJob.spec.runtimeRef.kind is set correctly for both TrainingRuntime and ClusterTrainingRuntime?

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -332,7 +332,7 @@ def get_train_job(
             annotations=annotations,
         ),
         spec=models.TrainerV1alpha1TrainJobSpec(
-            runtimeRef=models.TrainerV1alpha1RuntimeRef(name=runtime_name),
+            runtimeRef=models.TrainerV1alpha1RuntimeRef(name=runtime_name, kind="TrainingRuntime"),
```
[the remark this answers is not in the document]
andreyvelich: @Sumitha007 Can you you constant here?
`TRAINING_RUNTIME_KIND`

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -578,7 +578,7 @@ def create_train_job(
             creationTimestamp=datetime.datetime(2025, 6, 1, 10, 30, 0),
         ),
         spec=models.TrainerV1alpha1TrainJobSpec(
-            runtimeRef=models.TrainerV1alpha1RuntimeRef(name=TORCH_RUNTIME),
+            runtimeRef=models.TrainerV1alpha1RuntimeRef(name=TORCH_RUNTIME, kind="TrainingRuntime"),
```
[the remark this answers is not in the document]
andreyvelich: Same comment.

kubeflow/trainer/types/types.py
```diff
@@ -269,6 +270,7 @@ def set_command(self, command: tuple[str, ...]):
 class Runtime:
     name: str
     trainer: RuntimeTrainer
+    kind: Literal["TrainingRuntime", "ClusterTrainingRuntime"] | None = None
```
andreyvelich: It cannot be None, right?

```suggestion
    kind: Literal["TrainingRuntime", "ClusterTrainingRuntime"] 
```

andreyvelich: Maybe we should use enum here? Like we do for other types: https://github.com/Sumitha007/sdk/blob/73251d52d73193fd2e32ed8783093486c48d1dfa/kubeflow/trainer/types/types.py#L103-L107

Thus, you don't need to define constants for them: https://github.com/Sumitha007/sdk/blob/73251d52d73193fd2e32ed8783093486c48d1dfa/kubeflow/trainer/constants/constants.py#L23-L24

andreyvelich (commented):
/ok-to-test

kubeflow/trainer/types/types.py
Sumitha007: Thanks for the review! I've addressed the feedback by replacing the hardcoded runtime kind strings in the tests with the corresponding constants and updating Runtime.kind to be non-optional since it is always populated. I've pushed the latest changes. Please let me know if you'd still prefer using an enum for Runtime.kind; I'm happy to make that change as well.

andreyvelich: Thanks @Sumitha007! Yes, let's switch to enums for now. After you switch to enums, you can delete `ClusterTrainingRuntime` from constants.

Sumitha007: Thanks for the review! I've switched the runtime kind constants to a RuntimeKind enum, removed the old constants, updated all usages, and verified the relevant tests pass. Please let me know if there's any further feedback.

kubeflow/trainer/constants/constants.py
```diff
@@ -20,15 +21,16 @@
 VERSION = "v1alpha1"
 API_VERSION = f"{GROUP}/{VERSION}"
 
-# The Kind name for the ClusterTrainingRuntime.
-CLUSTER_TRAINING_RUNTIME_KIND = "ClusterTrainingRuntime"
+
+# The Kind name for the ClusterTrainingRuntime, TrainingRuntime.
+class RuntimeKind(str, Enum):
+    CLUSTER_TRAINING_RUNTIME = "ClusterTrainingRuntime"
+    TRAINING_RUNTIME = "TrainingRuntime"
```
andreyvelich: Please can you move it to `types.py`: https://github.com/Sumitha007/sdk/blob/2bfee9c3ce9bb4f061fb57cb1e79b8059ae382ef/kubeflow/trainer/types/types.py#L100
```suggestion
# The Kind name for the ClusterTrainingRuntime, TrainingRuntime.
class RuntimeKind(Enum):
    CLUSTER_TRAINING_RUNTIME = "ClusterTrainingRuntime"
    TRAINING_RUNTIME = "TrainingRuntime"
```

kubeflow/trainer/types/types.py
```diff
@@ -269,6 +270,7 @@ def set_command(self, command: tuple[str, ...]):
 class Runtime:
     name: str
     trainer: RuntimeTrainer
+    kind: Literal["TrainingRuntime", "ClusterTrainingRuntime"]
```
andreyvelich: We should switch it to enums as we discussed here, right @Sumitha007?

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: @Sumitha007 Yes, let's add a unit test for `train()` API to ensure the correct Runtime is selected.

Sumitha007:
Hi, I've addressed all the requested review comments, updated the tests, and verified the changes locally. Could you please review the latest commit? Thanks!

kubeflow/trainer/backends/kubernetes/backend.py
andreyvelich: @Sumitha007 Did you get a chance to address this comment?

andreyvelich (commented):
/ok-to-test
/retest

Sumitha007:
Hi @andreyvelich, I'm working on the additional train() API unit test. While implementing it, I noticed that in the current test setup the namespaced TrainingRuntime is always selected when both namespaced and cluster-scoped runtimes are available. Should the new test verify that train() prefers the namespaced TrainingRuntime, or is there another runtime selection scenario you'd like me to cover? Thanks!

Sumitha007:
Hi @andreyevlich, gentle follow-up on my previous question. I want to make sure I implement the expected behavior correctly before proceeding. Could you please clarify the intended runtime selection behavior? Thanks!

andreyvelich:
>While implementing it, I noticed that in the current test setup the namespaced TrainingRuntime is always selected when both namespaced and cluster-scoped runtimes are available

Can we extent the unit test to add test case with TrainingRuntime and ClusterTrainingRuntime?

Sumitha007:
Thanks! I'll extend the unit test to include a scenario where both TrainingRuntime and ClusterTrainingRuntime are available and verify the runtime selection behavior.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1031,6 +1062,7 @@ def test_list_runtimes(kubernetes_backend, test_case):
                         device_count="1",
                         image="example.com/image",
                     ),
+                    kind=types.RuntimeKind.TRAINING_RUNTIME,
```
andreyvelich, line 1065: @Sumitha007 Did you also add the test case to `train()` API to ensure the namespace-scoped TrainingRuntime is selected?
https://github.com/Sumitha007/sdk/blob/2e262a80431518d338f633d9d2172d94316aad8d/kubeflow/trainer/backends/kubernetes/backend_test.py#L1388

Sumitha007:
Thanks! I've added a train() API test covering the namespace-scoped TrainingRuntime selection and updated the PR.

andreyvelich (commented):
Thanks @Sumitha007!
/lgtm
/approve

PR #554, opened 2026-07-03
fix(trainer): validate polling_interval in wait_for_job_status

### LINKED ISSUE #550: bug(trainer): wait_for_job_status() accepts zero or negative polling_interval causing CPU busy-loop or cryptic error
### What happened?

`TrainerClient.wait_for_job_status()` accepts a `polling_interval` parameter but performs **no validation** on its value. When a user passes `0` or a negative number, one of two silent failure modes occurs:   

- `polling_interval=0` → `time.sleep(0)` is called in the `while True:` polling loop, causing a **CPU busy-loop** that hammers the Kubernetes API thousands of times per second with no visible error.

- `polling_interval=-5` → Python raises a **cryptic stdlib error** (`ValueError: sleep length must be non-negative`) deep inside the call stack with no hint that the caller passed a bad value.

Neither case gives the user a helpful, actionable error message.

**Method signature (from `kubeflow/trainer/api/trainer_client.py`):**

```python

def wait_for_job_status(

    self,

    name: str,

    status: set[str] = {"Complete"},

    timeout: int = 600,

    polling_interval: int = 2,   # ← no validation

    callbacks: list[Callable[[TrainJob], None]] | None = None,

) -> TrainJob:

```

**Steps to reproduce:**

```python

client = TrainerClient()

# Triggers CPU busy-loop:

client.wait_for_job_status("my-job", polling_interval=0)



# Triggers cryptic stdlib ValueError:

client.wait_for_job_status("my-job", polling_interval=-5)

```

**Note:** The same gap exists in `ContainerBackend.wait_for_job_status()` (tracked in #411). This issue covers the **Kubernetes backend** path in `TrainerClient`.

A clear, descriptive `ValueError` should be raised immediately when `polling_interval <= 0`, before any polling loop begins. Example: 

ValueError: polling_interval must be a positive number, got: 0

This is consistent with how the SDK already documents `ValueError` in the `wait_for_job_status()` raises section ("The input values are incorrect") — the validation just needs to be implemented.  

### Environment

- **Kubeflow SDK version:** v0.4.0 (latest)
- **Affected file:** `kubeflow/trainer/api/trainer_client.py`
- **Affected method:** `TrainerClient.wait_for_job_status()`
- **Related issue:** #411 (same bug in `ContainerBackend`)
- **Python:** 3.9+
- **Backend:** KubernetesBackend (default)

## Proposed Fix  Add input validation at the top of `wait_for_job_status()` in `kubeflow/trainer/api/trainer_client.py`:  ```python if not isinstance(polling_interval, (int, float)) or polling_interval <= 0:     raise ValueError(         f"polling_interval must be a positive number, got: {polling_interval}"     ) ```  **Files to change:** 1. `kubeflow/trainer/api/trainer_client.py` — add the guard clause above 2. `test/unit/trainer/test_trainer_client.py` — add unit tests for `polling_interval=0` and `polling_interval=-5`  **Test cases to add:** ```python def test_wait_for_job_status_invalid_polling_interval():     client = TrainerClient()     with pytest.raises(ValueError, match="polling_interval"):         client.wait_for_job_status("some-job", polling_interval=0)     with pytest.raises(ValueError, match="polling_interval"):         client.wait_for_job_status("some-job", polling_interval=-5) ```  ---  I would like to work on this issue. Could a maintainer please assign it to me? 🙏  **My plan:** 1. Add the `ValueError` guard in `trainer_client.py` 2. Add unit tests covering both failure modes 3. Update the docstring `Raises` section to explicitly mention `polling_interval` 4. Open a PR referencing this issue and #411 for consistency

HarshPopat23:

Adds input validation to `TrainerClient.wait_for_job_status()` to reject

non-positive `polling_interval` values before delegating to the backend.

Previously, passing `polling_interval=0` could cause a CPU busy-loop

(`time.sleep(0)` in a tight polling loop), and negative values raised a

cryptic stdlib `ValueError` from deep inside `time.sleep()` with no

context about which argument was invalid.

Fixes #550

**Checklist:**

- [x] Docs included if any changes are user facing (docstring already documented `ValueError: The input values are incorrect.` — no change needed)


kubeflow/trainer/api/trainer_client_test.py
```diff
@@ -70,3 +70,18 @@ def test_backend_selection(test_case):
         client = TrainerClient(backend_config=test_case["backend_config"])
         backend_name = client.backend.__class__.__name__
         assert backend_name == test_case["expected_backend"]
+
+
+def test_wait_for_job_status_invalid_polling_interval():
```
andreyvelich: Could you change name for that.

```suggestion
def test_wait_for_job_status():
```

kubeflow/trainer/api/trainer_client_test.py
```diff
@@ -70,3 +70,18 @@ def test_backend_selection(test_case):
...
+        patch("kubernetes.config.load_kube_config"),
+        patch("kubernetes.client.CustomObjectsApi") as mock_custom_api,
+        patch("kubernetes.client.CoreV1Api") as mock_core_api,
+    ):
+        mock_custom_api.return_value = Mock()
+        mock_core_api.return_value = Mock()
+
+        client = TrainerClient()
+
+
```
andreyvelich: Fix the formatting.

kubeflow/trainer/api/trainer_client.py
```diff
@@ -259,6 +259,16 @@ def wait_for_job_status(
...
         """
+        if polling_interval <= 0:
+            raise ValueError(
+                f"Polling interval must be a positive number, got polling_interval={polling_interval}"
+            )
+        if polling_interval >= timeout:
+            raise ValueError(
+                "Polling interval must be strictly less than timeout. "
+                f"Received polling_interval={polling_interval}, timeout={timeout}"
+            )
```
andreyvelich: Could you also add  this validation for OptimizerClient()?
We might need similar for SparkClient after this PR: https://github.com/kubeflow/sdk/pull/521
And PipelinesClient(): https://github.com/kubeflow/sdk/pull/540

Maybe it makes sense to move it to common utils: https://github.com/kubeflow/sdk/blob/main/kubeflow/common/utils.py

Additionally, you can check if timeout is > 0.

cc @Goku2099 @MStokluska 
```suggestion
        utils.validate_wait_intervals(polling_interval, timeout)
```

andreyvelich (commented):
/ok-to-test
/retest

kubeflow/trainer/api/trainer_client_test.py
HarshPopat23: I hade changed name **test_wait_for_job_status_invalid_polling_interval():** -> to **def test_wait_for_job_status():**

kubeflow/trainer/api/trainer_client.py
HarshPopat23: I have added the unit test for  **OptimizerClient()**.

kubeflow/trainer/api/trainer_client_test.py
HarshPopat23: @andreyvelich I'm a new contributor, and this is my first issue, so I may have made some mistakes. I'm trying my best to contribute to this organization.

I fixed the other two issues, but I overlooked the formatting in this commit. I'll make sure to follow the correct formatting in my next commit. Sorry about that, and thank you for the feedback!

HarshPopat23: i have Fixed this formatting.

Goku2099:
Small naming suggestion: since this helper validates both polling_interval and timeout specifically for wait_for_job_status, would something like validate_wait_for_job_status_args() (or ..._params()) make its purpose a bit clearer?
 Not a blocker, just a readability suggestion.

HarshPopat23:
> Small naming suggestion: since this helper validates both polling_interval and timeout specifically for wait_for_job_status, would something like validate_wait_for_job_status_args() (or ..._params()) make its purpose a bit clearer? Not a blocker, just a readability suggestion.

Hi @Goku2099, thank you for taking the time to review my PR and for the suggestion! I actually used this function name based on a recommendation from @andreyvelich. If they agree that **validate_wait_for_job_status_args() (or a similar name)** is more appropriate, I'd be happy to update it. Thanks again for the helpful feedback!

kubeflow/trainer/api/trainer_client_test.py
Goku2099: @HarshPopat23 Since this is your first contribution, you may also find the  CONTRIBUTING guide helpful.
https://github.com/kubeflow/sdk/blob/main/CONTRIBUTING.md

andreyvelich:
@HarshPopat23 @Goku2099 Maybe we can just name it as: `validate_wait_for_job_status()`?

Goku2099:
> @HarshPopat23 @Goku2099 Maybe we can just name it as: `validate_wait_for_job_status()`?

Fine by my side.

HarshPopat23:
> @HarshPopat23 @Goku2099 Maybe we can just name it as: `validate_wait_for_job_status()`?

Hi @andreyvelich, sounds good — I've renamed it to `validate_wait_for_job_status()` 
as suggested and pushed the update.

Also, thanks @Goku2099 for the suggestion and for taking interest in this PR!

Looking forward to your review whenever you get a chance.

kubeflow/common/utils_test.py
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+
+from kubeflow.common import utils
+
+
+def test_validate_wait_for_job_status():
```
andreyvelich: Please can you also refactor this test by using TestCases as described here: https://github.com/kubeflow/sdk/blob/main/AGENTS.md#3-testing-requirements

kubeflow/trainer/api/trainer_client_test.py
```diff
@@ -70,3 +70,25 @@ def test_backend_selection(test_case):
         client = TrainerClient(backend_config=test_case["backend_config"])
         backend_name = client.backend.__class__.__name__
         assert backend_name == test_case["expected_backend"]
+
+
+def test_wait_for_job_status():
+    """Test TrainerClient.wait_for_job_status validates polling_interval and timeout."""
```
andreyvelich: You can remove this, since you test it in the utils.


```suggestion
```

HarshPopat23: Good catch, removed. Thanks!

kubeflow/common/utils_test.py
HarshPopat23: Hi @andreyvelich, done! Refactored to use the TestCase dataclass pattern 
as described in AGENTS.md. Just pushed the update. Thanks for pointing 
me to the reference doc.

andreyvelich: @HarshPopat23 Did you push the commit? I only see that you deleted the other test function as part of: https://github.com/kubeflow/sdk/pull/554/changes/5e0d4e01a4113e829d1b7d7c25b58ed60a60fe86

HarshPopat23: Hi @andreyvelich, apologies — you were right, my earlier commit only removed 
the redundant test from trainer_client_test.py but didn't actually apply the 
TestCase refactor to utils_test.py. Just pushed the correct fix now in the 
latest commit. Thanks for catching that!

andreyvelich (commented):
Thanks @HarshPopat23!
/lgtm
/approve

PR #555, opened 2026-07-03
fix(trainer): stream localprocess logs to the caller

### LINKED ISSUE #435: bug(trainer): LocalJob.logs(follow=True) prints directly to stdout instead of yielding to caller
### What happened?

Two bugs in `LocalJob.logs(follow=True)` that make it pretty unusable as a streaming interface:

**1. Duplicate output.** The method calls `print()` internally, so if you iterate `get_job_logs(follow=True)` and print each line yourself, everything shows up twice — once from inside the SDK, once from your own code.

**2. No real-time streaming.** Despite `follow=True` implying otherwise, nothing is yielded until the job finishes. It's just a batched return wearing a streaming API's clothes.

### Steps to Reproduce

```python
from kubeflow.trainer.backends.localprocess.backend import LocalProcessBackendConfig

client = TrainerClient(backend_config=LocalProcessBackendConfig())
job_id = client.train(...)

for line in client.get_job_logs(name=job_id, follow=True):
    print(line)  # every line prints twice
```

### Where it's happening

`kubeflow/trainer/backends/localprocess/job.py` — `logs()` method:

```python
def logs(self, follow=False) -> list[str]:
    if not follow:
        return self._stdout.splitlines()
    try:
        for chunk in self.stream_logs():
            print(chunk, end="", flush=True)  # ← writes directly to stdout, caller has no say
    except StopIteration:
        pass
    return self._stdout.splitlines()  # ← blocks until job is done, then dumps everything at once
```

### What should happen instead

- No `print()` inside the method — yielding lines and leaving output decisions to the caller is the whole point of this API.
- Lines should reach the caller as they're produced, not in one batch after the job exits.

### Environment

- Kubeflow SDK: `0.4.0`
- Backend: `LocalProcessBackend`
- No Kubernetes cluster needed to reproduce

/kind bug
/area local

immanuwell:
What this PR does / why we need it

`LocalJob.logs(follow=True)` was a bit busted. It printed inside the SDK and only yielded after the job finished, so callers got duplicate output and no real streaming.

This returns `stream_logs()` to the caller, keeps `follow=False` as-is, and adds regression tests.

Repro:
```python
from kubeflow.trainer.backends.localprocess.job import LocalJob

job = LocalJob(
    name="demo",
    command=["python3", "-c", "import time; print("line1", flush=True); time.sleep(0.2); print("line2", flush=True)"],
)
job.start()
for line in job.logs(follow=True):
    print(line, end="")
```

Before this fix, output was printed from inside the SDK and nothing was yielded until the process exited. After this fix, the caller gets log chunks as they show up.

Which issue(s) this PR fixes

Fixes #435

Checklist

### DIFF
--- a/kubeflow/trainer/backends/localprocess/job.py
+++ b/kubeflow/trainer/backends/localprocess/job.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from collections.abc import Iterator
 from datetime import datetime
 import logging
 import os
@@ -149,30 +150,32 @@ class LocalJob(threading.Thread):
     def returncode(self):
         return self._returncode
 
-    def logs(self, follow=False) -> list[str]:
+    def logs(self, follow: bool = False) -> list[str] | Iterator[str]:
+        """Get stdout for the local job.
+
+        Args:
+            follow: Whether to stream logs in realtime as they are produced.
+
+        Returns:
+            Buffered stdout lines when ``follow`` is disabled. Otherwise, an
+            iterator that yields stdout lines as they become available.
+        """
         if not follow:
             return self._stdout.splitlines()
 
-        try:
-            for chunk in self.stream_logs():
-                print(chunk, end="", flush=True)  # stream to console live
-        except StopIteration:
-            pass
+        return self.stream_logs()
 
-        return self._stdout.splitlines()
-
-    def stream_logs(self):
-        """Generator that yields new output lines as they come in."""
-        last_index = 0
-        while self.is_alive() or last_index < len(self._stdout):
+    def stream_logs(self) -> Iterator[str]:
+        """Yield newly available stdout lines as they are written by the process."""
+        emitted_line_count = 0
+        while self.is_alive() or emitted_line_count < len(self._stdout.splitlines()):
             self._output_updated.wait(timeout=1)
             with self._lock:
-                data = self._stdout
-                new_data = data[last_index:]
-                last_index = len(data)
+                lines = self._stdout.splitlines()
+                new_lines = lines[emitted_line_count:]
+                emitted_line_count = len(lines)
                 self._output_updated.clear()
-            if new_data:
-                yield new_data
+            yield from new_lines
 
     @property
     def creation_time(self):
--- a/kubeflow/trainer/backends/localprocess/job_test.py
+++ b/kubeflow/trainer/backends/localprocess/job_test.py
@@ -1,0 +1,53 @@
+# Copyright The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


kubeflow/trainer/backends/localprocess/job.py
```diff
@@ -149,17 +150,11 @@ def cancel(self):
...
-            for chunk in self.stream_logs():
-                print(chunk, end="", flush=True)  # stream to console live
-        except StopIteration:
-            pass
-
-        return self._stdout.splitlines()
+        return self.stream_logs()
 
     def stream_logs(self):
         """Generator that yields new output lines as they come in."""
```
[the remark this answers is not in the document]
immanuwell: fixed, I updated `stream_logs()` to return `Iterator[str]` and changed the docstring to say it yields stdout chunks, not lines

kubeflow/trainer/backends/localprocess/job_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_logs_without_follow_returns_buffered_lines():
+    """Test LocalJob.logs returns buffered output when follow is disabled."""
+    job = LocalJob(name="test-job", command=["echo", "unused"])
+    job._stdout = "line-1\nline-2\n"
+
+    assert job.logs() == ["line-1", "line-2"]
+
+
+def test_logs_with_follow_returns_stream_without_printing():
```
andreyvelich: Please can you refactor this by using test cases: https://github.com/kubeflow/sdk/blob/main/AGENTS.md#3-testing-requirements

```suggestion
def test_logs():
```

andreyvelich (commented):
/ok-to-test
/cc @szaher @Fiona-Waters

kubeflow/trainer/backends/localprocess/job.py
```diff
@@ -149,20 +150,14 @@ def cancel(self):
     def returncode(self):
         return self._returncode
 
-    def logs(self, follow=False) -> list[str]:
+    def logs(self, follow: bool = False) -> list[str] | Iterator[str]:
```
Fiona-Waters, line 153: The return type is now a union where follow=False returns split lines (list[str]) and follow=True returns raw chunks (Iterator[str]). These are different units of data — consider having stream_logs() yield individual lines for a uniform contract across both branches. Also, this public method needs a Google-style docstring per project standards ([AGENTS.md section 5](https://github.com/kubeflow/sdk/blob/main/AGENTS.md#5-documentation-standards)).

Fiona-Waters (commented):
Thanks @immanuwell I've left a comment.

kubeflow/trainer/backends/localprocess/job_test.py
immanuwell: done

kubeflow/trainer/backends/localprocess/job.py
immanuwell: done

immanuwell:
@andreyvelich @Fiona-Waters addressed all suggestions, please take a look

kubeflow/trainer/backends/localprocess/job.py
```diff
@@ -149,30 +150,32 @@ def cancel(self):
...
+    def stream_logs(self) -> Iterator[str]:
+        """Yield newly available stdout lines as they are written by the process."""
+        emitted_line_count = 0
+        while self.is_alive() or emitted_line_count < len(self._stdout.splitlines()):
             self._output_updated.wait(timeout=1)
             with self._lock:
-                data = self._stdout
-                new_data = data[last_index:]
-                last_index = len(data)
+                lines = self._stdout.splitlines()
```
Fiona-Waters, line 174: Each iteration splits the entire accumulated stdout buffer. For long-running jobs with substantial output this is O(n) on every poll. The old code used a character index (last_index) which was O(1). A middle ground would be to track a character offset and only split the new data?

PR #559, opened 2026-07-06
chore(ci): enforce docstring, type annotations and type hardening

tariq-hasan:

This PR turns on the Google-style docstring enforcement as well as checking of type annotations.

Type hardening is also enabled and the bugs that have surfaced as a result of this check have been resolved.

The changes are scoped exclusively to the Spark package.

**Update:**

The changes now apply repo-wide.

Fixes #

**Checklist:**


pyproject.toml
```diff
@@ -123,13 +123,35 @@ select = [
...
+"kubeflow/optimizer/**" = ["D", "ANN"]
+"kubeflow/trainer/**"   = ["D", "ANN"]
+"test/**"               = ["D", "ANN"]
+# Non-package paths.
+".github/**"            = ["D", "ANN"]
+"docs/**"               = ["D", "ANN"]
+"examples/**"          = ["D", "ANN"]
+# Tests and test helpers don't need docstring/annotation rules.
+"**/*_test.py"          = ["D", "ANN"]
+"kubeflow/spark/test/**" = ["D", "ANN"]
```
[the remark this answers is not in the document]
tariq-hasan: The ruff and ty settings are set up so that the changes apply for Spark only.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -296,7 +302,7 @@ def get_spark_connect_info_from_cr(
 
     return SparkConnectInfo(
         name=spark_connect_cr.metadata.name,
-        namespace=spark_connect_cr.metadata.namespace,
+        namespace=spark_connect_cr.metadata.namespace or "",
         state=state,
```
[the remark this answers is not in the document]
tariq-hasan: The namespace is always expected to be populated in the Spark Connect CR by virtue of the fact that this is the `metadata.namespace` field from the API server. The `""` has only been done to make the `ty` typechecker work.

Replacing `""` with the `default` namespace is also a wrong choice since the client may not be tied to the `default` namespace.

I have instead failed explicitly if the namespace is missing.

pyproject.toml
```diff
@@ -123,13 +123,35 @@ select = [
...
+  "ANN", # flake8-annotations (public signatures fully typed)
 ]
 
 ignore = [
-  "B006", # mutable-argument-default
-  "E501", # Embedded HTML/SVG has long lines
+  "B006",   # mutable-argument-default
+  "E501",   # Embedded HTML/SVG has long lines
+  "ANN401", # Any is used intentionally
+  "D107",   # __init__ docstring optional (class docstring documents it)
```
andreyvelich: Why do we need to add them to ignores?

pyproject.toml
andreyvelich: Could you fix them in this PR to align with Google conventions that we use everywhere?

pyproject.toml
```diff
@@ -149,3 +171,33 @@ filterwarnings = ["ignore::pydantic.warnings.PydanticDeprecatedSince20"]
 
 [tool.ty.environment]
 python-version = "3.10"
+
+# The generated kubeflow_spark_api models use Pydantic aliases with
+# populate_by_name=True, so field-name kwargs (e.g. spark_version=,
+# main_application_file=, driver_info=) are valid at runtime. ty models the
+# alias-only __init__ and reports unknown-argument / missing-argument. Scope
+# those two rules down where we construct these models. Remove when ty gains
+# Pydantic-alias awareness.
```
andreyvelich: Do they have tracking issue for Pydantic-alias awareness?

pyproject.toml
tariq-hasan: I have removed the ignores and applied the docstring fixes repo-wide.

pyproject.toml
tariq-hasan: The ANN401 ignore allowed free usage of `Any` and the D107 ignore enabled docstrings to be avoided for the `__init__` constructor. I have removed these ignore rules.

pyproject.toml
tariq-hasan: There is a tracking issue but it has been closed: https://github.com/astral-sh/ty/issues/412.

I have removed this rule and applied necessary exemptions on the appropriate lines in the actual code.

tariq-hasan: I have also removed updated the `verify` target in the Makefile so that type checks are now enforced [repo-wide](https://github.com/tariq-hasan/sdk/blob/facb9f320d41eb001b5ce6971f223ce714b63a2c/Makefile#L66).

kubeflow/optimizer/types/algorithm_types.py
```diff
@@ -12,15 +12,16 @@
...
 import abc
 from dataclasses import dataclass, fields
-from typing import Any
 
 from kubeflow_katib_api import models
 
 
-def algorithm_to_katib_spec(obj: Any) -> models.V1beta1AlgorithmSpec:
-    """Convert any dataclass-based algorithm to a Katib AlgorithmSpec"""
+def algorithm_to_katib_spec(obj: "GridSearch | RandomSearch") -> models.V1beta1AlgorithmSpec:
```
andreyvelich, line 23: Do you see any problems to define types similar to how we set it here: https://github.com/tariq-hasan/sdk/blob/facb9f320d41eb001b5ce6971f223ce714b63a2c/kubeflow/trainer/api/trainer_client.py#L116C17-L118 ?

```suggestion
def algorithm_to_katib_spec(obj: types.GridSearch | types.RandomSearch) -> models.V1beta1AlgorithmSpec:
```

kubeflow/optimizer/types/optimization_types.py
```diff
@@ -43,7 +45,8 @@ class Objective:
     metric: str = "loss"
     direction: Direction = Direction.MINIMIZE
 
-    def __post_init__(self):
+    def __post_init__(self) -> None:
+        """Normalize a string direction into a Direction enum member."""
```
andreyvelich, line 49: Is that rule force us to define doc string for every function? Even if it is internal helper that we don't expose to the end-user?

PR #562, opened 2026-07-06
fix(trainer): don't report Complete when all container statuses are Unknown

### LINKED ISSUE #561: bug(trainer): container backend reports Complete when all container statuses are Unknown
### What happened?

`aggregate_status_from_containers()` in `kubeflow/trainer/backends/container/utils.py` decides a job is `Complete` with:

```python
if all(s == constants.TRAINJOB_COMPLETE for s in container_statuses if s != UNKNOWN):
```

When every status is `Unknown` (or the list is empty), the filtered generator is empty, so `all([])` returns `True` and the job is reported as `Complete`.

This is reachable through `aggregate_container_statuses()`: `get_container_status()` returns `UNKNOWN` whenever the adapter raises while inspecting a container. So a job whose containers cannot be read reports `Complete` instead of `Unknown`.

### Reproduction

```python
from kubeflow.trainer.backends.container.utils import aggregate_status_from_containers

aggregate_status_from_containers([])                      # 'Complete' (expected 'Unknown')
aggregate_status_from_containers(["Unknown"])             # 'Complete' (expected 'Unknown')
aggregate_status_from_containers(["Unknown", "Unknown"])  # 'Complete' (expected 'Unknown')
```

An empty or all `Unknown` set of container statuses should aggregate to `Unknown`, not `Complete`. `Complete` should require at least one known status and every known status being `Complete`.

Affected APIs on the container backend: `get_job()`, `list_jobs()`, `wait_for_job_status()`, `get_job_logs()`. A caller polling for `Complete` would treat an unreadable job as finished.

### Environment

* Kubeflow SDK: `main`
* Backend: container (local development)

SahilKumar75:
## What this does

Fixes a false positive `Complete` status in the container backend.

`aggregate_status_from_containers()` ran `all()` over container statuses filtered to exclude `UNKNOWN`:

```python
if all(s == constants.TRAINJOB_COMPLETE for s in container_statuses if s != UNKNOWN):
```

When the list is empty or every status is `UNKNOWN`, the filtered sequence is empty, so `all([])` returns `True` and the job is reported `Complete`.

This is reachable through `aggregate_container_statuses()`: `get_container_status()` returns `UNKNOWN` whenever the adapter raises while inspecting a container. A job whose containers cannot be read therefore reports `Complete` to `get_job()`, `list_jobs()`, `wait_for_job_status()` and `get_job_logs()`.

## The fix

Require at least one known status and every known status to be `Complete` before returning `Complete`. Empty or all `UNKNOWN` inputs now aggregate to `Unknown`. `Failed` and `Running` precedence and the `Created` fallback are unchanged.

## Testing

* Added parametrized unit tests for `aggregate_status_from_containers()`, which had no prior coverage: empty, all unknown, all complete, complete with unknown, failed and running precedence, created with unknown.
* `make test-python`: 401 passed.
* `make verify`: all checks passed.

Fixes #561

### DIFF
--- a/kubeflow/trainer/backends/container/utils.py
+++ b/kubeflow/trainer/backends/container/utils.py
@@ -158,7 +158,8 @@ def aggregate_status_from_containers(container_statuses: list[str]) -> str:
         return constants.TRAINJOB_FAILED
     if constants.TRAINJOB_RUNNING in container_statuses:
         return constants.TRAINJOB_RUNNING
-    if all(s == constants.TRAINJOB_COMPLETE for s in container_statuses if s != UNKNOWN):
+    known_statuses = [s for s in container_statuses if s != UNKNOWN]
+    if known_statuses and all(s == constants.TRAINJOB_COMPLETE for s in known_statuses):
         return constants.TRAINJOB_COMPLETE
     if any(s == constants.TRAINJOB_CREATED for s in container_statuses):
         return constants.TRAINJOB_CREATED
--- a/kubeflow/trainer/backends/container/utils_test.py
+++ b/kubeflow/trainer/backends/container/utils_test.py
@@ -14,7 +14,9 @@
 
 import pytest
 
+from kubeflow.common.constants import UNKNOWN
 from kubeflow.trainer.backends.container import utils as container_utils
+from kubeflow.trainer.constants import constants
 from kubeflow.trainer.test.common import FAILED, SUCCESS, TestCase
 from kubeflow.trainer.types import types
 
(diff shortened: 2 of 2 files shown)


andreyvelich (commented):
Thanks @SahilKumar75!
/ok-to-test
/retest
/lgtm
/approve

/cc @Fiona-Waters

brightyorcerf:
Question about the mixed case here. After this PR, ["Complete", "Unknown"] still returns Complete, the "complete with unknown is complete" test case. Was that deliberate?

Asking because get_container_status() returns Unknown whenever the adapter raises during inspect, and container_status_to_trainjob_status() also falls through to Unknown for paused/dead/restarting. So on a two-node job where node-0 exited 0 and node-1 can't be read, wait_for_job_status() returns Complete on that poll without node-1's outcome ever being determined.

Is the intent that an unreadable node shouldn't block a Complete verdict, or was the mixed case left for a follow-up?

SahilKumar75:
Thanks for pointing this out. The mixed case was deliberate in the narrow scope of this PR. The original issue focused on preventing an empty or entirely unknown status set from being reported as Complete, while preserving the existing treatment of unknown entries when at least one known status was available.

Your two node example shows that this preserved behavior can still report completion too early. If one container is Complete and another is Unknown because inspection failed, the overall result is not yet known and should probably remain Unknown rather than Complete.

Since this PR is already merged, I think the mixed case should be handled in a follow up issue with a regression test for `[Complete, Unknown]`. I can open that issue and propose the status precedence separately.

brightyorcerf:
Makes sense, I already have the fix ready, so I'll open it, if that works for you. 

Update: Opened #788 for the mixed case; [Complete, Unknown] now aggregates to Unknown, everything else keeps its current precedence. Would appreciate your review on it.

PR #570, opened 2026-07-07
fix(trainer): strictly type steps in LocalProcess backend to prevent NoneType iteration

### LINKED ISSUE #569: bug: TypeError: 'NoneType' object is not iterable in LocalProcess backend when iterating over steps
### What happened?

When calling methods like `get_job_logs()` or deleting a job in the `LocalProcess` backend (`kubeflow.trainer.backends.localprocess.backend.LocalProcessBackend`), the code directly iterates over `_job[0].steps` (e.g. `for _step in _job[0].steps:`) without checking if it is `None`. 

In `kubeflow/trainer/backends/localprocess/types.py`, the `steps` field is typed as `list[LocalBackendStep] | None = []`. If it is ever passed as `None` instead of an empty list, this results in a `TypeError: 'NoneType' object is not iterable`, crashing the process. (Note: This is currently being caught and reported by `uv run ty check` as well).

**Reproduction Steps**

```python
# Minimal reproduction example
from kubeflow.trainer.backends.localprocess.types import LocalBackendJobs
from kubeflow.trainer.backends.localprocess.backend import LocalProcessBackend

# Job with None steps
job = LocalBackendJobs(name="my-job", runtime=None, steps=None)

backend = LocalProcessBackend()
# Force inject the job for demonstration
backend._LocalProcessBackend__local_jobs = [job]

# This will crash with TypeError: 'NoneType' object is not iterable
backend.get_job_logs(name="my-job")


The code should gracefully handle the `None` case in all iterations by using a fallback (e.g., `for _step in (_job[0].steps or []):`), or the type definition should strictly enforce `list[LocalBackendStep]` by removing `| None` from the Pydantic model definition.


### Environment

Kubernetes version:
```bash
N/A - This is a static code issue in the Python SDK.

```
Kubeflow Trainer version:
```bash
N/A

```
Kubeflow Python SDK version:
```bash
$ main (latest commit)

```

### Impacted by this bug?

NaveenKumarG-dev:
Fixes a critical runtime bug in the `LocalProcess` backend where iterating over `_job[0].steps` could result in a `TypeError: 'NoneType' object is not iterable` if `steps` was passed as `None`. 

This PR modifies the `LocalBackendJobs` Pydantic model to properly enforce `steps` as an iterable list using `Field(default_factory=list)`, completely removing the `| None` ambiguity. Additionally, it cleans up a few related `invalid-parameter-default` type hinting errors in `job.py` flagged by the `ty check` analyzer.

Fixes #569

**Checklist:**

- [x] Tested locally (Ran `make verify`, `uv run ty check`, and `make test-python`)

### DIFF
--- a/kubeflow/trainer/backends/localprocess/job.py
+++ b/kubeflow/trainer/backends/localprocess/job.py
@@ -27,9 +27,9 @@ class LocalJob(threading.Thread):
         self,
         name,
         command: list | tuple[str] | str,
-        execution_dir: str = None,
-        env: dict[str, str] = None,
-        dependencies: list = None,
+        execution_dir: str | None = None,
+        env: dict[str, str] | None = None,
+        dependencies: list | None = None,
     ):
         """Creates a LocalJob.
 
--- a/kubeflow/trainer/backends/localprocess/job.py
+++ b/kubeflow/trainer/backends/localprocess/job.py
@@ -26,10 +26,10 @@ class LocalJob(threading.Thread):
     def __init__(
         self,
         name,
-        command: list | tuple[str] | str,
+        command: list[str] | tuple[str, ...],
         execution_dir: str | None = None,
         env: dict[str, str] | None = None,
-        dependencies: list | None = None,
+        dependencies: list["LocalJob"] | None = None,
     ):
         """Creates a LocalJob.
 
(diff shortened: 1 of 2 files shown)


lntutor (changes requested):
The model change is the right direction, but this PR has no regression coverage for the runtime failure it fixes. Please add a test that constructs a `LocalBackendJobs` instance without `steps` and exercises the LocalProcess path that previously iterated it (for example `get_job_logs` or job registration/listing), asserting it behaves as an empty list rather than raising `TypeError`. A construction-only assertion would not protect the caller-side iteration that motivated this fix.

PR #571, opened 2026-07-08
fix: preserve runtime kind in TrainJob runtimeRef

Sumitha007:
## Summary

This PR preserves the runtime kind (`TrainingRuntime` or `ClusterTrainingRuntime`) when constructing `TrainJob.runtimeRef`.

## Changes

- Added a `kind` field to the `Runtime` type.
- Propagated the runtime kind through the Kubernetes backend.
- Set `runtimeRef.kind` when creating a `TrainJob`.
- Updated unit tests to validate the correct runtime kind for both namespaced and cluster-scoped runtimes.

## Testing

Verified the Kubernetes backend test suite passes:

bash
pytest kubeflow/trainer/backends/kubernetes/backend_test.py

50 passed in 10.35s

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -594,6 +594,7 @@ class KubernetesBackend(RuntimeBackend):
 
         return types.Runtime(
             name=runtime_cr.metadata.name,
+            kind = runtime_cr.kind,
             trainer=utils.get_runtime_trainer(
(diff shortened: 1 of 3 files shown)


andreyvelich (commented):
@Sumitha007 You are working on this feature as part of this PR, right: https://github.com/kubeflow/sdk/pull/549 ?
/hold

Sumitha007:
I'm not sure whether this should be part of #549. I was treating it as an independent fix. If you prefer it to be included in #549, I'm happy to do that.

andreyvelich:
Let's fix it in #549
/close

PR #575, opened 2026-07-10
fix(trainer): validate HuggingFace storage_uri has both user and repo

SahilKumar75:
## What this does

Follow-up to #564. The HuggingFace initializers only rejected a `storage_uri` with an empty path, so URIs missing either component were still accepted and only failed later at download time:

* `hf://user/` (no repo name)
* `hf:///model` (no user name)

`HuggingFaceDatasetInitializer` also had no test coverage for its validation.

## The fix

Consolidate the validation into a shared `_validate_hf_storage_uri` helper used by both `HuggingFaceModelInitializer` and `HuggingFaceDatasetInitializer`, requiring a non-empty user and repo. Valid URIs such as `hf://user/model` are unchanged.

## Testing

* Added a parametrized test for `HuggingFaceDatasetInitializer` (previously uncovered) and new rejected-form cases for both initializers.
* `make test-python`: 404 passed.
* `make verify`: all checks passed.

### DIFF
--- a/kubeflow/trainer/types/types.py
+++ b/kubeflow/trainer/types/types.py
@@ -322,6 +322,19 @@ class BaseInitializer(abc.ABC):
     storage_uri: str
 
 
+def _validate_hf_storage_uri(storage_uri: str, entity: str) -> None:
+    """Validate an 'hf://<user_name>/<repo_name>' storage URI."""
+    if not storage_uri.startswith("hf://"):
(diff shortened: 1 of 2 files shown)


andreyvelich (commented):
/ok-to-test
/retest
/lgtm
/approve

Thanks @SahilKumar75!
/cc @akshaychitneni

PR #582, opened 2026-07-13
fix(trainer): Move TrainJob cleanup to finally block in get_runtime_packages

### LINKED ISSUE #580: Potential TrainJob resource leak in get_runtime_packages() when exceptions occur
### What happened?

While reviewing the Kubernetes backend implementation, I noticed that get_runtime_packages() does not guarantee cleanup of the temporary TrainJob it creates.

The current implementation creates a temporary TrainJob to inspect the runtime environment:

job_name = self.train(...)

self.wait_for_job_status(job_name)
print("\n".join(self.get_job_logs(name=job_name)))
self.delete_job(job_name)

The problem is that delete_job() is only called if all previous operations complete successfully.

If any of the following raises an exception:

wait_for_job_status()
get_job_logs()
or any unexpected exception after the TrainJob has been created

the method exits immediately, leaving the temporary TrainJob behind without cleanup.

As a result, temporary TrainJobs (and their associated Kubernetes resources) may remain in the cluster even though they were only created to inspect the runtime environment.

Why this matters

get_runtime_packages() is intended to be a helper/debugging utility.

If cleanup is skipped after a failure:

temporary TrainJobs remain in the cluster,
unnecessary Pods and other Kubernetes resources continue to exist,
users may need to remove them manually,
repeated failures can accumulate stale resources over time.

Since this helper creates temporary infrastructure resources, cleanup should be guaranteed regardless of whether the operation succeeds or fails.

Suggested Fix

Wrap the TrainJob lifecycle in a try/finally block so that cleanup is always executed.

Example:

job_name = self.train(...)

try:
    self.wait_for_job_status(job_name)
    print("\n".join(self.get_job_logs(name=job_name)))
finally:
    self.delete_job(job_name)

This ensures that the temporary TrainJob is deleted even if an exception occurs while waiting for completion or retrieving logs.

Environment

Kubeflow Python SDK version:

pip show kubeflow
Version: 0.4.1

This issue was identified while reviewing the Kubernetes backend implementation and appears to affect the cleanup flow of temporary TrainJobs created by get_runtime_packages().

Hi! This would be my first contribution to Kubeflow. I have already investigated this issue and prepared a fix locally. If this approach looks good, I'd be happy to work on it and open a pull request. I'd appreciate it if this issue could be assigned to me.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

kingabdulsalamsas111-lgtm:

This PR moves temporary TrainJob cleanup into a finally block in get_runtime_packages()`.

Previously, if  wait_for_job_status()  or  get_job_logs()  raised an exception, the temporary TrainJob would not be deleted. This change ensures that  delete_job()  is always called after a TrainJob has been created, preventing temporary TrainJobs from being left behind when an error occurs.

Fixes #580

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -267,7 +267,9 @@ class KubernetesBackend(RuntimeBackend):
                 trainer=types.CustomTrainer(
                     func=print_packages,
                     num_nodes=1,
-                    resources_per_node=({"cpu": 1} if runtime_copy.trainer.device != "gpu" else None),
+                    resources_per_node=(
+                        {"cpu": 1} if runtime_copy.trainer.device != "gpu" else None
+                    ),
                 ),
             )
             self.wait_for_job_status(job_name)
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -276,7 +276,14 @@ class KubernetesBackend(RuntimeBackend):
             print("\n".join(self.get_job_logs(name=job_name)))
         finally:
             if job_name is not None:
-                self.delete_job(job_name)
+                try:
+                    self.delete_job(job_name)
+                except Exception:
+                    logger.exception(
+                        "Failed to delete temporary TrainJob %s/%s created by get_runtime_packages",
+                        self.namespace,
+                        job_name,
+                    )
 
     def train(
         self,
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -260,18 +260,21 @@ class KubernetesBackend(RuntimeBackend):
 
         # Create the TrainJob and wait until it completes.
         # If Runtime trainer has GPU resources use them, otherwise run TrainJob with 1 CPU.
-        job_name = self.train(
-            runtime=runtime_copy,
-            trainer=types.CustomTrainer(
-                func=print_packages,
-                num_nodes=1,
-                resources_per_node=({"cpu": 1} if runtime_copy.trainer.device != "gpu" else None),
-            ),
-        )
-
-        self.wait_for_job_status(job_name)
-        print("\n".join(self.get_job_logs(name=job_name)))
-        self.delete_job(job_name)
+        job_name = None
+        try:
+            job_name = self.train(
+                runtime=runtime_copy,
+                trainer=types.CustomTrainer(
+                    func=print_packages,
+                    num_nodes=1,
+                    resources_per_node=({"cpu": 1} if runtime_copy.trainer.device != "gpu" else None),
+                ),
+            )
+            self.wait_for_job_status(job_name)
+            print("\n".join(self.get_job_logs(name=job_name)))
+        finally:
+            if job_name is not None:
+                self.delete_job(job_name)
 
     def train(
         self,
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -1037,11 +1037,9 @@ def test_list_runtimes(kubernetes_backend, test_case):
         ),
     ],
 )
-def test_get_runtime_packages(kubernetes_backend, mocker, test_case):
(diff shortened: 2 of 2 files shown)


kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -260,18 +260,21 @@ def print_packages():
...
+                    func=print_packages,
+                    num_nodes=1,
+                    resources_per_node=({"cpu": 1} if runtime_copy.trainer.device != "gpu" else None),
+                ),
+            )
+            self.wait_for_job_status(job_name)
+            print("\n".join(self.get_job_logs(name=job_name)))
+        finally:
+            if job_name is not None:
+                self.delete_job(job_name)
```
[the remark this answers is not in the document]
andreyvelich: That make sense to me @kingabdulsalamsas111-lgtm.

andreyvelich: Can you also update unit test to `assert_called_with` delete TrainJob call: 
https://github.com/kingabdulsalamsas111-lgtm/sdk/blob/e71b4af83b3ceba9de996096e1691552c6779190/kubeflow/trainer/backends/kubernetes/backend_test.py#L1040

Sridhar1030:
hey @kingabdulsalamsas111-lgtm , can u fix the DCO signoff

kubeflow/trainer/backends/kubernetes/backend.py
kingabdulsalamsas111-lgtm: Updated the unit test to verify that delete_job is called with the TrainJob returned by train() using assert_called_with.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -1037,15 +1037,23 @@ def test_list_runtimes(kubernetes_backend, test_case):
         ),
     ],
 )
-def test_get_runtime_packages(kubernetes_backend, test_case):
+def test_get_runtime_packages(kubernetes_backend, mocker, test_case):
```
andreyvelich: Instead of doing this, can yo use `kubernetes_backend` parameter directly, like here:
https://github.com/kingabdulsalamsas111-lgtm/sdk/blob/89251ea49544e9804d42346df1d6f866642e299f/kubeflow/trainer/backends/kubernetes/backend_test.py#L1386


```suggestion
def test_get_runtime_packages(kubernetes_backend, test_case):
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -260,18 +260,23 @@ def print_packages():
...
+                    num_nodes=1,
+                    resources_per_node=(
+                        {"cpu": 1} if runtime_copy.trainer.device != "gpu" else None
+                    ),
+                ),
+            )
+            self.wait_for_job_status(job_name)
+            print("\n".join(self.get_job_logs(name=job_name)))
+        finally:
+            if job_name is not None:
```
andreyvelich, line 278: Could you address this as well: https://github.com/kubeflow/sdk/pull/582#discussion_r3573690213 ?

kubeflow/trainer/backends/kubernetes/backend_test.py
kingabdulsalamsas111-lgtm: Thanks! I updated the unit test to use the existing kubernetes_backend fixture directly and removed the unnecessary mocker parameter, following the pattern used in the other tests in this file.

kubeflow/trainer/backends/kubernetes/backend.py
kingabdulsalamsas111-lgtm: Thanks for the suggestion! I wrapped the temporary delete_job() call in a try/except block and log cleanup failures with logger.exception(), so cleanup errors no longer mask the original exception from wait_for_job_status() or get_job_logs().    


Both comments have been addressed in the latest commit. Please let me know if you see anything else that should be improved.

andreyvelich (commented):
Thanks @kingabdulsalamsas111-lgtm!
/lgtm
/approve

PR #583, opened 2026-07-14
fix(trainer): don't report Complete when all container statuses are Unknown

anxkhn:

The container backend can report a TrainJob as `Complete` (success) when its
status is actually indeterminate.

`aggregate_status_from_containers` in
`kubeflow/trainer/backends/container/utils.py` reduces per-container statuses to
a single TrainJob status. After the `Failed`/`Running` checks it used:

```python
if all(s == constants.TRAINJOB_COMPLETE for s in container_statuses if s != UNKNOWN):
```

When every status is `Unknown` (or the input list is empty), the filtered
generator is empty, so `all()` is vacuously `True` and the function returns
`Complete` for a job whose status could not be determined.

This is reachable, not hypothetical. `get_container_status` returns `Unknown` on
any adapter error, and `aggregate_container_statuses` feeds those into
`aggregate_status_from_containers`, which `ContainerBackend.get_job()` uses to set
`TrainJob.status`. So if the Docker/Podman daemon is unreachable (or every
`container_status` call raises), `get_job(name).status` becomes `Complete`, and
`wait_for_job_status` returns as if training succeeded instead of continuing to
poll or timing out.

The fix guards the empty case: only report `Complete` when at least one status is
known (not `Unknown`) and every known status is `Complete`; otherwise
all-`Unknown` or empty inputs fall through to the existing `Created` check and
finally return `Unknown`. The `Failed`/`Running`/`Created` paths are unchanged.

Adds parametrized unit tests for `aggregate_status_from_containers` covering the
all-`Unknown` and empty-list cases (which fail before this change) plus
keep-green regressions for the existing paths.

Fixes #

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/container/utils.py
+++ b/kubeflow/trainer/backends/container/utils.py
@@ -158,7 +158,8 @@ def aggregate_status_from_containers(container_statuses: list[str]) -> str:
         return constants.TRAINJOB_FAILED
     if constants.TRAINJOB_RUNNING in container_statuses:
         return constants.TRAINJOB_RUNNING
-    if all(s == constants.TRAINJOB_COMPLETE for s in container_statuses if s != UNKNOWN):
+    known_statuses = [s for s in container_statuses if s != UNKNOWN]
+    if known_statuses and all(s == constants.TRAINJOB_COMPLETE for s in known_statuses):
         return constants.TRAINJOB_COMPLETE
     if any(s == constants.TRAINJOB_CREATED for s in container_statuses):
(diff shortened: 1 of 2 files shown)


Sridhar1030:
Thanks @anxkhn for the PR , can you rebase your pr , apart from that everything else looks good

anxkhn:
Thanks. While rebasing, I found that #562 merged the same fix into `main`. The rebase therefore dropped this commit as already applied. I’ll close this duplicate PR.

PR #588, opened 2026-07-14
fix(trainer): validate LoraConfig params and correctly pass falsy-but…

### LINKED ISSUE #579: LoraConfig/TorchTuneConfig accept invalid PEFT hyperparameters, and valid 0/[] values are silently dropped
### What happened?

While going through `kubeflow/trainer/types/types.py` and
`kubeflow/trainer/backends/kubernetes/utils.py`, I found two related gaps in the
TorchTune LoRA/PEFT fine-tuning path:

**1. `LoraConfig` has no validation on its hyperparameters.**

```python
from kubeflow.trainer import LoraConfig

LoraConfig(lora_rank=0)       # accepted, will fail deep inside TorchTune at runtime
LoraConfig(lora_rank=-8)      # accepted
LoraConfig(lora_alpha=-1)     # accepted
LoraConfig(lora_dropout=1.5)  # accepted, should be constrained to 0.0-1.0
```

None of these raise an error at construction time, so a typo or a bad config value
surfaces later as an opaque failure inside the TorchTune runtime instead of a clear
`ValueError` where the config was created.

**2. `batch_size=0`, `epochs=0`, and `lora_attn_modules=[]` are silently dropped.**

In `get_args_using_torchtune_config`:

```python
if fine_tuning_config.batch_size:   # falsy check — batch_size=0 is silently skipped
    args.append(f"batch_size={fine_tuning_config.batch_size}")

if fine_tuning_config.epochs:       # same issue for epochs=0
    args.append(f"epochs={fine_tuning_config.epochs}")
```

And in `get_args_from_peft_config`:

```python
if peft_config.lora_attn_modules:   # an explicit empty list is silently dropped too
    args.append(f"model.lora_attn_modules=[{','.join(peft_config.lora_attn_modules)}]")
```

If a user explicitly sets one of these to a falsy-but-valid value, it's dropped with
no error and no indication that anything happened.

- Constructing a `LoraConfig` with an out-of-range `lora_rank`/`lora_alpha`/`lora_dropout`
  should raise a `ValueError` immediately, with a clear message about which field and
  why — not fail later inside TorchTune.
- Explicitly-set falsy values (`batch_size=0`, `epochs=0`, `lora_attn_modules=[]`)
  should not be silently dropped from the generated trainer args.

### Environment

Kubeflow Python SDK version:
```bash
$ pip show kubeflow
Version: 0.4.1
```

Kubernetes / Kubeflow Trainer version: not applicable since this is reproducible in pure
Python, without a cluster, since it's purely in the SDK's dataclass and arg-building
logic

mohar-xe:

Fixes two related bugs in the TorchTune LoRA/PEFT configuration:

1. `LoraConfig` now validates hyperparameters at construction time, raising clear `ValueError` for `lora_rank <= 0`, `lora_alpha <= 0`, and `lora_dropout` outside `[0.0, 1.0]`. Previously these were accepted silently and surfaced later as opaque TorchTune runtime failures.

2. Argument-building utilities now use `is not None` instead of truthiness checks for `batch_size` and `epochs`, and always emit `lora_attn_modules`. Previously, explicitly-set falsy values (`batch_size=0`, `epochs=0`, `lora_attn_modules=[]`) were silently dropped with no error or indication.

Fixes #579

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -497,11 +497,11 @@ def get_args_using_torchtune_config(
         args.append(f"dtype={fine_tuning_config.dtype.value}")
 
     # Override the batch size if it is provided.
-    if fine_tuning_config.batch_size:
+    if fine_tuning_config.batch_size is not None:
         args.append(f"batch_size={fine_tuning_config.batch_size}")
 
     # Override the epochs if it is provided.
-    if fine_tuning_config.epochs:
+    if fine_tuning_config.epochs is not None:
         args.append(f"epochs={fine_tuning_config.epochs}")
 
     # Override the loss if it is provided.
@@ -563,8 +563,7 @@ def get_args_from_peft_config(peft_config: types.LoraConfig) -> list[str]:
             args.append(f"{arg_name}={value}")
 
     # Override the LoRA attention modules if they are provided.
-    if peft_config.lora_attn_modules:
-        args.append(f"model.lora_attn_modules=[{','.join(peft_config.lora_attn_modules)}]")
+    args.append(f"model.lora_attn_modules=[{','.join(peft_config.lora_attn_modules)}]")
 
     return args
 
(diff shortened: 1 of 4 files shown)


lntutor (approved):
Reviewed the validation and argument-generation changes. The `is not None` checks preserve explicitly supplied zero values, the empty attention-module override is now explicit, and the LoRA bounds checks cover valid boundaries plus invalid rank, alpha, and dropout values. The corresponding regression tests exercise each changed behavior. Approved.

PR #590, opened 2026-07-15
fix: preserve byte count for decimal Kubernetes memory units

### LINKED ISSUE #581: _memory_kubernetes_to_spark does not preserve byte count when converting decimal Kubernetes memory units (K/M/G/T)
## What happened?

`_memory_kubernetes_to_spark()` converts Kubernetes-style memory quantities into Spark/JVM-style memory strings before submitting a Spark Connect session.

Kubernetes resource quantities support two distinct families of memory units:

| Unit Family | Examples | Base |
|-------------|----------|------|
| Binary | `Ki`, `Mi`, `Gi`, `Ti` | Powers of 1024 |
| Decimal | `K`, `M`, `G`, `T` | Powers of 1000 |

The current implementation correctly maps binary suffixes (e.g. `Mi → m`, `Gi → g`).

However, decimal suffixes are effectively passed through after lowercasing (e.g. `M → m`, `G → g`) without converting the numeric value.

Spark/JVM memory suffixes (`k`, `m`, `g`, `t`) are always interpreted as **binary (1024-based)** units, regardless of case.

As a result, decimal Kubernetes quantities are silently interpreted as binary quantities, increasing the requested byte count.

### Example

| Input | Intended Bytes | Current Output | Spark Interprets As | Difference |
|-------|---------------:|---------------|--------------------:|-----------:|
| `512M` | 512,000,000 | `512m` | 536,870,912 bytes | +4.86% |
| `4G` | 4,000,000,000 | `4g` | 4,294,967,296 bytes | +7.37% |
| `1T` | 1,000,000,000,000 | `1t` | 1,099,511,627,776 bytes | +9.95% |

Depending on how a pod's memory limit is configured, this can cause the Spark driver or executor JVM to request more memory than the Kubernetes container is actually allowed, increasing the risk of an `OOMKilled` termination rather than a clean failure.

---

## Verification

This behavior is confirmed against Spark's implementation and the dependency versions currently used by this SDK.

- Spark's `JavaUtils.byteStringAsBytes` (used for parsing memory configuration values) treats **unsuffixed numeric values as bytes**.
- `Utils.memoryStringToMb` in Spark core explicitly converts values through bytes first because **a value without a suffix is assumed to already represent bytes**.
- Therefore, returning a plain integer (e.g. `512000000`) is an unambiguous and version-safe way to preserve the exact byte count.

Additionally:

- Spark only recently gained support for explicit Kubernetes-style suffixes (`Ki`, `Mi`, `Gi`, etc.) through **SPARK-55051** (merged January 2026).
- This repository currently targets **Spark 4.0.1**:
  - `pyspark-connect==4.0.1` (pinned in `pyproject.toml`)
  - `constants.DEFAULT_SPARK_VERSION = "4.0.1"`

Since Spark 4.0.1 predates SPARK-55051, relying on `Ki`/`Mi`/`Gi` suffixes would not be safe for the version currently supported by this SDK.

Returning the exact byte count as a plain integer avoids this compatibility issue.

---

## Steps to Reproduce

```python
from kubeflow.spark.backends.kubernetes.utils import _memory_kubernetes_to_spark

print(_memory_kubernetes_to_spark("512M"))
print(_memory_kubernetes_to_spark("4G"))
```

### Current Output

```text
512m
4g
```

---

## Expected Behavior

The conversion should preserve the original byte count exactly.

For example:

- `512M` (512,000,000 bytes)

should become

```text
512000000
```

rather than

```text
512m
```

because Spark interprets an unsuffixed integer as an exact byte count.

---
## Actual Behavior

Decimal Kubernetes units (`K`, `M`, `G`, `T`) are converted only by changing the suffix to lowercase without converting the numeric value.

Consequently, Spark interprets them as binary units, resulting in a larger memory allocation than originally requested.

---

## Suggested Fix

Handle binary and decimal Kubernetes units separately.

1. Parse the Kubernetes quantity and its suffix.
2. If the suffix is binary (`Ki`, `Mi`, `Gi`, `Ti`), keep the existing direct suffix mapping since these represent the same number of bytes.
3. If the suffix is decimal (`K`, `M`, `G`, `T`):
   - Compute the exact byte count (`value × 1000ⁿ`).
   - Return the result as a plain integer string (no suffix), which Spark interprets as bytes.
4. Leave already-Spark-format lowercase values (e.g. `512m`) unchanged for backward compatibility.

This preserves the requested memory exactly for both Kubernetes unit families while remaining compatible with the Spark version currently targeted by the SDK.

---

### Environment

- **Repository:** `kubeflow/sdk`
- **Branch:** `main`
- **Component:** `kubeflow.spark` (`SparkClient`)
- **Target Spark Version:** `4.0.1`

Can any maintainer please verify my approach so that I can implement the fix and generate a PR ?

### Impacted by this bug?

Sumitha007:
## Summary

Fixes the handling of Kubernetes decimal memory units in `_memory_kubernetes_to_spark()`.

Previously, decimal Kubernetes memory units (for example `512M` and `2G`) were converted directly to Spark/JVM memory suffixes (`512m` and `2g`). Since Spark interprets these suffixes as binary units, this could result in a larger memory allocation than intended.

This change preserves the original byte count by:
- Leaving existing Spark/JVM memory values (e.g. `512m`, `4g`) unchanged.
- Mapping Kubernetes binary units (`Ki`, `Mi`, `Gi`, `Ti`) to the corresponding Spark/JVM suffixes.
- Converting Kubernetes decimal units (`K`, `M`, `G`, `T`) to their exact byte count as an unsuffixed integer.

## Tests

- Updated the existing unit tests for memory conversion.
- Added test coverage for Kubernetes decimal memory units to verify that the exact byte count is preserved.

Fixes #581


Ishan-AI-coder:
Hi @Sumitha007 .Thanks for tackling this — the core idea (converting Kubernetes decimal units to an exact byte count instead of passing the suffix through) is right, and the M/G/T test cases match the issue. A few things before this can merge:

### 1. Lowercase `k` reintroduces the same bug it's fixing

Per the Kubernetes `resource.Quantity` spec, decimal-kilo is the *only* decimal SI suffix that's lowercase (`k`) — M/G/T/P/E are all uppercase. The current "already Spark format" passthrough check:

```python
if suffix in {"k", "m", "g", "t", "p", "e"}:
```

treats lowercase `k` as an already-Spark value and returns it unchanged:

```python
_memory_kubernetes_to_spark("512k")  # -> "512k"
```

Spark interprets `"512k"` as binary (1024-based), so this becomes `524288` bytes instead of the intended `512000` — the same class of bug described in #581, just shifted onto the `k` suffix (missed because the issue's own example table only covers M/G/T).

I'd suggest special-casing lowercase `k` into the decimal-conversion branch instead of the passthrough branch. Happy to share a suggested diff if that's helpful.

### 2. Branch appears to be based on a stale `main`

The diff shows `kubeflow/trainer/backends/kubernetes/backend.py` losing the `try/finally` cleanup around `get_runtime_packages`, and `test_aggregate_status_from_containers` being deleted even though `aggregate_status_from_containers` still exists and is called on `main`.

Checking the merge-base, this branch is missing exactly one upstream commit — **#582 "Move TrainJob cleanup to finally block in `get_runtime_packages`"** — which explains both. GitHub is diffing against the current `main`, so code you never touched shows up as "removed."

A quick rebase should make these unrelated changes disappear from the diff entirely:

```bash
git fetch upstream
git rebase upstream/main
```

### 3. `AddConnectJar` is a real, unrelated behavior change

Separately from the staleness issue, this PR also converts the always-on `connect_jar_url` default in `build_spark_connect_cr` into an opt-in `AddConnectJar` option.

The new class itself looks solid and matches the existing options pattern, but it changes the default behavior. Previously every session received the Ivy-cache workaround automatically (per the comment on `main`: *"avoid Ivy cache (container may not have writable ~/.ivy2)"*). After this change, users must explicitly pass `AddConnectJar()`, or their session can fail in containers with a read-only `~/.ivy2`.

Since this is a behavioral change independent of the memory-unit fix, I'd recommend splitting it into its own PR with its own description and discussion rather than including it alongside this bugfix.

Once this is rebased and scoped to just the `utils.py` memory-conversion fix (including the lowercase `k` handling), it should be quick to re-review.

Sumitha007:
Thanks for the review! I have addressed the feeback by rebasing onto the latest main removing the unrelated AddConnectJar changes, fixing the lowercase k handling for decimal Kubernetes memory units, and adding a test for the 512k case.

Ishan-AI-coder:
Re-reviewed after the update — this addresses both concerns from my earlier comment.

**Lowercase `k` collision (main issue):** fixed. The decimal-conversion branch now checks `suffix` by exact case (`k/K/M/G/T/P/E` → decimal bytes) instead of `.lower()`, so it no longer collides with the Spark-passthrough branch (`m/g/t/p/e`). Verified locally:

```text
512k -> 512000  ✓ (was 512k before, which Spark would've read as
                    binary/524288 bytes)
```

**Scope:** the trainer/`AddConnectJar` changes from the earlier version are gone — diff is now scoped to just `utils.py` + `utils_test.py`, as requested.

**Tests:** the new `("512k", "512000")` case covers exactly the regression I was worried about, and the old `("2G", "2g")` assertion — which was asserting the *original* bug — is correctly updated to `("2G", "2000000000")`. Ran the full `utils_test.py` locally: **37/37** pass.

One non-blocking note: legacy `"500Kb"`/`"2Gb"`-style suffixes now fall through unchanged rather than being stripped to `500k`/`2g`. I believe Spark's parser accepts the long-form `kb/mb/gb/tb` suffixes directly, so this is probably fine, but wasn't part of #581 either way — not something I'd hold this PR on.

**LGTM** — thanks for the quick turnaround on the `k` case, and good catch fixing it once it was pointed out.

Ishan-AI-coder:
Hey @Sumitha007 ! Just noticed this PR has a few merge conflicts with main that'll need resolving before it can be merged — might be worth rebasing when you get a chance.

Sumitha007:
Thanks for pointing it out! I've rebased the branch on the latest main, resolved the merge conflicts, and pushed the updated changes.

Ishan-AI-coder:
@tariq-hasan @astefanutti 
Would appreciate a review here..

Ishan-AI-coder:
Hey @Sumitha007 . Can you please resolve the conflicts here?

Sumitha007:
Thanks! I've resolved the merge conflicts, synchronized with the latest main, updated the affected tests to reflect the current implementation, and confirmed the changes by running the test suite locally (74 passed).

lntutor (changes requested):
There is one remaining correctness issue for fractional Kubernetes binary quantities.

`_memory_kubernetes_to_spark("1.5Gi")` currently returns `"1.5g"`, but Spark 4.0.1 rejects fractional byte strings: its `JavaUtils.byteStringAs` parses the numeric capture with `Long.parseLong`, and routes the fractional pattern to `NumberFormatException`. So a valid Kubernetes resource quantity becomes an invalid `spark.executor.memory` value.

Could we preserve this byte count using an integral representation instead (for example, `1.5Gi` → `1536m` or `1610612736`) and add a regression case that verifies the generated value is accepted by Spark’s byte-string parser? The existing integer `Gi` behavior can remain unchanged.

tariq-hasan:
Hi @Ishan-AI-coder and @Sumitha007! Thanks for the issue and the detailed PR implementation and review.

This needs a further rebase to address some merge conflicts on `utils_test.py`. In addition the fix needs to be overlaid on top of the existing [`_memory_kubernetes_to_spark`](https://github.com/Sumitha007/sdk/blob/50f1a99da11f0afd281a2e10acbc9c24ba8c8153/kubeflow/spark/backends/kubernetes/utils.py#L186-L241 ) starting at line 186 of `utils.py`.

I'm happy to take a further look once the conflicts are addressed.

Sumitha007:
Thanks for the review! I've rebased onto the latest main, resolved the merge conflicts, and verified the updated tests locally. I'll update _memory_kubernetes_to_spark() to preserve integer byte values for fractional binary quantities (e.g. 1.5Gi) and add the requested regression test.

Ishan-AI-coder:
Hi @Sumitha007, thanks for the fractional-unit fix (`858dbb5`) — nice catch on `1.5Gi` → `1536m`, that lines up with how Spark's `JavaUtils.byteStringAs` actually parses these (it uses `Long.parseLong` on the numeric part, so a fractional string like `1.5g` would throw `NumberFormatException`). Confirmed that's resolved correctly now.

A couple of things still worth addressing before this is ready for maintainer review:

1. **Duplicate `_memory_kubernetes_to_spark` in `utils.py`.** There are now two definitions of this function — the original at line 186 (already on `main`) and the new one added lower down, right after `validate_spark_connect_url` (~line 329). Diffing this branch against `main` shows that's literally the only change to `utils.py`, so the new implementation was added alongside the old one rather than replacing it. Since Python just uses whichever definition comes last, the line-186 version is now unreachable dead code. Could you fold the two into a single function? This is what's meant by "overlay the fix on top of the existing function."

2. **Regression test only checks the literal string, not that Spark would accept it.** The `1.5Gi` → `1536m` cases (in `test_memory_kubernetes_to_spark`, `test_resolve_driver_resources`, and `test_resolve_executor_resources`) all assert equality against the hardcoded string. It'd be stronger to also assert the output has an integer coefficient (no `.`) with a valid JVM suffix, so a future regression that reintroduces a fractional output would actually fail a test, rather than only failing if the exact string changes.

For what it's worth, `utils_test.py` itself looks clean — diffed against `main` and the only differences are the `2G` expected-output update and two blank-line removals, so no test cases were dropped or duplicated there.

Sumitha007:
Thank you for the review and the helpful feedback. I have updated the original _memory_kubernetes_to_spark() implementation and removed the duplicate function. The changes preserve decimal Kubernetes memory units while continuing to handle fractional binary units in a Spark-compatible way. I also reran the Spark Kubernetes utility tests, and all tests are now passing (74 passed). Please let me know if you have any further suggestions or if there are additional improvements I should make.

Ishan-AI-coder:
Hi @Sumitha007, thanks for pushing `05dfee4` — the duplicate `_memory_kubernetes_to_spark` is resolved now, there's a single consolidated function at line 186 and I traced all the test cases (including `2G` → `2000000000` and `1.5Gi` → `1536m`) through the logic and they all check out. Nice work folding the two implementations together cleanly.

One smaller thing left: the fractional-memory regression cases (`test_memory_kubernetes_to_spark`, and the driver/executor resource tests) only assert equality against the literal string `"1536m"`. It'd be a bit more robust to also assert the output has an integer coefficient and a valid JVM suffix (e.g. via a regex like `^\d+[kmgtp]$`), so a future change that reintroduces a fractional output (like the original `1.5g` bug) would fail the test even if someone changes the exact numeric value being tested against. Not blocking, just a nice-to-have if you're touching that file again.

Thanks again for iterating on this — looks solid otherwise.

Sumitha007:
Thanks for the suggestion! I updated the fractional-memory regression tests to also validate the JVM memory format using a regex while keeping the existing equality assertions. Thanks again for the helpful feedback.

Ishan-AI-coder:
Thanks @Sumitha007 for addressing the changes . @tariq-hasan @andreyvelich  @szaher   Would appreciate a look  when you have the  time .

Sumitha007:
Hello Maintainers,
Just a Follow-up regarding i raised this PR last month. Could you please review it when you get a time? Thank you!

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -228,16 +230,28 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
...
+        }
+        return str(int(float(coefficient) * decimal_units[suffix]))
 
     spark_suffix = {10: "k", 20: "m", 30: "g", 40: "t", 50: "p"}.get(exponent)
     if "." not in coefficient and spark_suffix is not None:
         return coefficient + spark_suffix
 
     total_bytes = math.ceil(float(coefficient) * (2**exponent))
+
     return f"{math.ceil(total_bytes / (2**20))}m"
```
tariq-hasan, line 256: The implementation here would need to be reworked using the suggestion [here](https://github.com/kubeflow/sdk/issues/581#issuecomment-5266935559). The Spark example E2E tests pass at this time on the PR since they do not exercise decimal suffixes. Adding in those cases in the E2E will help validate the approach.

Sumitha007: Thanks for the clarification. I'll rework the decimal unit conversion to normalize the values to MiB and add decimal-suffix coverage to the Spark E2E tests.

Sumitha007:
I've reworked the decimal Kubernetes memory conversion to normalize the values to MiB, added unit-test coverage for decimal units, and added decimal-memory coverage to the Spark E2E example.

HarshPopat23:
Thanks @Sumitha007  for reworking this according to @tariq-hasan’s MiB guidance. The `M/G/T` conversions and the `2G` E2E coverage now follow the intended direction, but I found a few remaining correctness gaps.

1. Kubernetes decimal kilo is lowercase `k`, but it still follows the Spark/binary passthrough path:

```python
_memory_kubernetes_to_spark("512000k")  # currently "512000k"
```

Kubernetes means 512,000,000 bytes here, while Spark interprets `k` as binary. With MiB normalization, this should become `489m`. The decimal suffix set should use `"k"` rather than `"K"`.

2. Decimal `E` is currently unreachable. `suffix_lower == "e"` fails the `exponent_by_suffix` check before reaching the decimal conversion branch, so `"1E"` is returned unchanged.

3. Could we use `decimal.Decimal` with `ROUND_CEILING` instead of `float()`? Kubernetes quantities are fixed-point, and valid large quantities can lose enough float precision to produce an off-by-one MiB result.

Please add unit cases for at least lowercase `k`, `P`, `E`, and a fractional decimal quantity. The PR description should also be updated because it still describes returning unsuffixed bytes, whereas the implementation now rounds up to MiB.

Apart from these conversion edge cases, I found no unrelated blocker in the three changed files.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1116,6 +929,7 @@ def test_resolve_executor_resources(test_case: TestCase) -> None:
         assert instances == constants.DEFAULT_NUM_EXECUTORS
         assert cores == 2
         assert memory == "1536m"
+        assert re.fullmatch(r"\d+[kmgtp]", memory)
 
     print("test execution complete")
```
tariq-hasan, line 1148: Please rebase on `main` to fix the merge conflict on `utils_test.py`.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -205,14 +200,16 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
         return memory
 
     match = re.match(
-        r"^(\d+(?:\.\d+)?)\s*([KMGTPE]i?|[kmgtp]b?)$",
+        r"^(\d+(?:\.\d+)?)\s*([KMGTPE]i?|k|m|g|t|p|e|kb|mb|gb|tb)?$",
```
tariq-hasan: This does not need to change. `r"^(\d+(?:\.\d+)?)\s*([KMGTPE]i?|[kmgtp]b?)$",` is perfectly valid. It captures both K8s and JVM-native forms.

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan, line 197: Let's make this more comprehensive.

```suggestion
    Kubernetes quantities use two unit families: binary (``Ki``..``Ei``, powers
    of 1024) and decimal SI (``k``, ``M``, ``G``, ``T``, ``P``, ``E``, powers of
    1000). Spark has no decimal units -- ``JavaUtils`` reads ``k``/``m``/``g``/
    ``t``/``p`` as powers of 1024 regardless of case -- so decimal quantities are
    resolved to an absolute MiB value rather than passed through. Fractional
    quantities (e.g. ``1.5Gi``) are also resolved to MiB, since Spark rejects
    fractional byte strings.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -233,16 +230,28 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
         "ei": 60,
     }
```
tariq-hasan, line 236: Let's add `e` as well.

```suggestion
"p": 50,
"pb": 50,
"ei": 60,
"e": 60,
}
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -205,14 +200,16 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
...
         memory,
         re.IGNORECASE,
     )
     if not match:
         return memory
 
-    coefficient, suffix = match.group(1), (match.group(2) or "").lower()
+    coefficient = match.group(1)
+    suffix = match.group(2) or ""
+    suffix_lower = suffix.lower()
```
tariq-hasan: We should keep the original case.

```suggestion
    # Keep the original case: Kubernetes decimal kilo is "k", while "K" is not a
    # valid Kubernetes suffix and is treated as Spark's binary "k".
    coefficient, suffix = match.group(1), match.group(2)
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -233,16 +230,28 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
...
+        }
+        total_bytes = math.ceil(float(coefficient) * decimal_units[suffix])
+        return f"{math.ceil(total_bytes / (2**20))}m"
     spark_suffix = {10: "k", 20: "m", 30: "g", 40: "t", 50: "p"}.get(exponent)
     if "." not in coefficient and spark_suffix is not None:
         return coefficient + spark_suffix
 
     total_bytes = math.ceil(float(coefficient) * (2**exponent))
+
     return f"{math.ceil(total_bytes / (2**20))}m"
```
tariq-hasan: Let's rewrite as follows. A few things occur in this change.

Firstly, `if suffix in {"K", "M", "G", "T", "P", "E"}:` won't match `k` which is what K8s uses for decimal. Secondly, the code uses fixed-point conversion which is more exact than the floating point approximation using `float()`. Thirdly, the decimal conversion table can be skipped using `multiplier = 1000 ** (exponent // 10)`. Fourthly, we round down to an exact integer MiB for decimal or fractional binary units.

```suggestion
    if suffix.lower() not in exponent_by_suffix:
        return memory

    exponent = exponent_by_suffix[suffix.lower()]

    # Kubernetes quantities are fixed-point, so scale the coefficient into an
    # integer and keep all arithmetic exact.
    whole, _, fraction = coefficient.partition(".")
    scaled_coefficient = int(whole + fraction)
    scale = 10 ** len(fraction)

    # Kubernetes decimal SI units (case-sensitive) -> absolute MiB.
    if suffix in {"k", "M", "G", "T", "P", "E"}:
        multiplier = 1000 ** (exponent // 10)
        mebibytes = scaled_coefficient * multiplier // (scale * 2**20)
        return f"{max(mebibytes, 1)}m"

    spark_suffix = {10: "k", 20: "m", 30: "g", 40: "t", 50: "p"}.get(exponent)
    if scale == 1 and spark_suffix is not None:
        return coefficient + spark_suffix

    mebibytes = scaled_coefficient * (2**exponent) // (scale * 2**20)
    return f"{max(mebibytes, 1)}m"
```

tariq-hasan (commented):
I have added a few more comments.

tariq-hasan:
Hi @Sumitha007! Please address the review comments so we can move forward with the PR.

Sumitha007:
Sorry for the delay, and thank you for your review and suggestions. I’ve addressed the comments, updated the memory conversion logic and tests, and fixed the formatting issues. The tests are passing now. Please take another look when you have time. Thank you

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -208,12 +208,12 @@ def _memory_kubernetes_to_spark(memory: str) -> str:
     match = re.match(
         r"^(\d+(?:\.\d+)?)\s*([KMGTPE]i?|[kmgtp]b?)$",
         memory,
-        re.IGNORECASE,
     )
```
tariq-hasan, line 211: Please bring back `re.IGNORECASE` - otherwise for example 2GB and 2Gb will be passed through unchanged while only 2gb will pass through the unit conversion machinery.

examples/spark/spark_advanced_options.py
```diff
@@ -297,6 +318,7 @@ def main():
 
     try:
         example_labels_and_annotations()
+        example_decimal_memory()
         if os.environ.get("SPARK_E2E_RUN_IN_CLUSTER") != "1":
             example_node_selection()
             example_tolerations()
```
tariq-hasan, line 324: Please remove unit conversion tests from `examples/spark/spark_advanced_options.py`. The goal of this specific E2E is to test the Spark options [surface](https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/options/kubernetes.py) - which is unrelated to the unit conversion logic.

Instead you can take the cpu and memory quantities in `examples/spark/spark_advanced_options.py` and replace some of the binary memory quantities with decimal memory quantities (retaining the actual scale of the memory) to test that the E2E passes without errors.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -170,9 +166,19 @@ def sample_function_with_args(name: str, age: int):
...
+            config={"k8s_memory": "512M"},
+            expected_output="488m",
+        ),
+        TestCase(
+            name="Convert decimal T to MiB",
+            config={"k8s_memory": "1T"},
+            expected_output="953674m",
         ),
         TestCase(
             name="Convert fractional Gi to Mi",
```
tariq-hasan, line 184: The test cases need to be more comprehensive. These are the different categories of test cases that need to go in here.

- decimal units (512M, 2G, 4G, 1T, 1P, 1E)
- lowercase decimal vs uppercase binary (512k vs 512K)
- fractional decimal and binary (1.5G, 1.3Gi)
- sub-MiB (tiny value)
- case-compatibility (1.5GB, 1.5Gb, 1.5MB)

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -183,7 +189,12 @@ def sample_function_with_args(name: str, age: int):
 )
 def test_memory_kubernetes_to_spark(test_case: TestCase) -> None:
     """Tests _memory_kubernetes_to_spark."""
-    assert _memory_kubernetes_to_spark(test_case.config["k8s_memory"]) == test_case.expected_output
+    result = _memory_kubernetes_to_spark(test_case.config["k8s_memory"])
+
+    assert result == test_case.expected_output
+
+    if test_case.config["k8s_memory"] == "1.5Gi":
+        assert re.fullmatch(r"\d+[kmgtp]", result)
```
tariq-hasan: We should not hardcode units in the actual test logic.

tariq-hasan (changes requested):
Hi @Sumitha007! Thanks for addressing the prior comments. Overall lgtm. I have included some additional comments - mostly around the tests.

PR #600, opened 2026-07-16
feat(spark): add typed enums aligned with Spark Operator

### LINKED ISSUE #594: KEP-107: Leverage typed enums from Python models

The goal of the issue is to use typed operator states (instead of bare strings) by pulling in enums directly from the generated Python models in the spark-operator repo.

The pattern would be generic and apply to other enums on the CRDs as well.

### Why is this needed?

This is needed to enable better detection of drift with upstream operator changes.

### Love this feature?

SIDDHESH1564:
## Summary
- Add `SparkApplicationType`, `SparkApplicationDeployMode`, and `SparkApplicationState` enums aligned with spark-operator Go constants
- Add `SparkJobStatus.from_operator_state` mapping through `SparkApplicationState`
- Export the new types from `kubeflow.spark`

`kubeflow_spark_api` still types operator state/type/mode fields as plain strings, so this PR mirrors the operator constants in the SDK for typed usage and drift detection until generated enums are available upstream.

Fixes #594

## Test plan
- [x] `uv run pytest -q kubeflow/spark/types/types_test.py`


tariq-hasan:
The https://github.com/kubeflow/sdk/issues/594 issue needs some investigation to determine what changes are needed on the Python models to expose the enums.
Let me close this PR for now.
We can raise a separate PR if enum generation is found to be feasible from the CRD schema and controller-gen tools.
/close

PR #604, opened 2026-07-17
fix(trainer): use writable generated script paths

danish9039:

Writes non-MPI CustomTrainer generated Python files and package-installation logs under `/tmp` instead of the image working directory.

Restricted non-root containers can have a root-owned working directory, which prevents the SDK from creating its generated files. The generated script restores the image working directory as the first Python import root while preserving existing import paths, including Python safe-path mode. MPI continues to use `DEFAULT_MPI_USER_HOME`.

This is required by kubeflow/trainer#3702, which intentionally preserves each image's configured working directory.

Follow-up to kubeflow/trainer#3702.

**Validation**:

- [x] 127 focused Kubernetes backend tests
- [x] Read-only working-directory execution in normal and safe-path modes
- [x] Working-directory precedence over a same-named `/tmp` module
- [x] Existing `PYTHONPATH` preservation
- [x] Concurrent behavioral execution
- [x] MPI path preservation
- [x] Ruff lint and formatting checks
- [x] `git diff --check`

**Checklist:**

- [ ] Documentation included if any changes are user facing

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -302,7 +302,7 @@ def get_custom_trainer(
         "if sys.path and os.path.realpath(sys.path[0]) == "
         "_kubeflow_generated_script_directory:\n"
         "    sys.path[0] = _kubeflow_working_directory\n"
-        "else:\n"
+        'elif not getattr(sys.flags, "safe_path", False):\n'
         "    sys.path.insert(0, _kubeflow_working_directory)\n\n"
         "def sample_train_func() -> None:\n"
         '    """Sample training function."""\n'
--- a/kubeflow/trainer/backends/kubernetes/backend_test.py
+++ b/kubeflow/trainer/backends/kubernetes/backend_test.py
@@ -288,19 +288,28 @@ def get_custom_trainer(
     install_script = utils.get_script_for_python_packages(
         packages_to_install=packages_to_install,
         pip_index_urls=pip_index_urls,
-        install_log_file="pip_install.log",
+        install_log_file="/tmp/pip_install.log",
     )
 
     # Append the embedded training function script that matches EXEC_FUNC_SCRIPT
     # with torchrun as the entrypoint and a fixed lambda for deterministic tests.
     func_script = (
         "\nread -r -d '' SCRIPT << EOM\n\n"
+        "import os\n"
+        "import sys\n\n"
+        "_kubeflow_generated_script_directory = os.path.realpath(os.path.dirname(__file__))\n"
+        "_kubeflow_working_directory = os.getcwd()\n"
+        "if sys.path and os.path.realpath(sys.path[0]) == "
+        "_kubeflow_generated_script_directory:\n"
+        "    sys.path[0] = _kubeflow_working_directory\n"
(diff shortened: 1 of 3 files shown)


lntutor (approved):
Reviewed the generated-script path change and its behavioral coverage. The non-MPI path keeps the image working directory first for imports while writing generated files and install logs under `/tmp`; MPI retains its existing user-home behavior. The read-only, safe-path, shadowing, `PYTHONPATH`, concurrent, and MPI cases cover the relevant execution paths. Approved.

danish9039:
Two updates since the approval.

1. `20501cb` makes the generated-script prelude honour Python safe path mode. The prelude restores the working directory as the first import root after the generated script moved to `/tmp`. Its `else` branch inserted that directory even when Python had deliberately left it out under `-P` or `PYTHONSAFEPATH=1`, whose entire purpose is to keep a writable directory off the import path. The branch is now `elif not getattr(sys.flags, "safe_path", False):`, so the working directory is added only when Python itself would have added a script directory. CPython made the same fix to `pdb` in https://github.com/python/cpython/pull/111762; `ipykernel` (https://github.com/ipython/ipykernel/issues/1311) and `pyright` (https://github.com/microsoft/pyright/issues/8168) resolved the same problem the same way. `getattr` rather than `sys.flags.safe_path` because the attribute only exists on 3.11+ and this package declares `requires-python = ">=3.10"`.
2. `6b8cd2a` fixes the `Test (Python 3.10)` failure that the first commit introduced: Python 3.10 has no safe path mode and ignores `PYTHONSAFEPATH`, so the prelude correctly falls back there and the safe-path test now asserts that fallback on 3.10 instead of the 3.11+ behaviour. Verified in `python:3.10-slim` and `python:3.12-slim`; `ruff check` and `ruff format` are clean.

| Case | Before | After |
|---|---|---|
| normal | working directory | working directory (unchanged) |
| normal with `PYTHONPATH` | working directory, `PYTHONPATH` behind it | unchanged |
| `-P` or `PYTHONSAFEPATH=1` (3.11+) | working directory, safe path defeated | `ModuleNotFoundError`, honoured |
| `-P` with `PYTHONPATH` (3.11+) | working directory ahead of `PYTHONPATH` | resolves from `PYTHONPATH` |
| Python 3.10, any mode | working directory | working directory (unchanged) |

Scoping note: this pull request fixes the generated script and package-installation log paths. It does not make a training function's own relative writes work; the Fashion-MNIST example still downloads into `./data` under the image working directory, which kubeflow/trainer#3702 now handles on the runtime side.

Separately, `kubeflow/trainer/constants/constants.py` uses `read -r -d '' SCRIPT << EOM` with an unquoted heredoc delimiter, so bash expands `$` and backticks inside the user's Python source before it is written. That is pre-existing and unrelated, so it is deliberately left out here; happy to raise it separately.

juliusvonkohout:
> Reviewed the generated-script path change and its behavioral coverage. The non-MPI path keeps the image working directory first for imports while writing generated files and install logs under `/tmp`; MPI retains its existing user-home behavior. The read-only, safe-path, shadowing, `PYTHONPATH`, concurrent, and MPI cases cover the relevant execution paths. Approved.

@andreyvelich for /approve
/lgtm

Issue #605, opened 2026-07-18
fix: create a new status set for each function call


vjkumar2756:
### What you would like to be added?

The SDK should ensure that a new status set is created for every function invocation when the status  parameter is  not  provided.
Currently, when  status is None, it should be initialized by creating a fresh set:
if status is None:
    status = {constants.TRAINJOB_COMPLETE}
This guarantees that each function call receives an independent set instance, preventing unintended sharing of mutable state  across invocations. The change follows Python best practices for handling mutable objects and  improves the reliability and maintainability of the SDK  without changing its  existing behavior.

### Why is this needed?

The SDK  should create a new status set whenever the status parameter is not provided. Creating a fresh set for each function invocation  ensures that mutable state is not unintentionally shared  across  calls.
Following Python's best practices for  mutable objects improves the reliability and maintainability of the code while preserving the existing behavior. This change makes the implementation safer, more predictable, and easier to reason about without introducing any user-facing changes.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

Sumitha007:
Hi @vjkumar2756, I'd like to work on this issue if it's available. I'll first verify the current implementation and submit a PR if the issue is still valid.

vjkumar2756:
@Sumitha007
  hi
 I have already completed the work for this issue and submitted a pull request for review.

Sumitha007:
Thanks for letting me know! I wasn't aware that you had already submitted a PR for this issue. I'll close my PR to avoid duplicate work.

andreyvelich:
Ref: https://github.com/kubeflow/sdk/pull/606#pullrequestreview-4740152678
/close

PR #606, opened 2026-07-18
fix: create a new status set for each function call

vjkumar2756:
This PR fixes the handling of the default status value by ensuring that a new set is created whenever status is not provided.
Previously, when status was None, the function did not explicitly create a fresh set for each invocation. This change initializes status as:
if status is None:
    status = {constants.TRAINJOB_COMPLETE}
By creating a new set on every function call, each invocation receives its own independent status object. This follows Python best practices for handling mutable objects, avoids unintended sharing of state between calls, and improves the reliability and maintainability of the code without changing the existing behavior.
Ensures each function call receives an independent status object.
Prevents potential issues related to mutable state being shared across invocations.
Preserves the existing functionality while improving code robustness.
Which issue(s) this PR fixes

Fixes #<issue-#605>.

### DIFF
--- a/kubeflow/trainer/api/trainer_client.py
+++ b/kubeflow/trainer/api/trainer_client.py
@@ -239,6 +239,11 @@ class TrainerClient:
         polling_interval: int = 2,
         callbacks: list[Callable[[types.TrainJob], None]] | None = None,
     ) -> types.TrainJob:
+        
+         if status is None:
+            status={constants.TRAINJOB_COMPLETE}
+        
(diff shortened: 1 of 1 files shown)


HarshPopat23:
Good work and approach @vjkumar2756 .

andreyvelich (commented):
Thanks for this PR @vjkumar2756, but we previously discussed that we don't get users feedback that they want to set status as None: https://github.com/kubeflow/sdk/pull/320#pullrequestreview-3849789318
/close

Issue #609, opened 2026-07-18
KEP-107: Align patterns and expand coverage for Spark package unit tests


tariq-hasan:
### What you would like to be added?

The goal of this issue is two-fold.
- Align test patterns for Spark package to match Trainer package unit tests.
- Improve coverage of test cases for Spark package unit tests.

### Why is this needed?

Consistency between the SDK components is necessary for maintainability.

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

Nagasirikancha:
Hi @tariq-hasan! I'd like to work on this issue if it's available.

I noticed that PRs #610, #614, and #621 address parts of the work described here. Are there still remaining test patterns or coverage gaps that you'd like someone to implement? If so, I'd be happy to take those on.

lntutor:
I reviewed the related open PRs (#610, #614, and #621) and the current Spark test suite. A remaining non-overlapping slice appears to be `kubeflow/spark/api/spark_client_test.py`: it currently covers construction only, while `SparkClient.connect()` mode selection/delegation, existing-server URL validation, token propagation, and list/get/delete/log delegation are not directly covered at the API layer.

Would a focused test-only PR for that API surface be welcome under this issue? I would avoid the backend initialization, backend `connect()`, and CR-spec utility areas already covered by those PRs.

PR #610, opened 2026-07-18
chore(spark): cover backend configuration loading

floze-the-genius:

Adds parameterized Spark SDK unit coverage for every Kubernetes backend configuration-loading path:

- explicit kubeconfig file
- explicit Kubernetes context
- successful in-cluster configuration
- fallback to the default kubeconfig when in-cluster configuration is unavailable

The tests also verify namespace selection and construction of both Kubernetes API clients. This follows the direct, parameterized assertion style used by Trainer SDK tests and covers Spark backend constructor branches that were not previously executed.

Part of #609.

**Validation**:

- `uv run pytest -q kubeflow/spark/backends/kubernetes/backend_init_test.py` - 4 passed
- `uv run coverage run --source=kubeflow.spark -m pytest -q kubeflow/spark` - 105 passed
- `uv run pytest -q ./kubeflow/` - 481 passed
- `uv lock --check`
- `uv run ruff check --show-fixes .`
- `uv run ruff format --check kubeflow`
- `uv run ty check kubeflow/hub`
- `uv run pre-commit run --all-files`

**AI assistance disclosure**:

OpenAI Codex was used to help prepare this test-only change. The resulting diff was reviewed and all validation commands above were run locally.

**Checklist:**

- [ ] Docs included if any changes are user facing (not applicable: test-only change)

### DIFF
--- a/kubeflow/spark/backends/kubernetes/backend_init_test.py
+++ b/kubeflow/spark/backends/kubernetes/backend_init_test.py
@@ -1,0 +1,108 @@
+# Copyright 2026 The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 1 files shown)


lntutor (commented):
Thanks for adding coverage of the configuration-loading branches. Could you add one case with both `config_file` and `context` set? `KubernetesBackend.__init__` deliberately gives `config_file` precedence via `if ... elif`, and an explicit test would protect that behavior from a future branch-order regression.

floze-the-genius:
Added the requested precedence case in `164c912`: when both `config_file` and `context` are set, the test now verifies that only `config_file` is passed to `load_kube_config` and in-cluster loading is not attempted.

Validation: `backend_init_test.py` now has 5 passing cases; targeted Ruff lint and format checks also pass.

PR #629, opened 2026-07-22
feat: [KEP-107] add FuncJob support for Python function submission

Goku2099:
## Summary

This PR introduces `FuncJob`, enabling users to submit Python functions directly as Spark batch jobs without creating a separate Python script.

### What's included

- Add `FuncJob` API for Python function submission.
- Package Python functions into executable scripts for Spark job execution.
- Support passing keyword arguments via `func_args`.
- Add comprehensive unit tests for `FuncJob`.
- Add an end-to-end example demonstrating the complete `FuncJob` lifecycle.

## Testing

- [x] Unit tests
- [x] E2E tests
- [x] Verified `batch_func_job_lifecycle.py` runs successfully


Goku2099:

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -614,13 +616,112 @@ def get_spark_job_executor_spec(
...
+
+    Raises:
+        ValueError:
+            If ``func`` is not callable.
+    """
+    if not callable(func):
+        raise ValueError(f"Expected a callable function, got {type(func)}.")
+
+    func_source = textwrap.dedent(
+        inspect.getsource(func),
```
andreyvelich, line 682: **Self-contained functions only.** `inspect.getsource` captures just the function body — module-level imports, globals, closures, and decorators are dropped, so any such reference raises `NameError`/`SyntaxError` in the driver pod. This breaks the KEP's own `etl_pipeline` example (module-level `from pyspark.sql import SparkSession`).

Please document in the `FuncJob` and `build_func_job_script` docstrings that the function must be fully self-contained (all imports **inside** the body, no closures/decorators), and consider rejecting decorated functions client-side (`getattr(func, "__wrapped__", None)`).

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -809,6 +810,41 @@ def _validate_file_job(
...
+            inspect.getsource(job.func)
+        except (OSError, TypeError) as e:
+            raise ValueError("`job.func` source code could not be retrieved.") from e
+
+        if job.func_args is not None:
+            if not isinstance(job.func_args, dict):
+                raise ValueError("`job.func_args` must be a dictionary.")
+
+            if not all(isinstance(key, str) for key in job.func_args):
+                raise ValueError("All `job.func_args` keys must be strings.")
```
andreyvelich, line 903: `func_args` values are embedded into the generated script via `repr()`, but only the keys are validated. Non-round-trippable values fail only in the driver — e.g. `datetime(...)` → `NameError`, arbitrary objects → `SyntaxError`.

Consider validating that each value is a JSON-like primitive (`str`/`int`/`float`/`bool`/`None`, nested `list`/`dict`/`tuple`) here and raising a clear `ValueError` at submit time, and add a test for a non-serializable value.

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,263 @@
...
+
+import os
+
+from kubeflow.common.types import KubernetesBackendConfig
+from kubeflow.spark import FuncJob, SparkClient, SparkJobStatus
+
+
+def estimate_pi(samples: int) -> None:
+    """Simple function executed as a Spark FuncJob."""
+    print(f"Estimating Pi with {samples} samples.")
```
andreyvelich: `estimate_pi` never creates a `SparkSession`, so the e2e test (which only checks for `FUNCJOB LIFECYCLE COMPLETE!`) never actually exercises a Spark FuncJob. Make it a real Spark workload with imports **inside** the function — this also validates the self-contained requirement:

```suggestion
def estimate_pi(samples: int) -> None:
    """Simple function executed as a Spark FuncJob."""
    from pyspark.sql import SparkSession

    spark = SparkSession.builder.appName("estimate-pi").getOrCreate()
    count = spark.range(samples).count()
    print(f"Estimated Pi over {count} samples.")
    spark.stop()
```

andreyvelich (commented):
Thanks a lot for driving FuncJob support forward, @Goku2099 — nice, well-tested work overall. I left a few inline comments below. Note: this review was produced with the help of AI tools, so please apply your own judgment on each point.

examples/spark/spark_failed_job.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
andreyvelich: This file isn't referenced anywhere in the repo and is unrelated to `FuncJob` (`batch_failed_job.py` still uses `REMOTE_PI`). Per scope discipline, please remove it from this PR, or wire it up in a separate PR.

kubeflow/spark/backends/kubernetes/constants.py
```diff
@@ -50,3 +52,20 @@
...
+FUNC_JOB_SCRIPT_DELIMITER = "__KUBEFLOW_FUNC_JOB_SCRIPT__"
+FUNC_JOB_SCRIPT_TEMPLATE = textwrap.dedent(
+    f"""
+    read -r -d '' SCRIPT << '{FUNC_JOB_SCRIPT_DELIMITER}'
+    {{func_code}}
+    {FUNC_JOB_SCRIPT_DELIMITER}
+
+    printf "%s" "$SCRIPT" > "{{func_file}}"
+    """
+)
```
andreyvelich, line 71: `read -r -d '' SCRIPT` always exits non-zero at EOF and only works because `printf` runs last — a future `set -e` (or a `printf` failure) would silently leave an empty `spark_job.py`. Prefer `cat` + `set -euo pipefail` + a non-empty check so the initContainer fails loudly:

```suggestion
FUNC_JOB_SCRIPT_TEMPLATE = textwrap.dedent(
    f"""
    set -euo pipefail
    cat > "{{func_file}}" << '{FUNC_JOB_SCRIPT_DELIMITER}'
    {{func_code}}
    {FUNC_JOB_SCRIPT_DELIMITER}
    test -s "{{func_file}}"
    """
)
```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -723,6 +727,106 @@ def test_defaults(self):
...
+    def sample_function():
+        """Simple function for testing."""
+        print("hello")
+
+    @staticmethod
+    def sample_function_with_args(name: str, age: int):
+        """Function with arguments for testing."""
+        print(name, age)
+
+    def test_build_script_without_args(self):
```
andreyvelich: These tests only assert substrings, so an assembly bug (bad indentation, a decorator artifact) would still pass while producing a script that fails only on the driver. Add `import ast` and `ast.parse(script)` in both tests to assert the generated script is valid Python.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -809,6 +810,41 @@ def _validate_file_job(
...
+        if inspect.iscoroutinefunction(job.func):
+            raise ValueError("Async functions are not supported.")
+
+        if job.func.__name__ == "<lambda>":
+            raise ValueError("Lambda functions are not supported.")
+
+        try:
+            inspect.getsource(job.func)
+        except (OSError, TypeError) as e:
+            raise ValueError("`job.func` source code could not be retrieved.") from e
```
andreyvelich: `OSError` (interactive/dynamically-defined function) and `TypeError` (builtin/C callable) have different, actionable causes but collapse into one generic message. Consider splitting them:

```suggestion
        try:
            inspect.getsource(job.func)
        except TypeError as e:
            raise ValueError(
                "`job.func` must be a pure-Python function; built-in or "
                "C-implemented callables are not supported."
            ) from e
        except OSError as e:
            raise ValueError(
                "`job.func` source could not be read. Functions defined "
                "interactively (REPL/Jupyter) or generated dynamically are "
                "not supported; define it in a Python module."
            ) from e
```

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,263 @@
...
+    print("Verified job has been deleted.")
+
+    JOB_NAME = None
+
+    print("\nDelete job example complete.\n")
+
+
+def main():
+    """Run all batch job lifecycle examples."""
+    print("E2E: Starting batch_job_lifecycle.py", flush=True)
```
andreyvelich: Wrong filename (copy-paste from `batch_job_lifecycle.py`):

```suggestion
    print("E2E: Starting batch_func_job_lifecycle.py", flush=True)
```

kubeflow/spark/backends/kubernetes/constants.py
andreyvelich: I think, it should be fine to use read since we do the same for TrainJobs.

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,268 @@
+#!/usr/bin/env python3
```
andreyvelich, line 1: @Goku2099 @tariq-hasan Are we planning to add Spark docs for `submit_job()` in the followup PR?
https://sdk.kubeflow.org/en/latest/spark/index.html

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,268 @@
...
+
+    print("=" * 70)
+    print("SUBMIT SPARK BATCH JOB")
+    print("=" * 70)
+
+    client = _client()
+
+    print("\nSubmitting Spark job...")
+
+    JOB_NAME = client.submit_job(
```
andreyvelich: Could we just make it simpler ?

```suggestion
def example_submit_and_wait():
    """Submit a Spark batch job and wait for completion."""
    global JOB_NAME

    print("=" * 70)
    print("SUBMIT SPARK BATCH JOB")
    print("=" * 70)

    client = SparkClient(
      backend_config=(
        KubernetesBackendConfig(
          namespace=os.environ.get("SPARK_TEST_NAMESPACE", namespace_default)
          )
        )

    print("\nSubmitting Spark job...")

    JOB_NAME = client.submit_job(
```

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,268 @@
...
+        },
+    )
+
+    print(f"Job submitted successfully: {JOB_NAME}")
+
+    print("\nWaiting for job to complete...")
+
+    job = client.wait_for_job_status(
+        JOB_NAME,
+        status={SparkJobStatus.COMPLETED},
```
andreyvelich: Can we set the Completed status as a default in `wait_for_job_status`, like we do in Trainer ?
https://github.com/Goku2099/sdk/blob/4d6d26d833d2a698413f1621d2b9ea5c2ee7d87c/kubeflow/trainer/backends/kubernetes/backend.py#L470
```suggestion
```

examples/spark/batch_func_job_lifecycle.py
```diff
@@ -0,0 +1,268 @@
...
+    print("=" * 70)
+    print("KUBEFLOW SPARKCLIENT - FUNCJOB LIFECYCLE")
+    print("=" * 70)
+
+    try:
+        example_submit_and_wait()
+        example_get_job()
+        example_list_jobs()
+        example_get_job_logs()
+        example_delete_job()
```
andreyvelich: We don't really need to create client for each of this call. Any particular reason you want to separate them by separate function?
@tariq-hasan @Shekharrajak @Goku2099 Maybe we should consider to convert our Spark examples using Jupyter Notebooks? Similar to Trainer example, that will give users simple reproducible artifacts they can run.

Also, majority of Spark users are running Notebooks.
This can also be a great reference for Kubeflow Notebooks v2 docs in the future.

cc @thesuperzapper @andyatmiami @christian-heusel

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -782,7 +783,8 @@ def _validate_job(
             return
 
         if isinstance(job, FuncJob):
-            raise NotImplementedError("Function-based jobs are not supported in Phase 1.")
+            self._validate_func_job(job)
```
andreyvelich, line 789: Could we create tracking issue to use same validation for TrainerClient and SparkClient.
For TrainJob, we currently only validate this: 
https://github.com/Goku2099/sdk/blob/4d6d26d833d2a698413f1621d2b9ea5c2ee7d87c/kubeflow/trainer/backends/kubernetes/utils.py#L348-L353

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -614,13 +616,118 @@ def get_spark_job_executor_spec(
...
+    )
+
+    if func_args is None:
+        invocation = f"    {func.__name__}()"
+    else:
+        invocation = f"    {func.__name__}(**{repr(func_args)})"
+
+    return f'{func_source}\n\nif __name__ == "__main__":\n{invocation}\n'
```
andreyvelich: Could we also use the same method to execute the function for now?
https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/utils.py#L376

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -849,14 +925,30 @@ def submit_job(
...
+                arguments=job.args,
+                num_executors=num_executors,
+                resources_per_executor=resources_per_executor,
+            )
+
+        else:
+            script = build_func_job_script(
+                func=job.func,
+                func_args=job.func_args,
+            )
```
andreyvelich: To ensure consistency between clients, can this util function generate command?

```suggestion
            command = get_command_using_spark_func(
                func=job.func,
                func_args=job.func_args,
            )
```

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -1060,6 +1231,35 @@ def test_submit_job(kubernetes_backend, test_case):
     print("test execution complete")
 
 
+def test_submit_func_job_uses_generated_script(kubernetes_backend):
```
andreyvelich: Can you just add another test cases to the `test_submit_job`

```suggestion
```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -723,6 +728,110 @@ def test_defaults(self):
         assert spec.service_account == constants.DEFAULT_SERVICE_ACCOUNT
 
 
+class TestGetFuncJobVolume:
```
andreyvelich: Same suggestion as before, can you ask LLM to refactor those tests using best practices defined in the AGENTS.md with test cases and single `test_` function per util function.

kubeflow/spark/backends/kubernetes/constants.py
```diff
@@ -50,3 +52,20 @@
...
+FUNC_JOB_MAIN_FILE = f"local://{FUNC_JOB_SCRIPT_DIR}/{FUNC_JOB_SCRIPT_NAME}"
+FUNC_JOB_SCRIPT_DELIMITER = "__KUBEFLOW_FUNC_JOB_SCRIPT__"
+FUNC_JOB_SCRIPT_TEMPLATE = textwrap.dedent(
+    f"""
+    read -r -d '' SCRIPT << '{FUNC_JOB_SCRIPT_DELIMITER}'
+    {{func_code}}
+    {FUNC_JOB_SCRIPT_DELIMITER}
+
+    printf "%s" "$SCRIPT" > "{{func_file}}"
+    """
```
andreyvelich, line 70: Any problem to use the same template as here?
https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/constants/constants.py#L134-L140

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -614,13 +616,118 @@ def get_spark_job_executor_spec(
...
+def get_func_job_volume_mount() -> models.IoK8sApiCoreV1VolumeMount:
+    """Build volume mount for function-based Spark jobs.
+
+    Returns:
+        Kubernetes volume mount for the generated Spark application.
+    """
+    return models.IoK8sApiCoreV1VolumeMount(
+        name=constants.FUNC_JOB_VOLUME_NAME,
+        mount_path=constants.FUNC_JOB_SCRIPT_DIR,
+    )
```
andreyvelich: We don't really need to separate this to util functions since it is used in a single place. I would remove them for simplicity.

```suggestion
```

andreyvelich (commented):
Thanks for this @Goku2099!
I left a few thoughts.

kubeflow/spark/backends/kubernetes/constants.py
tariq-hasan: I'm actually thinking the quoted `'{FUNC_JOB_SCRIPT_DELIMITER}'` delimiter is more robust against shell expansion, command substitution, etc. than the plain `EOM`.

I'm thinking if this should technically be moved to the `common` module and made reusable for both Trainer and Spark packages - possibly as a separate issue.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -661,6 +769,28 @@ def build_spark_application_cr(
...
+            get_func_job_init_container(func_script),
+        ]
+
+        spark_application.spec.driver.volume_mounts = [
+            volume_mount,
+        ]
+
+        spark_application.spec.executor.volume_mounts = [
+            volume_mount,
+        ]
```
tariq-hasan: This can be removed as only the driver container needs to read the file on the mount.

```suggestion
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -809,6 +811,80 @@ def _validate_file_job(
...
+        value: Any,
+    ) -> bool:
+        """Return whether a FuncJob argument value is supported."""
+
+        if value is None:
+            return True
+
+        if isinstance(value, (str, int, float, bool)):
+            return True
+
```
tariq-hasan: We'd want to reject non-finite floats since downstream [serialization](https://github.com/Goku2099/sdk/blob/4d6d26d833d2a698413f1621d2b9ea5c2ee7d87c/kubeflow/spark/backends/kubernetes/utils.py#L718) of `float('inf')`, for example, would just be `inf` - a value which by itself is not recognizable to Python.

```suggestion
        if isinstance(value, (str, int, bool)):
            return True

        if isinstance(value, float):
            return math.isfinite(value)

```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -809,6 +811,80 @@ def _validate_file_job(
...
+                "`job.func` must be a pure-Python function; built-in or "
+                "C-implemented callables are not supported."
+            ) from e
+        except OSError as e:
+            raise ValueError(
+                "`job.func` source could not be read. Functions defined "
+                "interactively (REPL/Jupyter) or generated dynamically are "
+                "not supported; define it in a Python module."
+            ) from e
+
```
tariq-hasan: This here to make the decorator detection logic more robust.

```suggestion
        try:
            func_source = textwrap.dedent(inspect.getsource(job.func))
        except TypeError as e:
            raise ValueError(
                "`job.func` must be a pure-Python function; built-in or "
                "C-implemented callables are not supported."
            ) from e
        except OSError as e:
            raise ValueError(
                "`job.func` source could not be read. Functions defined "
                "interactively (REPL/Jupyter) or generated dynamically are "
                "not supported; define it in a Python module."
            ) from e

        try:
            func_tree = ast.parse(func_source)
        except SyntaxError as e:
            raise ValueError("`job.func` source could not be parsed.") from e

        if any(
            isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and node.decorator_list
            for node in func_tree.body
        ):
            raise ValueError("Decorated functions are not supported.")

```

examples/spark/batch_func_job_lifecycle.py
Goku2099: Yes, we're planning to add the Spark documentation for submit_job() in a follow-up PR.

tariq-hasan (commented):
I have added a few comments. Thanks!

examples/spark/batch_func_job_lifecycle.py
Goku2099: @andreyvelich Thanks for the suggestion. I will simplify this example into a single linear workflow in this PR. I also opened https://github.com/kubeflow/sdk/issues/635 to track converting the Spark SDK examples to Jupyter notebooks, similar to the Trainer examples, so we can address that separately.

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: tracking issue : https://github.com/kubeflow/sdk/issues/637

kubeflow/spark/backends/kubernetes/constants.py
Goku2099: I created a tracking issue https://github.com/kubeflow/sdk/issues/638 for moving this template into a common module so Trainer and Spark can share the same implementation. For this PR, I'd prefer to keep the current quoted delimiter since it's more robust against shell expansion and command substitution.

andreyvelich: @tariq-hasan Any particular reason you think that `FUNC_JOB_SCRIPT_DELIMITER` might be better than EOM?

tariq-hasan: A quoted EOM would do as well. It’s more so in the fact that the quotes on the starting delimiter - be it FUNC_JOB_SCRIPT_DELIMITER or EOM - prevents the shell from interpreting the enclosed script so it passes it in as a raw string. This is just for robustness.

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -849,14 +939,30 @@ def submit_job(
...
+                func_args=job.func_args,
+            )
+
+            spark_application = build_spark_application_cr(
+                name=job_name,
+                namespace=self.namespace,
+                main_file=constants.FUNC_JOB_MAIN_FILE,
+                num_executors=num_executors,
+                resources_per_executor=resources_per_executor,
+                func_script=command,
```
andreyvelich: Can you refactor this util function to ensure get_command_using_spark_func returns the entire container command, like in Trainer.
 

```suggestion
                command=command,
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -614,13 +616,97 @@ def get_spark_job_executor_spec(
...
+    )
+
+    return models.IoK8sApiCoreV1Container(
+        name=constants.FUNC_JOB_INIT_CONTAINER_NAME,
+        image=constants.DEFAULT_SPARK_IMAGE,
+        command=[
+            "bash",
+            "-c",
+            init_script,
+        ],
```
andreyvelich:
```suggestion
        command=command
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -849,14 +939,30 @@ def submit_job(
...
+                resources_per_executor=resources_per_executor,
+            )
+
+        else:
+            command = get_command_using_spark_func(
+                func=job.func,
+                func_args=job.func_args,
+            )
+
+            spark_application = build_spark_application_cr(
```
andreyvelich: I would suggest to split this between two function, so we have less `if` conditions and None parameters there.
For common code, you can have dedicated util funcs.
For refernce: https://github.com/Goku2099/sdk/blob/e77e144ee757f0bb322900955cc9227cdd65571e/kubeflow/trainer/backends/kubernetes/backend.py#L816-L822


You can move `get_command_using_spark_func` script to the `get_spark_application_cr_from_func_job` function too.


```suggestion
            spark_application = utils.get_spark_application_cr_from_file_job(
                name=job_name,
                namespace=self.namespace,
                main_file=job.file_source,
                arguments=job.args,
                num_executors=num_executors,
                resources_per_executor=resources_per_executor,
            )

        else:
            spark_application = utils.get_spark_application_cr_from_func_job(
```

andreyvelich:
Thanks! 
/lgtm
/assign @tariq-hasan @Shekharrajak

tariq-hasan:
Thanks @Goku2099!
/lgtm

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -614,21 +616,103 @@ def get_spark_job_executor_spec(
...
+    if func_args is None:
+        func_call = f"{func.__name__}()"
+    else:
+        func_call = f"{func.__name__}(**{repr(func_args)})"
+
+    func_code = f"{func_code}\n{func_call}\n"
+    return [
+        "bash",
+        "-c",
+        constants.FUNC_JOB_SCRIPT_TEMPLATE.format(
```
abhijeet-dhumal, line 694: QQ: What happens if the generated source contains __KUBEFLOW_FUNC_JOB_SCRIPT__? 
Would the heredoc terminate early and leave the driver with a corrupt/empty script? Should _validate_func_job reject that case?

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -809,6 +814,91 @@ def _validate_file_job(
...
+        except SyntaxError as e:
+            raise ValueError("`job.func` source could not be parsed.") from e
+
+        if any(
+            isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)) and node.decorator_list
+            for node in func_tree.body
+        ):
+            raise ValueError("Decorated functions are not supported.")
+
+        if job.func_args is not None:
```
abhijeet-dhumal, line 898: Nit: Would it make sense to call 'inspect.signature(job.func).bind(**job.func_args)'
here so unknown/missing kwargs fail at submit time instead of on the driver?

abhijeet-dhumal:
Thanks @Goku2099 , it looks great 🚀 
Just left some nit picks.. otherwise LGTM !

kubeflow/spark/backends/kubernetes/backend.py
Goku2099: Added signature validation using inspect.signature(job.func).bind(**job.func_args) so invalid or missing keyword arguments fail during submission instead of at runtime.

kubeflow/spark/backends/kubernetes/utils.py
Goku2099: Added validation in _validate_func_job() to reject functions whose source contains the reserved heredoc delimiter, preventing premature heredoc termination.

andreyvelich (commented):
I think, we should be good to move this forward.
Thanks for this work @Goku2099!
/lgtm
/approve

PR #632, opened 2026-07-22
fix(trainer): deduplicate restarted TrainJob pods

### LINKED ISSUE #25: [SDK] Get the correct TrainJob components using `get_job()` API

reckless-sherixx:
## Summary

Fixes #25 by returning only the newest Pod for each TrainJob component in `get_job()`.

- Sorts Pods by `creation_timestamp` descending and keeps the first occurrence.
- Uses the component role for initializer keys and `role-index` for launcher/node keys.
- Adds a parametrized restart scenario covering both `node-0` and `node-1`, with newer Pending Pods taking precedence over older Running Pods.

## Why

Kubernetes can recreate Pods under restart policies. Without de-duplication, `get_job()` returns stale and current Pods for the same component, producing conflicting statuses.

## Validation

- `python -m uv lock --check`
- `python -m uv run ruff check --show-fixes --output-format=github .`
- `python -m uv run ruff format --check kubeflow`
- `python -m uv run ty check kubeflow/hub`
- `python -m uv run coverage run --source=kubeflow -m pytest ./kubeflow/` (629 passed)

Supersedes the focused deduplication work in #160 without its unrelated changes.

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/backend.py
+++ b/kubeflow/trainer/backends/kubernetes/backend.py
@@ -698,14 +698,32 @@ class KubernetesBackend(RuntimeBackend):
             if not pod_list:
                 return trainjob
 
-            for pod in pod_list.items:
+            sorted_pods = sorted(
+                pod_list.items,
+                key=lambda pod: (
+                    pod.metadata is not None and pod.metadata.creation_timestamp is not None,
+                    pod.metadata.creation_timestamp if pod.metadata else None,
+                ),
+                reverse=True,
+            )
+            seen_step_keys: set[str] = set()
+            for pod in sorted_pods:
                 # Pod must have labels to detect the TrainJob step.
                 # Every Pod always has a single TrainJob step.
                 if not (pod.metadata and pod.metadata.name and pod.metadata.labels and pod.spec):
                     raise Exception(f"TrainJob Pod is invalid: {pod}")
 
+                role = pod.metadata.labels[constants.JOBSET_RJOB_NAME_LABEL]
+                step_key = role
+                if role in {constants.LAUNCHER, constants.NODE}:
+                    step_key = f"{role}-{pod.metadata.labels[constants.JOB_INDEX_LABEL]}"
+
+                if step_key in seen_step_keys:
+                    continue
+                seen_step_keys.add(step_key)
+
                 # Get the Initializer step.
-                if pod.metadata.labels[constants.JOBSET_RJOB_NAME_LABEL] in {
+                if role in {
                     constants.DATASET_INITIALIZER,
                     constants.MODEL_INITIALIZER,
                 }:
@@ -717,7 +735,7 @@ class KubernetesBackend(RuntimeBackend):
                         )
                     )
                 # Get the Node step.
-                elif pod.metadata.labels[constants.JOBSET_RJOB_NAME_LABEL] in {
+                elif role in {
                     constants.LAUNCHER,
                     constants.NODE,
                 }:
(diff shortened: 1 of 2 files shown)


reckless-sherixx:
@andreyvelich Review this

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -210,6 +217,30 @@ def get_mock_pod_list():
     )
 
 
+def get_mock_pod_list_with_restarts() -> models.IoK8sApiCoreV1PodList:
+    """Create Pods where newer replacements share the same TrainJob component roles."""
+    old_timestamp = datetime.datetime(2025, 6, 1, 10, 0, 0)
+    new_timestamp = datetime.datetime(2025, 6, 1, 11, 0, 0)
+    old_pods = get_mock_pod_list().items
+    node_1_pod = copy.deepcopy(old_pods[-1])
```
andreyvelich: Select the node pod by role instead of position — `old_pods[-1]` silently breaks if `get_mock_pod_list()` is ever reordered.

```suggestion
    node_0_pod = next(
        p for p in old_pods if p.metadata.labels[constants.JOBSET_RJOB_NAME_LABEL] == constants.NODE
    )
    node_1_pod = copy.deepcopy(node_0_pod)
```

andreyvelich (commented):
Thanks for tackling this, @reckless-sherixx! This review was performed with the help of AI tools.

kubeflow/trainer/backends/kubernetes/backend_test.py
```diff
@@ -763,6 +795,25 @@ def get_train_job_data_type(
...
+    train_job_name: str,
+) -> types.TrainJob:
+    """Create the expected TrainJob after newer replacement Pods are selected."""
+    train_job = get_train_job_data_type(runtime_name, train_job_name)
+
+    for step in train_job.steps:
+        step.pod_name = f"{step.pod_name}-restarted"
+        step.status = constants.POD_PENDING
+
+    node_1_step = copy.deepcopy(train_job.steps[-1])
```
andreyvelich: Same positional fragility — clone the `node-0` step by name rather than by index.

```suggestion
    node_0_step = next(s for s in train_job.steps if s.name == "node-0")
    node_1_step = copy.deepcopy(node_0_step)
```

kubeflow/trainer/backends/kubernetes/backend.py
```diff
@@ -698,14 +698,32 @@ def __get_trainjob_from_cr(
             if not pod_list:
                 return trainjob
 
-            for pod in pod_list.items:
+            sorted_pods = sorted(
+                pod_list.items,
+                key=lambda pod: (
+                    pod.metadata is not None and pod.metadata.creation_timestamp is not None,
```
andreyvelich, line 704: This `is not None` guard exists to survive Pods with no `creation_timestamp`, but no test exercises it. A future simplification to `key=lambda pod: pod.metadata.creation_timestamp` would pass every test yet raise `TypeError: '<' not supported between 'datetime' and 'NoneType'` in production. Worth adding a case where one restarted Pod has `creation_timestamp=None` and the timestamped Pod still wins.

reckless-sherixx: added

andreyvelich (commented):
Thanks for this fix @reckless-sherixx!
/lgtm
/approve

PR #636, opened 2026-07-23
feat(trainer): support custom env and volumes for container backend

### LINKED ISSUE #624: feat(trainer): extend local execution backends with customizable runtime options

Currently, the local execution backends (`ContainerBackend` and `LocalProcessBackend`) provide baseline support for running training jobs locally without requiring a full Kubernetes cluster. However, advanced local customization is limited—for example, users cannot easily mount local volumes/directories, inject custom environment variables, or specify CPU/memory resource limits when running containers or sub-processes locally.

I propose extending the configuration capabilities of `ContainerBackendConfig` and `LocalProcessBackendConfig` to support additional developer-friendly options:

1. **`ContainerBackendConfig` Enhancements:**
   * Support for volume mounts (`volumes` / `volume_mounts`) to link local datasets directly into the execution container.
   * Configurable environment variables (`env`) passed into the container runtime.
   * Basic resource constraints (e.g., `--cpus`, `--gpus`, `--memory`).

2. **`LocalProcessBackendConfig` Enhancements:**
   * Support for passing custom environment variables to subprocesses.
   * Improved error capturing and cleanup on process termination/interruption.

### Why is this needed?

Spinning up a full KinD or Minikube cluster adds significant overhead and setup time for simple local debugging sessions. Additionally, running containers manually outside the SDK breaks the unified SDK workflow. Extending local backends improves the developer loop and makes testing pipelines locally much faster.

I would like to work on this issue!

### Love this feature?

Sumitha007:

This PR extends the local container backend to support additional runtime configuration.

### Changes
- Add `env` to `ContainerBackendConfig` to allow users to specify additional environment variables for local container execution.
- Add `volumes` to `ContainerBackendConfig` to allow users to mount additional host directories into training and initializer containers.
- Merge configured environment variables with the trainer environment before launching containers.
- Merge configured volume mounts with the default workspace mount for both training and initializer containers.
- Add unit tests to verify custom environment variables and volume mounts are propagated correctly.

Fixes #624

**Checklist:**

- [ ] Docs included if any changes are user facing (not required for this internal backend change)

### DIFF
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -335,6 +335,7 @@ class ContainerBackend(RuntimeBackend):
 
             # Build base environment
             env = container_utils.build_environment(trainer)
+            env.update(self.cfg.env)
 
             # Construct pre-run command to install packages
             pre_install_cmd = container_utils.build_pip_install_cmd(trainer)
@@ -428,6 +429,7 @@ class ContainerBackend(RuntimeBackend):
                         "mode": "rw",
                     }
                 }
+                volumes.update(self.cfg.volumes)
 
                 logger.debug(f"Creating container {rank}/{num_nodes}: {container_name}")
 
@@ -600,6 +602,7 @@ class ContainerBackend(RuntimeBackend):
                 "mode": "rw",
             }
         }
+        volumes.update(self.cfg.volumes)
 
         logger.debug(f"Starting {container_init.name} container: {container_name}")
 
--- a/kubeflow/trainer/backends/container/backend.py
+++ b/kubeflow/trainer/backends/container/backend.py
@@ -603,6 +603,10 @@ class ContainerBackend(RuntimeBackend):
             }
         }
         volumes.update(self.cfg.volumes)
+        container_init_env = {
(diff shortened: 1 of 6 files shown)


lntutor (commented):
The new `ContainerBackendConfig.env` is only merged into training containers. `_run_single_initializer()` still passes `environment=container_init.env`, so backend-level variables do not reach dataset/model initializer containers. That is an inconsistent implementation of backend-wide container runtime configuration, and it matters for shared settings such as proxy or CA variables needed while initializers fetch data/models. Please merge config env into initializer env as well, preserving initializer-specific values on key conflicts (for example `{**self.cfg.env, **container_init.env}`), and add a regression test that inspects an initializer container rather than only a training container.

Sumitha007:
Thanks for the review! I've updated the initializer environment handling to merge backend and initializer-specific environment variables, and added a regression check to the existing initializer tests to verify the environment is propagated to initializer containers.

Sumitha007:
Hello Maintainers,
Just a Follow-up regarding i raised this PR 3 weeks ago. Could you please review it when you get a time? Thank you!

PR #651, opened 2026-07-25
feat(spark): support connect_timeout when connecting via base_url

### LINKED ISSUE #650: feat(spark): support connect_timeout when connecting to an existing Spark Connect server via base_url
What you would like to be added?

SparkClient.connect() already supports timeout and connect_timeout when creating a new Spark Connect session through create_and_connect().
However, when base_url is provided, the method directly calls:

SparkSession.builder.remote(base_url).getOrCreate()

without applying any timeout handling.
We should extend the existing connect_timeout behavior to the base_url connection path so users can configure a maximum time to wait when connecting to an existing Spark Connect server.

Example:

from kubeflow.spark import SparkClient

client = SparkClient(...)

spark = client.connect(
    base_url="sc://spark-connect:15002",
    connect_timeout=30,
)

If the connection cannot be established within the configured timeout, the SDK should raise a clear TimeoutError, similar to the existing session creation flow.

Why is this needed?

The session creation flow already enforces connect_timeout, but the direct base_url connection path does not.

As a result, applications connecting to an existing Spark Connect server may block indefinitely if the endpoint is unreachable or unresponsive.

Supporting connect_timeout consistently in both connection paths would provide a uniform API and allow applications to fail fast with a clear TimeoutError.

Love this feature?

Sumitha007:

This PR adds `connect_timeout` support when connecting to an existing Spark Connect server via `base_url` in `SparkClient.connect()`.

Previously, `connect(base_url=...)` called `SparkSession.builder.remote(...).getOrCreate()` directly, which could block indefinitely if the server was unreachable or unresponsive.

This change wraps `getOrCreate()` in a background thread and applies the existing `connect_timeout`, making the behavior consistent with the `create_and_connect()` path.

## Which issue(s) this PR fixes

Fixes #650

## Tests

- Added unit test for successful connection via `base_url`
- Added unit test for connection timeout
- Added unit test for exception propagation
- All SparkClient unit tests pass

## Checklist

- [x] Code follows the project style
- [x] Unit tests added for the new functionality
- [ ] Docs included if any changes are user facing (not required)


lntutor (commented):
The timeout wrapper mirrors the existing `KubernetesBackend.connect()` strategy and preserves direct exception propagation. One small regression gap: please make the successful `base_url` test pass a token and assert `mock_builder.config("spark.connect.authenticate.token", token)` before `getOrCreate()`. The production branch still contains that configuration, but this test currently would not catch it being dropped while changing this connection path.

Sumitha007:
Thanks for the suggestion! I've updated the success test to pass a token and added an assertion verifying that builder.config("spark.connect.authenticate.token", token) is called before getOrCreate(). The updated test passes.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -162,12 +159,92 @@ def test_submit_job_success(job):
...
+    with (
+        patch("kubeflow.spark.api.spark_client.validate_spark_connect_url"),
+        patch("kubeflow.spark.api.spark_client.SparkSession.builder") as mock_builder_factory,
+    ):
+        mock_builder_factory.remote.return_value = mock_builder
+
+        client = SparkClient()
+
+        with pytest.raises(RuntimeError, match="boom"):
+            client.connect(base_url="sc://localhost:15002")
```
tariq-hasan, line 251: The unit tests need to be fixed to pass the CI. In addition, let's use the `pytest` and `TestCase` pattern seen for other tests.

kubeflow/spark/api/spark_client.py
```diff
@@ -110,7 +111,30 @@ def connect(
...
+            if not thread.is_alive():
+                if exc_holder:
+                    raise exc_holder[0]
+                if result:
+                    return result[0]
+
+            raise TimeoutError(
+                f"Spark Connect connection to {base_url} did not complete "
+                f"within {connect_timeout}s."
+            )
```
tariq-hasan, line 126: Since we're reusing a standard [pattern](https://github.com/Sumitha007/sdk/blob/ecf269c0f491e72f8df18652f45e9b63961826e4/kubeflow/spark/backends/kubernetes/backend.py#L609-L627) let's move this logic to the utils module and reuse as applicable.

tariq-hasan (changes requested):

Sumitha007:
Thanks for the review! I have addressed the requested changes, resolved the merge conflicts, and reran the relevant unit tests—they are all passing. Could you please review the latest changes when you have time?

Sumitha007:
Hello Maintainers
Just a follow-up regarding i raised this PR two weeks ago.  Could you please review it when you get a chance. Thank You!

PR #657, opened 2026-07-26
fix(trainer): preserve local job runtime metadata

### LINKED ISSUE #653: LocalProcessBackend.list_jobs() returns incorrect runtime metadata for unfiltered job listings
### What happened?

While reviewing the `LocalProcessBackend`, I compared the implementations of `list_jobs()` and `get_job()` to make sure both APIs return consistent job information.

During the review, I noticed that `get_job()` correctly returns the runtime stored inside each job by using:

runtime=_job.runtime,

However, `list_jobs()` behaves differently.

In `kubeflow/trainer/backends/localprocess/backend.py`, every `TrainJob` returned by `list_jobs()` is created using the method parameter:

runtime=runtime,

instead of the runtime that actually belongs to the job:

runtime=_job.runtime,
```

This becomes a problem when `list_jobs()` is called without a runtime filter, which is also the default behavior.

For example:

backend.list_jobs()
```

Since the `runtime` parameter is `None`, every returned `TrainJob` is populated with:

runtime=None
```

even though each job already has a valid runtime stored internally.

At the same time, requesting the exact same job through `get_job()` returns the correct runtime because it uses `_job.runtime`.

So the same job exposes different runtime information depending on which API is used. This makes the two APIs inconsistent and causes incorrect runtime metadata to be returned by `list_jobs()`.


Affected File**


kubeflow/trainer/backends/localprocess/backend.py


*Current Code**


result.append(
    types.TrainJob(
        name=_job.name,
        creation_timestamp=_job.created,
        runtime=runtime,
        num_nodes=1,
        steps=[...],
    )
)
```

 **Expected Code**

result.append(
    types.TrainJob(
        name=_job.name,
        creation_timestamp=_job.created,
        runtime=_job.runtime,
        num_nodes=1,
        steps=[...],
    )
)

### **Why this matters**

`list_jobs()` should always return the runtime associated with each stored job. Instead, it returns the runtime filter passed to the function. When no filter is supplied, every returned `TrainJob` contains `runtime=None`, even though the actual runtime exists. This results in inconsistent API behavior and incorrect metadata being returned to callers.

list_jobs() should always return each TrainJob with its actual associated runtime (_job.runtime), regardless of whether a runtime filter is provided. The returned runtime information should be consistent with get_job().

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

lntutor:
Summary:
- Return each local job's stored runtime from unfiltered listings.
- Add regression coverage for list_jobs without a runtime filter.

Fixes #653

Validation:
- make test-python
- make verify
- uv run pytest kubeflow/trainer/backends/localprocess/backend_test.py -q

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -156,7 +156,7 @@ class LocalProcessBackend(RuntimeBackend):
                 types.TrainJob(
                     name=_job.name,
                     creation_timestamp=_job.created,
-                    runtime=runtime,
+                    runtime=_job.runtime,
                     num_nodes=1,
                     steps=[
                         types.Step(name=s.step_name, pod_name=s.step_name, status=s.job.status)
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -411,6 +411,28 @@ def test_list_jobs(local_backend, test_case):
     assert isinstance(jobs, list)
 
 
+def test_list_jobs_without_runtime_filter_preserves_stored_runtime(
+    local_backend, mock_train_environment
+):
+    """Test unfiltered job listings retain each job's runtime metadata."""
+    runtime = types.Runtime(
+        name=TORCH_RUNTIME,
+        trainer=types.RuntimeTrainer(
+            trainer_type=types.TrainerType.CUSTOM_TRAINER,
+            framework="torch",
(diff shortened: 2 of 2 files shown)


kingabdulsalamsas111-lgtm:
Thanks for picking this up! I reported the issue in #653, and it's great to see both the fix and a regression test included. Looks good to me.

PR #673, opened 2026-07-27
fix(hub): update type ignore comments for ty syntax

pboyd:

Update type ignore comments from mypy-style (type: ignore[arg-type])
to ty-style (ty: ignore[invalid-argument-type]). This ensures consistency
with the project's type checking toolchain.

**Checklist:**

### DIFF
--- a/kubeflow/hub/api/model_registry_client.py
+++ b/kubeflow/hub/api/model_registry_client.py
@@ -91,7 +91,7 @@ class ModelRegistryClient:
         self._registry = ModelRegistry(
             server_address=base_url,
             port=port,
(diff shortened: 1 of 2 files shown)


pboyd:
This updates the ignore comments under `kubeflow/hub` to support more recent versions of `ty`. It doesn't upgrade `ty` though. This fails on main:

```sh
uv lock --upgrade-package ty
make verify
```

But it should work on this branch. This should fix the GHA jobs on #649.

andreyvelich (commented):
Thanks for the fix @pboyd!
/lgtm
/approve

PR #680, opened 2026-07-28
chore(deps): Bump Kubeflow API dependencies

andreyvelich:
This PR set upper limits for Kubeflow API dependencies.
Given that we might introduce breaking changes for alpha API, I believe this approach is safer for now.

Ref discussion: https://cloud-native.slack.com/archives/C08KJBVDH5H/p1775776960309219?thread_ts=1775776033.630949&cid=C08KJBVDH5H

I will also submit the Trainer PR to fix the CI.

cc @kubeflow/kubeflow-sdk-team @Fiona-Waters @Sridhar1030 @tariq-hasan @mszadkow @robert-bell @akshaychitneni @abhijeet-dhumal @jaiakash @jonburdo

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,8 +29,8 @@ dependencies = [
   "kubernetes>=27.2.0,<36.0.0",  # 36.0.0 has multiple regressions; see kubernetes-client/python#2592, #2596
   "pydantic>=2.10.0",
   "requests>=2.25.0",
-  "kubeflow-trainer-api>=2.2.0",
-  "kubeflow-katib-api>=0.19.0",
+  "kubeflow-trainer-api>=2.2.0,<2.3",
+  "kubeflow-katib-api>=0.19.0,<0.20",
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -29,8 +29,8 @@ dependencies = [
   "kubernetes>=27.2.0,<36.0.0",  # 36.0.0 has multiple regressions; see kubernetes-client/python#2592, #2596
   "pydantic>=2.10.0",
   "requests>=2.25.0",
-  "kubeflow-trainer-api>=2.2.0",
-  "kubeflow-katib-api>=0.19.0",
+  "kubeflow-trainer-api>=2.2.0,<2.3",
```
robert-bell, line 32: Is the plan to support a broader range of versions/remove this once the api graduates to beta1?

I think this is fine for alpha while breaking changes are likely but longer term we should aim for a broader range of versions.

robert-bell (approved):
Thanks @andreyvelich. No objections if it unblocks ci.

/lgtm

pyproject.toml
andreyvelich: Probably, we can talk about it. Given that we are going to maintain several APIs under Trainer: OptimizationJob, TrainJob, TrainingRuntime, they might different GA lifecycle.
So we need to be careful with the supported version to not break previous Kubeflow SDK releases.

andreyvelich:
This should unblock the E2Es: https://github.com/kubeflow/trainer/pull/3823
/approve

pyproject.toml
robert-bell: Yeah that's what I'm thinking, plus users might have different versions of the various control planes installed.

PR #684, opened 2026-07-29
feat(ci): Allow to configure extras and groups in make install-dev

andreyvelich:
This should allow CI to only install needed dependencies. Since PySpark ~ 900MB we are seeing some OOM errors on GPU runners: https://github.com/kubeflow/trainer/pull/3823

Hopefully this will fix it: https://github.com/kubeflow/trainer/pull/3823/changes/24f5f05535093d915a0a4a60a8c2c81f3a9795fb

@kubeflow/kubeflow-sdk-team @Fiona-Waters @Sridhar1030 @tariq-hasan @robert-bell @jaiakash

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,10 +54,6 @@ uv: ## Install UV
 	  echo "✅ uv has been installed."; \
 	}
 
-.PHONY: ruff
-ruff: ## Install Ruff
-	@uv run ruff --help &> /dev/null || uv tool install ruff
-
 .PHONY: verify
 verify: install-dev  ## install all required tools
 	@uv lock --check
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -54,10 +54,6 @@ uv: ## Install UV
 	  echo "✅ uv has been installed."; \
 	}
 
-.PHONY: ruff
-ruff: ## Install Ruff
-	@uv run ruff --help &> /dev/null || uv tool install ruff
-
```
robert-bell, line 60: did you mean to remove ruff?

robert-bell (approved):
Assuming ruff was removed intentionally, happy to sign off on this.

/lgtm

Makefile
andreyvelich: Yeah, I don't think it is needed since it will be automatically installed like `ty` as part of `dev` packages.

Similar to this we don't have target for `ty`.

robert-bell: that sounds fine to me.

Issue #520, opened 2026-06-03
[Tracking]: Spark Client Batch Jobs, Lifecycle Management, Observability, and Production Readiness


Goku2099:
# Tracking Issue: Spark Client Batch Jobs, Lifecycle Management, Observability, and Production Readiness

## Overview

This tracking issue tracks implementation work for Spark batch jobs, lifecycle management, observability, testing, documentation, and production readiness for the Spark Client.

Related:
- KEP-107: Spark Client

## Design & Proposal

- [x] SparkClient long-running job submission proposal (extends KEP-107) #524 

### Batch Job Submission

- [x] FileJob-based batch job submission, lifecycle management APIs, and unit tests (`submit_job()`, `list_jobs()`, `get_job()`, `wait_for_job_status()`, `delete_job()`, `get_job_logs()`) #521
- [x] FuncJob support (`submit_job(job=FuncJob(...))`) #629 
- [x] `spark_conf` support for SparkApplication jobs #646 
- [x] `options` support for SparkApplication jobs #685 

### Testing & Validation

- [x] Unit and e2e test coverage — included in each feature PR

### Documentation & Examples

- [x] SparkClient batch jobs documentation and examples #730
- [x] Kubeflow blog post: "Batch Jobs for SparkClient" https://github.com/kubeflow/blog/pull/207

Goku2099:
cc @tariq-hasan @Shekharrajak @andreyvelich

andreyvelich:
/area spark
@Goku2099 Please could you also update this KEP to support Spark Batch jobs: https://github.com/kubeflow/sdk/tree/main/docs/proposals/107-spark-client ?

RobuRishabh:
Filed #646  for `spark_conf` on batch `submit_job()`.

Issue #596, opened 2026-07-16
KEP-107: Consolidate Spark `connect` API


tariq-hasan:
### What you would like to be added?

The goal of the issue is to track the discussion for consolidation on the Spark `connect` API.

Possible discussion points include the introduction of typed connection modes (similar in principle to the typed submission modes - `FileJob` and `FuncJob` - on the `submit_job` API) and moving `Driver` and `Executor` dependencies to the `options` parameter.

### Why is this needed?

The [KEP](https://github.com/kubeflow/sdk/tree/main/proposals/107-spark-client) for Spark SDK provides two modes to leverage Spark on K8s - interactive workflows via the `connect` API and batch workflows via the `submit_job` API.

A seamless user journey on the Kubeflow SDK interface requires that the APIs be symmetric and consistent.

The consolidated form of the [`submit_job`](https://github.com/kubeflow/sdk/blob/main/proposals/107-spark-client/README.md?plain=1#L540-L547) API therefore motivates further simplification on the [`connect`](https://github.com/kubeflow/sdk/blob/main/proposals/107-spark-client/README.md?plain=1#L455-L467) API.

Related: https://github.com/kubeflow/sdk/issues/274

### Love this feature?

Give it a 👍 We prioritize the features with most 👍

PR #685, opened 2026-07-29
feat: [KEP-107] Support `options` in `submit_job()` for SparkApplication jobs

Goku2099:
## Summary

This PR adds support for configuring SparkApplication jobs through the `options` parameter in `SparkClient.submit_job()`.

The following SparkApplication configurations are now supported through `options`:

- Labels
- Annotations
- NodeSelector
- Toleration
- DriverOption
- ExecutorOption

## Additional Changes

- Added unit tests covering the new functionality.
- Added an end-to-end example demonstrating SparkApplication configuration through `options`.

ref - #520

### DIFF
--- a/examples/spark/batch_job_options.py
+++ b/examples/spark/batch_job_options.py
@@ -168,11 +168,31 @@ def example_verify_options():
     ):
         raise RuntimeError("Expected driver toleration to be applied.")
 
+    executor = spec.get("executor", {})
+
+    executor_node_selector = executor.get("nodeSelector", {})
+
+    if executor_node_selector.get("kubernetes.io/os") != "linux":
+        raise RuntimeError("Expected executor nodeSelector to be applied.")
+
+    executor_tolerations = executor.get("tolerations", [])
+
+    if not any(
+        t.get("key") == "dedicated"
+        and t.get("operator") == "Equal"
+        and t.get("value") == "spark"
+        and t.get("effect") == "NoSchedule"
+        for t in executor_tolerations
+    ):
+        raise RuntimeError("Expected executor toleration to be applied.")
+
     print("✓ Name verified.")
     print("✓ Labels verified.")
     print("✓ Annotations verified.")
-    print("✓ NodeSelector verified.")
-    print("✓ Toleration verified.")
+    print("✓ Driver NodeSelector verified.")
+    print("✓ Driver Toleration verified.")
+    print("✓ Executor NodeSelector verified.")
+    print("✓ Executor Toleration verified.")
 
     print("\nOptions verification complete.\n")
 
@@ -212,8 +232,8 @@ def example_delete_job():
 
 
 def main():
-    """Run the batch job option examples."""
-    print("E2E: Starting batch_job_option.py", flush=True)
+    """Run the batch job options examples."""
+    print("E2E: Starting batch_job_options.py", flush=True)
     print()
     print("=" * 70)
     print("KUBEFLOW SPARKCLIENT - BATCH JOB OPTIONS")
--- a/examples/spark/batch_job_options.py
+++ b/examples/spark/batch_job_options.py
@@ -1,0 +1,306 @@
+#!/usr/bin/env python3
+# Copyright The Kubeflow Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import os
+
+from kubeflow.common.types import KubernetesBackendConfig
+from kubeflow.spark import (
+    Annotations,
+    DriverOption,
+    ExecutorOption,
+    FileJob,
+    Labels,
+    Name,
+    NodeSelector,
+    SparkClient,
+    SparkJobStatus,
+    Toleration,
+)
+
+REMOTE_JOB = "https://raw.githubusercontent.com/kubeflow/sdk/main/examples/spark/spark_job.py"
+
+JOB_NAME: str | None = None
+
+
+def _backend_config(namespace_default: str = "default"):
+    """Backend config; uses SPARK_TEST_NAMESPACE in CI."""
+    return KubernetesBackendConfig(
+        namespace=os.environ.get("SPARK_TEST_NAMESPACE", namespace_default)
+    )
+
+
+def _client() -> SparkClient:
+    """Create SparkClient."""
+    return SparkClient(backend_config=_backend_config())
+
+
+def example_submit_and_wait():
+    """Submit a Spark batch job and wait for completion."""
+    global JOB_NAME
+
+    print("=" * 70)
+    print("SUBMIT SPARK BATCH JOB")
+    print("=" * 70)
+
+    client = _client()
+
+    print("\nSubmitting Spark job...")
+
+    JOB_NAME = client.submit_job(
+        job=FileJob(
+            file_source=REMOTE_JOB,
+            args=["10"],
+        ),
+        options=[
+            Name("batch-job-options-demo"),
+            Labels(
+                {
+                    "app": "spark",
+                    "team": "ml",
+                }
+            ),
+            Annotations(
+                {
+                    "owner": "kubeflow",
+                    "environment": "dev",
+                }
+            ),
+            NodeSelector(
+                {
+                    "kubernetes.io/os": "linux",
+                }
+            ),
+            Toleration(
+                key="dedicated",
+                operator="Equal",
+                value="spark",
+                effect="NoSchedule",
+            ),
+            DriverOption(
+                image="spark:4.0.1",
+                resources={
+                    "cpu": "2",
+                    "memory": "1Gi",
+                },
+                java_options="-XX:+UseG1GC",
+                service_account="spark-operator-spark",
+            ),
+            ExecutorOption(
+                num_instances=2,
+                resources_per_executor={
+                    "cpu": "1",
+                    "memory": "512Mi",
+                },
+                java_options="-XX:+UseG1GC",
+            ),
+        ],
+    )
+
+    print(f"Job submitted successfully: {JOB_NAME}")
+
+    print("\nWaiting for job to complete...")
+
+    job = client.wait_for_job_status(
+        JOB_NAME,
+        status={SparkJobStatus.COMPLETED},
+        timeout=300,
+    )
+
+    print("Job completed successfully.")
+    print(f"Status: {job.status}")
+    print(f"Driver Pod: {job.driver_pod_name}")
+    print(f"Namespace: {job.namespace}")
+    print("\nSubmit and wait example complete.\n")
+
+
+def example_get_job():
+    """Get information about a Spark batch job."""
+    print("=" * 70)
+    print("GET SPARK BATCH JOB")
+    print("=" * 70)
+
+    if JOB_NAME is None:
+        raise RuntimeError("No job has been submitted.")
+
+    client = _client()
+
+    print(f"\nRetrieving job: {JOB_NAME}")
+
+    job = client.get_job(JOB_NAME)
+
+    if job.status != SparkJobStatus.COMPLETED:
+        raise RuntimeError(f"Expected COMPLETED status, got {job.status}.")
+
+    if not job.driver_pod_name:
+        raise RuntimeError("Expected driver pod name to be populated.")
+
+    print("Job retrieved successfully.")
+    print(f"Name: {job.name}")
+    print(f"Namespace: {job.namespace}")
+    print(f"Status: {job.status}")
+    print(f"Driver Pod: {job.driver_pod_name}")
+    print(f"Executors: {job.num_executors}")
+
+    print("\nGet job example complete.\n")
+
+
+def example_list_jobs():
+    """List Spark batch jobs."""
+    print("=" * 70)
+    print("LIST SPARK BATCH JOBS")
+    print("=" * 70)
+
+    if JOB_NAME is None:
+        raise RuntimeError("No job has been submitted.")
+
+    client = _client()
+
+    print("\nListing Spark jobs...")
+
+    jobs = client.list_jobs()
+
+    print(f"Found {len(jobs)} Spark job(s).\n")
+
+    job_found = False
+
+    for job in jobs:
+        print(f"- {job.name} | Status: {job.status} | Namespace: {job.namespace}")
+
+        if job.name == JOB_NAME:
+            job_found = True
+
+    if not job_found:
+        raise RuntimeError(f"Submitted job '{JOB_NAME}' not found in job list.")
+
+    print("\nSubmitted job found in job list.")
+
+    print("\nListing completed Spark jobs...")
+
+    completed_jobs = client.list_jobs(
+        status={SparkJobStatus.COMPLETED},
+    )
+
+    print(f"Found {len(completed_jobs)} completed Spark job(s).\n")
+
+    completed_job_found = False
+
+    for job in completed_jobs:
+        if job.status != SparkJobStatus.COMPLETED:
+            raise RuntimeError(f"Expected COMPLETED status, got {job.status}.")
+
+        print(f"- {job.name} | Status: {job.status} | Namespace: {job.namespace}")
+
+        if job.name == JOB_NAME:
+            completed_job_found = True
+
+    if not completed_job_found:
+        raise RuntimeError(f"Submitted completed job '{JOB_NAME}' not found in filtered job list.")
+
+    print("\nCompleted job filter verified.")
+    print("\nList jobs example complete.\n")
+
+
+def example_get_job_logs():
+    """Get logs from a Spark batch job."""
+    print("=" * 70)
+    print("GET SPARK BATCH JOB LOGS")
+    print("=" * 70)
+
+    if JOB_NAME is None:
+        raise RuntimeError("No job has been submitted.")
+
+    client = _client()
+
+    print(f"\nRetrieving logs for: {JOB_NAME}")
+
+    print("\nDriver logs (first 20 lines):")
+    print("-" * 70)
+
+    line_count = 0
+    for line in client.get_job_logs(JOB_NAME):
+        print(line.rstrip())
+        line_count += 1
+
+        if line_count >= 20:
+            print("...")
+            break
+
+    print("-" * 70)
+    print(f"Displayed {line_count} log lines.")
+
+    print("\nGet job logs example complete.\n")
+
+
+def example_delete_job():
+    """Delete a Spark batch job."""
+    print("=" * 70)
+    print("DELETE SPARK BATCH JOB")
+    print("=" * 70)
+
+    global JOB_NAME
+
+    if JOB_NAME is None:
+        raise RuntimeError("No job has been submitted.")
+
+    client = _client()
+
+    print(f"\nDeleting job: {JOB_NAME}")
+
+    client.delete_job(JOB_NAME)
+
+    print("Job deleted.")
+
+    try:
+        client.get_job(JOB_NAME)
+    except RuntimeError as e:
+        if "Spark job not found" not in str(e):
+            raise
+    else:
+        raise RuntimeError("Job still exists after deletion.")
+
+    print("Verified job has been deleted.")
+
+    JOB_NAME = None
+
+    print("\nDelete job example complete.\n")
+
+
+def main():
+    """Run all batch job lifecycle examples."""
+    print("E2E: Starting batch_job_lifecycle.py", flush=True)
+    print()
+    print("=" * 70)
+    print("KUBEFLOW SPARKCLIENT - BATCH JOB LIFECYCLE")
+    print("=" * 70)
+
+    try:
+        example_submit_and_wait()
+        example_get_job()
+        example_list_jobs()
+        example_get_job_logs()
+        example_delete_job()
+
+        print("=" * 70)
+        print("BATCH JOB LIFECYCLE COMPLETE!")
+        print("=" * 70)
+
+    except Exception as e:
+            print(f"\nError: {e}")
+            raise SystemExit(1) from e
+
+
+if __name__ == "__main__":
+    main()
--- a/examples/spark/batch_job_options.py
+++ b/examples/spark/batch_job_options.py
@@ -15,12 +15,11 @@
 
 
 import os
+import uuid
 
 from kubeflow.common.types import KubernetesBackendConfig
 from kubeflow.spark import (
     Annotations,
-    DriverOption,
-    ExecutorOption,
     FileJob,
     Labels,
     Name,
@@ -65,7 +64,7 @@ def example_submit_and_wait():
             args=["10"],
         ),
         options=[
-            Name("batch-job-options-demo"),
+            Name(f"batch-job-options-{uuid.uuid4().hex[:8]}"),
             Labels(
                 {
                     "app": "spark",
@@ -89,23 +88,6 @@ def example_submit_and_wait():
                 value="spark",
                 effect="NoSchedule",
             ),
-            DriverOption(
-                image="spark:4.0.1",
-                resources={
-                    "cpu": "2",
-                    "memory": "1Gi",
-                },
-                java_options="-XX:+UseG1GC",
-                service_account="spark-operator-spark",
-            ),
-            ExecutorOption(
-                num_instances=2,
-                resources_per_executor={
-                    "cpu": "1",
-                    "memory": "512Mi",
-                },
-                java_options="-XX:+UseG1GC",
-            ),
         ],
     )
 
@@ -126,41 +108,10 @@ def example_submit_and_wait():
     print("\nSubmit and wait example complete.\n")
 
 
-def example_get_job():
-    """Get information about a Spark batch job."""
+def example_verify_options():
+    """Verify that Spark options were applied to the SparkApplication."""
     print("=" * 70)
-    print("GET SPARK BATCH JOB")
-    print("=" * 70)
-
-    if JOB_NAME is None:
-        raise RuntimeError("No job has been submitted.")
-
-    client = _client()
-
-    print(f"\nRetrieving job: {JOB_NAME}")
-
-    job = client.get_job(JOB_NAME)
-
-    if job.status != SparkJobStatus.COMPLETED:
-        raise RuntimeError(f"Expected COMPLETED status, got {job.status}.")
-
-    if not job.driver_pod_name:
-        raise RuntimeError("Expected driver pod name to be populated.")
-
-    print("Job retrieved successfully.")
-    print(f"Name: {job.name}")
-    print(f"Namespace: {job.namespace}")
-    print(f"Status: {job.status}")
-    print(f"Driver Pod: {job.driver_pod_name}")
-    print(f"Executors: {job.num_executors}")
-
-    print("\nGet job example complete.\n")
-
-
-def example_list_jobs():
-    """List Spark batch jobs."""
-    print("=" * 70)
-    print("LIST SPARK BATCH JOBS")
+    print("VERIFY SPARK BATCH JOB OPTIONS")
     print("=" * 70)
 
     if JOB_NAME is None:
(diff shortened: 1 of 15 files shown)


Goku2099:
cc @tariq-hasan @Shekharrajak @andreyvelich

kubeflow/spark/types/options.py
```diff
@@ -408,4 +519,119 @@ def __call__(
...
+        memory = _convert_kubernetes_memory_to_spark(
+            resources.get("memory", DEFAULT_EXECUTOR_MEMORY)
+        )
+
+        resource.spec.executor.instances = instances
+        resource.spec.executor.cores = cores
+        resource.spec.executor.memory = memory
+
+        if self.java_options is not None:
+            resource.spec.executor.java_options = self.java_options
```
tariq-hasan: I don't think it is a good idea at this time to expose `driver` and `executor` through `options`.

For `SparkConnect`, `driver` and `executor` are defined in the `types` module already and exposed via the `connect` API. Exposing the same semantics on the `options` module would create redundancy at this stage and make the implementation somewhat complex.

There is definitely a tension currently in the fact that `driver` and `executor` are supported in the `connect` API and defined in the `types` module while `driver` and `executor` are not supported in the `submit_job` API.

I think we should take a phased approach to resolving this problem with two separate steps - firstly, [consolidating](https://github.com/kubeflow/sdk/issues/596) the `connect` API by removing direct `driver` and `executor` support from there (as part of some future iteration), and secondly, moving `driver` and `executor` from `types.py` to `options.py`. That way we would not have redundancy and at the same time ensure eventual `driver` and `executor` support through `options`.

The trade-off would be to not have advanced `driver` and `executor` support currently for job submission but I am thinking for an initial iteration the simpler approach would be ideal.

cc @andreyvelich @Shekharrajak Would be interested to understand your thoughts on the issue as well.

kubeflow/spark/types/options.py
```diff
@@ -21,11 +21,86 @@
...
+        return memory
+
+    exponent = exponent_by_suffix[suffix]
+
+    spark_suffix = {10: "k", 20: "m", 30: "g", 40: "t", 50: "p"}.get(exponent)
+    if "." not in coefficient and spark_suffix is not None:
+        return coefficient + spark_suffix
+
+    total_bytes = math.ceil(float(coefficient) * (2**exponent))
+    return f"{math.ceil(total_bytes / (2**20))}m"
```
tariq-hasan: I'm wondering if we can remove the duplicate helper by moving `from kubeflow.spark.types.options import Name` within the `extract_name_option` function in `utils.py`.

kubeflow/spark/types/options.py
```diff
@@ -162,29 +233,35 @@ class PodTemplateOverride:
...
+            raise ValueError("PodTemplateOverride is currently supported only for SparkConnect.")
+
         if self.role == "driver":
-            role_spec = spark_connect.spec.server
+            role_spec = resource.spec.server
         elif self.role == "executor":
-            role_spec = spark_connect.spec.executor
+            role_spec = resource.spec.executor
         else:
             raise ValueError(f"Invalid role '{self.role}'. Must be 'driver' or 'executor'.")
```
tariq-hasan, line 201: We'd want to support `PodTemplateOverride` for `SparkJob` as well.

kubeflow/spark/types/options.py
```diff
@@ -335,20 +428,40 @@ def __call__(
...
+            role_specs = [
+                resource.spec.driver,
+                resource.spec.executor,
+            ]
+
+            for role_spec in role_specs:
+                if role_spec.tolerations is None:
+                    role_spec.tolerations = []
+
+                role_spec.tolerations.append(toleration)
```
tariq-hasan: I think we should do an explicit `elif isinstance(resource, models.SparkV1alpha1SparkConnect)` with a terminal `raise` to make the support explicit.

Same for `NodeSelector`.

kubeflow/spark/types/options.py
```diff
@@ -335,20 +428,40 @@ def __call__(
...
 
 
 @dataclass
 class Name:
-    """Set a custom name for the SparkConnect session.
+    """Set a custom name for the Spark resource.
 
     This option sets the session name which becomes the Kubernetes resource name.
     If not provided, a name will be auto-generated with format: spark-connect-{uuid}
```
tariq-hasan: Let update `spark-connect-{uuid}` in the docstring since we support two formats.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -715,6 +774,7 @@ def get_spark_application_cr_from_file_job(
         arguments: Command-line arguments passed to the Spark application.
         num_executors: Number of executor instances.
         resources_per_executor: Resource requirements for each executor.
+        options: List of configuration options.
```
tariq-hasan, line 761: We can add `backend` param in the `Args` list as well.

Same for `get_spark_application_cr_from_func_job` below.

tariq-hasan (commented):
Hi @Goku2099! Thanks for raising the PR! I have added a few comments.

kubeflow/spark/types/options.py
Goku2099: Thanks, Tariq. I completely agree with this approach. It makes sense to avoid introducing duplicate semantics at this stage and keep the implementation simpler. I'll remove the DriverOption and ExecutorOption support from this PR for now and keep the options support focused on the remaining configurations. We can revisit driver and executor support once we consolidate the API in the future.

kubeflow/spark/types/options.py
Goku2099: yup, good call! I've moved the Name import inside extract_name_option() to avoid the circular dependency and removed the duplicate helper.

examples/spark/batch_job_options.py
```diff
@@ -0,0 +1,292 @@
...
+    try:
+        example_submit_and_wait()
+        example_get_job()
+        example_list_jobs()
+        example_get_job_logs()
+        example_delete_job()
+
+        print("=" * 70)
+        print("BATCH JOB LIFECYCLE COMPLETE!")
+        print("=" * 70)
```
andreyvelich: The new e2e test asserts `"BATCH JOB OPTIONS COMPLETE!"` but this prints `"BATCH JOB LIFECYCLE COMPLETE!"`, so `test_batch_job_options_example` fails. The startup line also names the wrong file.

```suggestion
    """Run the batch job options examples."""
    print("E2E: Starting batch_job_options.py", flush=True)
    print()
    print("=" * 70)
    print("KUBEFLOW SPARKCLIENT - BATCH JOB OPTIONS")
    print("=" * 70)

    try:
        example_submit_and_wait()
        example_get_job()
        example_list_jobs()
        example_get_job_logs()
        example_delete_job()

        print("=" * 70)
        print("BATCH JOB OPTIONS COMPLETE!")
        print("=" * 70)
```

andreyvelich (commented):
Thanks a lot for working on this, @Goku2099! 🙏

This review was performed with the help of AI tools. Please see the inline comments.

kubeflow/spark/types/options.py
```diff
@@ -161,60 +159,67 @@ class PodTemplateOverride:
     role: str  # "driver" or "executor"
     template: dict[str, Any]
 
-    def __call__(
-        self, spark_connect: models.SparkV1alpha1SparkConnect, backend: RuntimeBackend
+    def apply_to_manifest(
```
andreyvelich: **We don't need `PodTemplateOverride` in the batch job option surface.** Let's define options the way users actually want to express intent — `Name`, `Labels`, `Annotations`, `NodeSelector`, `Toleration` — which is also how Trainer options are shaped. When users need pod-level control there, they get typed dataclasses with named fields (`PodSpecPatch.node_selector`, `.tolerations`, `.affinity`, `.security_context` in `kubeflow/trainer/options/kubernetes.py`), never a raw `dict[str, Any]` manifest blob.

The raw-dict escape hatch is what forces all of the extra machinery in this PR: `apply_to_manifest`, `apply_manifest_overrides`, `normalize_pod_templates`, and `spark_application_from_response` exist only to serve it. Dropping it from `submit_job` lets every option apply in a single uniform pass over the resource, which is exactly the Trainer shape.

It also can't be validated well. Verified on this branch: `PodTemplateOverride(role="anything-else", ...)` silently patches the **executor**, because the `else` branch below swallows any non-`"driver"` value — and the `test_pod_template_invalid_role` case that used to catch this was deleted in this PR.

Since the option already exists for `connect()`, no need to remove it there in this PR — please just leave it out of the `submit_job` path and drop the manifest-override layer. If we do want raw pod-template control for Spark later, let's design it as a typed patch like Trainer's.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -292,6 +292,143 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+
+        options:
+            Configuration options supplied by the user.
+    """
+    if not options:
+        return
+
+    from kubeflow.spark.types.options import PodTemplateOverride
+
+    for option in options:
```
andreyvelich, line 325: Following on from the `PodTemplateOverride` comment: this `isinstance` dispatch is the part that diverges most from Trainer. Trainer applies every option in **one** loop (`trainer/backends/kubernetes/backend.py:313`) through a single uniform `__call__`, so a new option needs no backend changes at all. Here an option is applied from two places depending on its class, so adding one means editing this function — which defeats the *"extensibility without API changes (callable pattern)"* goal stated in `build_spark_connect_cr`.

With `PodTemplateOverride` out of `submit_job`, `apply_manifest_overrides` / `normalize_pod_templates` / `spark_application_from_response` can all go away and `apply_options` becomes the single application point.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -292,6 +292,143 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+        return default_name, []
+
+    name = default_name
+    filtered_options = []
+
+    for option in options:
+        if isinstance(option, Name):
+            name = option.name
+        else:
+            filtered_options.append(option)
```
andreyvelich: Filtering `Name` out of the list means `Name.__call__` never runs, so its `KubernetesBackend` check never fires — `Name` silently accepts unsupported backends, unlike every sibling option. Trainer avoids this by keeping `Name` in the same pass: it writes `metadata["name"]` and the backend reads it back out (`trainer/options/common.py:47`, `trainer/backends/kubernetes/backend.py:317`).

Worth aligning on that so `Name` isn't a special case with a dead code path.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -292,6 +292,143 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+            Spark resource to configure.
+
+        options:
+            List of configuration options.
+
+        backend:
+            Backend used for option validation.
+    """
+    if not options or backend is None:
+        return
```
andreyvelich: `backend` defaults to `None` on both `get_spark_application_cr_from_*_job`, so a caller that forgets it gets **all options silently dropped**. Verified: `get_spark_application_cr_from_file_job(..., options=[Labels({"a": "b"})])` returns a CR with `metadata.labels is None` and no error. Fail loudly instead:

```suggestion
    if not options:
        return

    if backend is None:
        raise ValueError("A backend instance is required to apply Spark options.")
```

kubeflow/spark/types/options_test.py
```diff
@@ -29,6 +29,7 @@
     PodTemplateOverride,
     Toleration,
 )
+from kubeflow.trainer.test.common import FAILED, SUCCESS, TestCase
```
andreyvelich: Spark tests shouldn't import trainer test helpers — `kubeflow/spark/test/common.py` has an identical `TestCase` (including `expected_error`) and is what every other spark test uses.

```suggestion
from kubeflow.spark.test.common import FAILED, SUCCESS, TestCase
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -292,6 +292,143 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+
+    The Kubernetes API may serialize empty PodTemplateSpec fields as
+    ``containers: null``. The generated SDK models expect ``containers`` to
+    always be a list, so normalize null values to an empty list before
+    deserialization.
+    """
+
+    spec = body.get("spec", {})
+
+    for role in ("driver", "executor", "server"):
```
andreyvelich: If this normalization survives: `"server"` is only reachable for SparkConnect, but the SparkConnect paths (`backend.py:205`, `234`, `283`) still call `SparkV1alpha1SparkConnect.from_dict(response)` without normalizing — only the SparkApplication paths go through `spark_application_from_response`. Either wire it in there too or drop `"server"`.

kubeflow/spark/backends/base.py
```diff
@@ -149,6 +149,7 @@ def submit_job(
         job: FileJob | FuncJob,
         num_executors: int | None = None,
         resources_per_executor: dict[str, str] | None = None,
+        options: list | None = None,
```
andreyvelich, line 152: The `Args:` block below still omits `options`. Please document it here too, matching the Kubernetes backend override.

examples/spark/batch_job_options.py
```diff
@@ -0,0 +1,292 @@
...
+
+    print("\nSubmitting Spark job...")
+
+    JOB_NAME = client.submit_job(
+        job=FileJob(
+            file_source=REMOTE_JOB,
+            args=["10"],
+        ),
+        options=[
+            Name("batch-job-options-demo"),
```
andreyvelich: A hard-coded name makes the e2e test non-idempotent: if a run dies before `example_delete_job`, every later run fails with `AlreadyExists`. Consider omitting `Name` here, or suffixing it from an env var / the namespace.

Two related notes on this example:
- It never verifies the options actually landed on the CR (labels, annotations, nodeSelector, tolerations) — that's the point of an options example, and `example_get_job` only checks status and driver pod name.
- The `get`/`list`/`logs`/`delete` sections are a verbatim copy of `batch_job_lifecycle.py`; trimming to submit + verify would keep it focused.

kubeflow/spark/types/options.py
andreyvelich: @Goku2099 Let's ensure we define options that are specific to SparkApplication or SparkConnect API.
PodTemplateOverrides don't exist there.

kubeflow/spark/types/options_test.py
andreyvelich: Can we move them to `kubeflow.common` ?

kubeflow/spark/types/options.py
Goku2099: Thanks for the detailed feedback, Andrey. That makes sense. I'll remove PodTemplateOverride from the batch job path, keep the existing connect() behavior, and align the Spark options with the typed option pattern used in Trainer.

kubeflow/spark/types/options_test.py
Goku2099: @andreyvelich Good point. Since this involves reorganizing shared test utilities across multiple components, I'd prefer to keep it separate from this feature PR. I can open a follow-up issue to move the common test helpers into a shared module and update the imports accordingly.

Goku2099:
@andreyvelich - I removed the pod template override support and simplified the implementation accordingly. I also updated the batch job options example to keep it focused on demonstrating the options API: it now uses a unique job name for idempotent execution and verifies that the configured Name, Labels, Annotations, NodeSelector, and Toleration are applied to the created SparkApplication.

examples/spark/batch_job_options.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
```
andreyvelich, line 1: Since we decided to convert all Spark examples to Jupyter Notebooks, do we need this?

kubeflow/spark/api/spark_client.py
```diff
@@ -218,13 +218,11 @@ def submit_job(
...
             raise NotImplementedError("spark_conf support is not yet implemented.")
 
-        if options is not None:
-            raise NotImplementedError("options are not supported in Phase 1.")
-
         return self.backend.submit_job(
             job=job,
             num_executors=num_executors,
             resources_per_executor=resources_per_executor,
+            options=options,
```
andreyvelich, line 225: Can you also update the docs on how to use options w/ Spark.
https://sdk.kubeflow.org/en/latest/spark/index.html

examples/spark/batch_job_options.py
Goku2099: Good point. My understanding is that we're planning to migrate the Spark examples to Jupyter notebooks, but that migration hasn't happened yet. I added this Python example to document and exercise the new options in the meantime. Once the notebook migration happens, I'm happy to move this example there and remove the .py version.

kubeflow/spark/api/spark_client.py
Goku2099: I agree the Spark documentation should cover the options API. Since that would be a broader documentation update beyond the new submit_job support introduced here, would you prefer that in a follow-up PR, or would you like me to include a concise usage example in this PR?

andreyvelich: Do we have a tracking issue for that?

Goku2099: Yes, we already have a tracking issue for the follow-up documentation work: #520. It includes the Spark batch job documentation and examples. I can add the Spark options documentation under that tracking issue and link the follow-up PR to it.

kubeflow/spark/types/options_test.py
Goku2099: @andreyvelich I've created a tracking issue for consolidating the shared test helpers into kubeflow.common: #700 . Since this involves shared infrastructure across multiple SDK components, I'd prefer to keep it separate from this feature PR.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -292,6 +292,34 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+    """
+    if not options:
+        return
+
+    if backend is None:
+        raise ValueError("A backend instance is required to apply Spark options.")
+
+    for option in options:
+        if callable(option):
+            option(resource, backend)
```
andreyvelich: Non-callable entries are silently dropped. Verified on this branch: `options=[{"labels": {"a": "b"}}, "team=ml"]` returns a CR with `metadata.labels is None` and no error — before this PR that case raised `NotImplementedError` at the client. Fail loudly, and document both raises:

```suggestion
        backend:
            Backend used for option validation.

    Raises:
        ValueError:
            If options are provided without a backend instance.

        TypeError:
            If an option is not callable.
    """
    if not options:
        return

    if backend is None:
        raise ValueError("A backend instance is required to apply Spark options.")

    for option in options:
        if not callable(option):
            raise TypeError(
                f"Invalid Spark option: {option!r}. Options must be callable, for example "
                f"Name, Labels, Annotations, NodeSelector or Toleration."
            )

        option(resource, backend)
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -960,6 +933,8 @@ def submit_job(
                 arguments=job.args,
                 num_executors=num_executors,
                 resources_per_executor=resources_per_executor,
+                options=options,
+                backend=self,
```
andreyvelich, line 932: `Name` now overwrites `metadata.name` inside `apply_options`, but `job_name` (line 921) still holds the auto-generated name. So the log at line 923 and the `TimeoutError`/`RuntimeError` at lines 964/968 report `spark-job-<uuid>` — a resource that never existed — whenever a user passes `Name(...)`. `_extract_name_option` used to resolve the name *before* the CR was built.

Please read the effective name back off the built CR before logging:

```python
# The Name option may override the auto-generated name.
job_name = spark_application.metadata.name
```

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -187,7 +156,7 @@ def _create_session(
             spark_conf=spark_conf,
             driver=driver,
             executor=executor,
-            options=filtered_options,  # Use filtered list
+            options=options,  # Use filtered list
```
andreyvelich: No filtering happens any more — `_extract_name_option` was deleted in this PR.

```suggestion
            options=options,
```

`_create_session` has the same stale-name problem as `submit_job`: `name` (line 149) is the generated name, so the log at line 163 and the errors at lines 177/181 print a name that does not exist once `Name` is used.

kubeflow/spark/options/kubernetes.py
```diff
@@ -162,29 +160,35 @@ class PodTemplateOverride:
...
         if not isinstance(backend, KubernetesBackend):
             raise ValueError(
-                f"PodTemplateOverride option is not compatible with {type(backend).__name__}. "
-                f"Supported backends: KubernetesBackend"
+                f"PodTemplateOverride option is not compatible with "
+                f"{type(backend).__name__}. Supported backends: KubernetesBackend"
             )
 
+        if not isinstance(resource, models.SparkV1alpha1SparkConnect):
+            raise ValueError("PodTemplateOverride is currently supported only for SparkConnect.")
```
andreyvelich, line 193: `coverage report -m` on this branch shows the four guards added across the previous review rounds are all uncovered: `options.py:186`, `options.py:297`, `options.py:390` and `utils.py:316`. Please add `TestCase`s — `options_test.py` already has the `spark_application_model` + `mock_k8s_backend` fixtures and the `pytest.raises(match=...)` pattern.

Separately: the `PodTemplateOverride` class docstring (line 131) only says `Supported backends: Kubernetes`. Worth stating it is SparkConnect-only, since the class is exported from `kubeflow.spark` alongside the options that now do work with `submit_job()`.

kubeflow/spark/options/kubernetes.py
```diff
@@ -262,15 +264,37 @@ def __call__(
...
+                resource.spec.executor,
+            ]
+
+            for role_spec in role_specs:
+                if role_spec.node_selector is None:
+                    role_spec.node_selector = {}
+
+                role_spec.node_selector.update(self.selectors)
+        else:
+            raise TypeError(f"Unsupported Spark resource type: {type(resource).__name__}")
```
andreyvelich, line 306: `NodeSelector.__call__`'s `Raises:` block (line 255) lists only `ValueError`. Please add `TypeError` there — and the same for `Toleration.__call__` (raise at line 390, docstring at line 340).

kubeflow/spark/backends/base.py
```diff
@@ -149,13 +149,15 @@ def submit_job(
...
     ) -> SparkJob:
         """Submit a Spark batch job.
 
         Args:
             job: Spark application to execute.
             num_executors: Number of executor instances.
             resources_per_executor: Resource requirements for each executor.
+             options: List of additional Spark configuration options.
```
andreyvelich: 13-space indent vs 12 on the sibling entries, which breaks the Google-style Args block:

```suggestion
            options: List of configuration options. Use the Name option for a custom job name.
```

kubeflow/spark/backends/kubernetes/backend_test.py
```diff
@@ -1277,6 +1224,19 @@ def test_validate_job(kubernetes_backend, test_case):
                 "use_mock_command": True,
             },
         ),
+        TestCase(
+            name="valid remote file submission with options",
```
andreyvelich, line 1228: This case has no `use_mock_command`, so it falls into the `else` branch at lines 1273-1276, which calls `submit_job(job=...)` **without** `options` — the `Name`/`Labels` here are never passed and `apply_options` never runs. The FileJob + options path has no backend-level coverage.

Note that forwarding `options` in the `else` branch will also break `assert job.name.startswith("spark-job-")` at line 1279 (the fake API echoes the request body, so the name becomes `custom-job`), so that assertion needs to branch on the expected name.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -87,6 +90,33 @@ def spark_application_spec():
...
+@pytest.fixture
+def mock_k8s_backend():
+    """Create a mock KubernetesBackend."""
+    backend = Mock(spec=KubernetesBackend)
+    backend.__class__ = KubernetesBackend
+    return backend
+
+
+@pytest.fixture
+def mock_non_k8s_backend():
```
andreyvelich: This fixture is never requested by any test in this file — `options_test.py` defines its own. Please remove it.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1268,10 +1388,29 @@ def test_get_spark_application_cr_from_file_job(test_case: TestCase) -> None:
...
             },
         ),
+        TestCase(
+            name="build spark application from function with options",
+            expected_status=SUCCESS,
+            config={
+                "name": "test-job",
+                "namespace": "default",
+                "func": sample_function,
+                "func_args": [1, 2],
```
andreyvelich: `func_args` is `dict[str, Any] | None`; a list here generates `sample_function(**[1, 2])`. The sibling case at line 1209 uses a dict.

```suggestion
                "func_args": {"a": 1, "b": 2},
```

examples/spark/batch_job_options.py
```diff
@@ -0,0 +1,237 @@
...
+    print("Verified job has been deleted.")
+
+    JOB_NAME = None
+
+    print("\nDelete job example complete.\n")
+
+
+def main():
+    """Run the batch job option examples."""
+    print("E2E: Starting batch_job_option.py", flush=True)
```
andreyvelich: Filename is `batch_job_options.py` — this marker is what locates the example in the in-cluster e2e logs:

```suggestion
    """Run the batch job options examples."""
    print("E2E: Starting batch_job_options.py", flush=True)
```

examples/spark/batch_job_options.py
```diff
@@ -0,0 +1,237 @@
...
+    if labels.get("team") != "ml":
+        raise RuntimeError("Expected label team=ml.")
+
+    if annotations.get("owner") != "kubeflow":
+        raise RuntimeError("Expected annotation owner=kubeflow.")
+
+    if annotations.get("environment") != "dev":
+        raise RuntimeError("Expected annotation environment=dev.")
+
+    driver = spec.get("driver", {})
```
andreyvelich, line 153: `NodeSelector` and `Toleration` apply to driver **and** executor, but this only checks `spec.driver`. A regression that dropped the executor half would still pass the e2e — please add the equivalent `spec.executor` assertions.

Goku2099:
Hi @andreyvelich , I've addressed all of the review comments, Could you please take another look when you have a chance? Thanks!

kubeflow/spark/options/kubernetes.py
```diff
@@ -27,6 +27,8 @@
 
 from kubeflow.spark.backends.base import RuntimeBackend
```
andreyvelich, line 28: Similar to Trainer, can you move the options file to `kubeflow/spark/options/kubernetes.py`
Each backend can have its own options.
https://github.com/Goku2099/sdk/tree/8b34fc9e294b232099c2d1ae0c8fe3e959926220/kubeflow/trainer/options

kubeflow/spark/options/kubernetes.py
```diff
@@ -262,15 +273,37 @@ def __call__(
...
-                # PodSpec requires containers field (can be empty list)
-                role_spec.template.spec = models.IoK8sApiCoreV1PodSpec(containers=[])
-            if role_spec.template.spec.node_selector is None:
-                role_spec.template.spec.node_selector = {}
-            role_spec.template.spec.node_selector.update(self.selectors)
+        if isinstance(resource, models.SparkV1alpha1SparkConnect):
+            role_specs = [
+                resource.spec.server,
+                resource.spec.executor,
+            ]
```
andreyvelich, line 280: @abhijeet-dhumal @kramaranya @Fiona-Waters @robert-bell @szaher @tariq-hasan  What are you thoughts on this options implementations for SparkClient?

Do we want to allow users to use the same options for `submit_job()`: SparkApplication, and `connect()`: SparkConnect resources?

andreyvelich (commented):
Thanks @Goku2099!
I am fine to move this forward, so we can remove the PodTemplateOverrides from options in the followup PR: https://github.com/kubeflow/sdk/issues/720

/lgtm
/assign @tariq-hasan @Shekharrajak @RobuRishabh

tariq-hasan:
Thanks @Goku2099!
/lgtm

kubeflow/spark/options/kubernetes.py
tariq-hasan: From my side I'm thinking the basic options we provide for SparkClient share API semantics that are common enough to justify using the same options for both resources - even though the orchestration mechanism on the controller side may be different. Being uniform would also lead to a clean, consistent experience and less burden as well for having to maintain a dual set of options and possibly duplicated logic.

abhijeet-dhumal: +1 with @tariq-hasan .
Would it make sense to keep one shared options surface for both submit_job() and connect() for the common ones (Labels, Annotations, Name, NodeSelector, Toleration), and treat PodTemplateOverride as the Connect-only exception for now with the clearer split tracked in #720?

kubeflow/spark/options/kubernetes.py
```diff
@@ -118,21 +116,28 @@ def __call__(
...
-        spark_connect.metadata.annotations.update(self.annotations)
+        if resource.metadata.annotations is None:
+            resource.metadata.annotations = {}
+        resource.metadata.annotations.update(self.annotations)
 
 
 @dataclass
 class PodTemplateOverride:
-    """Override pod template specifications for driver or executors.
+    """Override pod template specifications for Spark Connect driver or executors.
```
abhijeet-dhumal, line 126: Trainer moved off PodTemplateOverrides to RuntimePatch.. Spark doesn’t have an equivalent API yet. Are we planning this to keep this Connect-only for now?

kubeflow/spark/options/kubernetes.py
abhijeet-dhumal: btw for deeper pod customization.. Trainer moved off PodTemplateOverrides to RuntimePatch, but Spark doesn’t have an equivalent API yet. Should we keep Spark’s PodTemplateOverride as Connect-only for now, or are we planning something closer to RuntimePatch for Spark later?

kubeflow/spark/options/kubernetes.py
Goku2099: we are planning to remove the PodTemplateOverrides from options in the followup PR #720

kubeflow/spark/options/kubernetes.py
robert-bell: @andreyvelich are you thinking whether we use the same options-object for both SparkApplication and SparkConnect?

If so, yes I agree with @tariq-hasan and @abhijeet-dhumal -- use the same set of option objects for both types. The type checking if/else adds pain for maintenance, but IMO that's a fine trade-off if it gives a much simpler end-user experience.

Are there any options that would apply to only one of SparkApplication or SparkConnect? That would complicate things, though could probably be fixed with type hints and a `typing.overload` on `submit_job()`.

andreyvelich: Yes, using the same options give users nice experience between `connect()` and `submit_job()` APIs.
We should remove PodTemplateOverrides from Spark options as part of this: https://github.com/kubeflow/sdk/issues/720

PR #692, opened 2026-07-30
feat(spark): add get_job_metrics to SparkClient

### LINKED ISSUE #689: Add Spark metrics collection via Spark REST API

Right now `SparkClient` can tell you the status of a job (`get_job`, `wait_for_job_status`) but nothing about what's actually going on inside it while it runs, no executor count, no task progress, nothing. If you want that today you're stuck going through kubectl or opening the Spark UI yourself.

I'd like to add a `get_job_metrics(name)` method that pulls basic runtime metrics straight from Spark's own REST API on the driver:

- `GET /api/v1/applications/<app-id>/executors` for executor count, active/completed tasks, memory used
- `GET /api/v1/applications/<app-id>/stages` for stage progress

Went through the actual Spark Operator source (`web_ui.go`, `pkg/util/sparkapplication.go`) before settling on an approach, a couple of things worth calling out:

The operator creates a dedicated Kubernetes Service for the driver UI on every SparkApplication, default port name `spark-driver-ui-port`, default port `4040` (driven by `spark.ui.port`, but can be overridden via `SparkUIOptions` on the SparkApplication spec). So the port should be read off the live Service object, not assumed.

The service name itself isn't safe to construct by hand either. The operator builds it as `<app-name>-ui-svc` only when that's 63 characters or under (Kubernetes' DNS label limit), otherwise it truncates the app name and appends an 8 character hash. What is reliable is the label the operator puts on every UI service it creates, `sparkoperator.k8s.io/app-name: <app-name>`. So the plan is to look the service up by that label selector instead of guessing the name.

Also, this UI service isn't guaranteed to exist at all. It's gated behind an operator-level flag (`--enable-ui-service`, defaults to true) that whoever deployed the Spark Operator in a given cluster could turn off. So `get_job_metrics` needs to treat "no UI service for this job" as a normal, expected case, not just something that happens after the job finishes and gets cleaned up.

To reach the service:
- In-cluster: hit it directly through the Service DNS name.
- Outside cluster (local dev, tests): port-forward to it. Same port-forward mechanism the Kubernetes backend already uses for Spark Connect (`get_connect_url` in `backend.py`, around lines 435-511), just pointed at this service instead.

Rough shape of what I'm thinking:

```python
@dataclass
class SparkJobMetrics:
    num_executors: int | None = None
    active_tasks: int | None = None
    completed_tasks: int | None = None
    failed_tasks: int | None = None
    active_stages: int | None = None
    completed_stages: int | None = None
```

`get_job_metrics` should return a mostly-empty `SparkJobMetrics` rather than raising whenever the UI service can't be found, whether that's because it was disabled at the operator level or because it's already been cleaned up after the job finished.

### Why is this needed?

This is part of the observability gap called out in #655, batch jobs work fine but there's basically no visibility into them once they're running. Metrics collection is one of the pieces on that epic that nobody's picked up yet.

### Love this feature?

Sharkyii:

`SparkClient.get_job` and `get_job_logs` cover status and logs for a batch job, but there's no visibility into what's actually happening while it runs, no executor count, no task progress, nothing. This adds `get_job_metrics(name)`, which pulls basic runtime metrics from Spark's own REST API on the driver's web UI (executor counts, task counts, stage progress).

The Spark Operator creates a Service for the driver web UI on every SparkApplication. Rather than constructing that name (`<app-name>-ui-svc`), which the operator itself truncates and hashes past the 63 character DNS label limit, this looks the service up by the `sparkoperator.k8s.io/app-name` label the operator sets on it, and reuses the existing port-forward pattern from `get_connect_url` for out-of-cluster access.

Returns an empty `SparkJobMetrics` instead of raising when metrics aren't available, covering three real cases: the driver hasn't started yet, the SparkApplication has already been cleaned up, or the cluster's Spark Operator was deployed with `--enable-ui-service=false` (defaults to true, but it's an operator-level flag, so it can't be assumed present).

Testing done:
- Unit tests: no app id yet, no UI service found, successful in-cluster fetch, REST API unreachable
- `make verify` (ruff check + format) clean
- `uv run pytest kubeflow/spark/` passes (268 tests)

Fixes #689

**Checklist:**


kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -1261,3 +1263,207 @@ def _stream() -> Iterator[str]:
                 ) from e
 
         return _stream()
+
+    def get_job_metrics(self, name: str) -> SparkJobMetrics:
+        """Get runtime metrics for a Spark job from the Spark REST API.
+
+        Args:
+            name: Name of the SparkApplication.
```
Shekharrajak: name or ID ?

kubeflow/spark/api/spark_client_test.py
```diff
@@ -171,3 +172,18 @@ def test_submit_job_success(job):
...
+
+    with patch("kubeflow.spark.api.spark_client.KubernetesBackend") as mock_backend:
+        backend = mock_backend.return_value
+        backend.get_job_metrics.return_value = SparkJobMetrics(num_executors=2)
+
+        client = SparkClient()
+
+        metrics = client.get_job_metrics("spark-job-123")
+
+        assert metrics == SparkJobMetrics(num_executors=2)
```
Shekharrajak, line 190: please share examples and running spark metrics as unit tests

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -1261,3 +1263,207 @@ def _stream() -> Iterator[str]:
...
+            )
+            stages_resp.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            logger.warning("Failed to fetch Spark metrics for %s/%s: %s", self.namespace, name, e)
+            return SparkJobMetrics()
+
+        executors = executors_resp.json()
+        stages = stages_resp.json()
+
+        return SparkJobMetrics(
```
Shekharrajak, line 1457: For basic metrics this is fine but I think there would be huge number of metrics that we can pull so we need to brainstorm how we can make it generic.

You can check how spark job metrics are used by data engineers using grafana/observability dashboard.

Shekharrajak (commented):
Check how Spark UI/ History server url can be shared to users to go through the job failure/success, logs, executors config, etc.

kubeflow/spark/backends/kubernetes/backend.py
Sharkyii: get_job, get_job_logs, delete_job, wait_for_job_status, they all call the Kubernetes API directly with name and never touch the Spark application ID.

kubeflow/spark/api/spark_client_test.py
Sharkyii: yeah this would be great to add, added this in the latest commit.

kubeflow/spark/backends/kubernetes/backend.py
Sharkyii: Hi @Shekharrajak, dug into the "make it generic" point before writing anything more, wanted to walk you through what I found.

Turns out Spark Operator already has a real answer for this, spec.monitoring with a Prometheus JMX exporter that pulls JVM, executor, shuffle, GC metrics, basically the whole Grafana use case you were describing. So the gap isn't that our metrics model is too small, it's that submit_job() doesn't expose any of that operator feature yet.

It's not just a flag to flip though. The JMX exporter jar isn't in the default Spark image, so you'd need a custom build. It also needs a Prometheus Operator stack already running in the cluster, plus a separate PodMonitor to actually get scraped. And it's not just us being cautious here, spark-operator's own docs mention this exact setup has broken across Spark versions before.

Here's what I'm thinking, in order:

- Leave get_job_metrics() as it is for this PR, just add a raw dict field so nothing ever silently goes missing, even with the small typed set we have now. No new dependencies, ready to go.
- Open a separate issue for the Prometheus side, and start with just a documented example, custom Dockerfile, the monitoring config, a PodMonitor manifest, Grafana import steps, before writing any SDK code. Want to see if people actually reach for it first.
- If they do, the SDK work would reuse the same initContainer trick FuncJob already does, so people aren't forced into a custom image, plus a check that fails clearly if Prometheus Operator isn't even installed instead of just quietly doing nothing.

Sharkyii:
@Shekharrajak PTAL

PR #697, opened 2026-07-31
feat: [KEP-107] Support spark_conf for Spark batch job submission

### LINKED ISSUE #646: KEP-107: Wire `spark_conf` through `submit_job()` for batch SparkApplication jobs

Enable `spark_conf` support for batch Spark jobs submitted via `SparkClient.submit_job()`.
Today, `submit_job()` accepts a `spark_conf: dict[str, str]` parameter but raises `NotImplementedError` when it is provided. The SparkConnect session path (`client.connect()`) already supports `spark_conf` end-to-end.
This issue tracks wiring `spark_conf` through the batch job submission path so the SDK sets `spec.sparkConf` on the `SparkApplication` CR created by the Kubernetes backend.

**Proposed changes:**
- Remove the `NotImplementedError` guard for `spark_conf` in `SparkClient.submit_job()`
- Pass `spark_conf` through `RuntimeBackend.submit_job()` → `KubernetesBackend.submit_job()` → `build_spark_application_cr()`
- Set `spec.spark_conf` on `SparkV1beta2SparkApplicationSpec` when the user provides a non-empty dict
- Add unit tests at the utils, backend, and client layers (mirroring existing Connect tests such as `TestBuildSparkConnectCr.test_with_spark_conf`)
- Update `examples/spark/batch_job_lifecycle.py` to show `spark_conf` usage

**Out of scope for this issue:**
- options pattern for batch jobs (Labels, Annotations, etc.)
- Driver / Executor dataclasses on submit_job() (see [#613](https://github.com/kubeflow/sdk/issues/613))

### Why is this needed?

PR #521 added batch job submission and lifecycle APIs (submit_job, get_job, list_jobs, wait_for_job_status, get_job_logs, delete_job) but deliberately deferred spark_conf support. KEP-107 documents spark_conf on submit_job() as part of the batch job API.

Without spark_conf, users cannot configure Spark properties through the SDK for batch workloads. Common settings include:
- Query tuning: spark.sql.adaptive.enabled, spark.sql.shuffle.partitions
- Performance: spark.serializer, memory and shuffle settings
- Storage integration: spark.hadoop.fs.s3a.* and similar Hadoop configs for S3/GCS/HDFS
- Dynamic allocation: spark.dynamicAllocation.enabled and related properties
Users who need these today must bypass the SDK and hand-author SparkApplication YAML, which defeats the purpose of KEP-107's simplified Python API.

The Spark Operator already supports spec.sparkConf on SparkApplication — this issue is SDK wiring only; no operator or CRD changes are required.

Related
- KEP-107: [proposals/107-spark-client/README.md](https://github.com/kubeflow/sdk/blob/main/proposals/107-spark-client/README.md)
- PR #521 (merged) — deferred spark_conf as future work
- Tracking issue #520 — "Advanced job configuration (spark_conf, options)"

### Love this feature?

Goku2099:
## Summary

This PR adds support for `spark_conf` when submitting Spark batch jobs with `SparkClient`.

Previously, passing `spark_conf` to `SparkClient.submit_job()` resulted in a `NotImplementedError`. With this change, user-provided Spark configuration is propagated to the generated `SparkApplication` resource.

## Changes

- Remove `NotImplementedError` for `spark_conf` in `SparkClient.submit_job()`.
- Forward `spark_conf` through the Kubernetes backend.
- Populate `SparkApplication.spec.sparkConf` with the provided Spark configuration.
- Update public API docstrings.
- Add unit tests covering `spark_conf` propagation to the generated `SparkApplication`.

## Example

```python
client.submit_job(
    job=FileJob(
        file_source="s3://bucket/job.py",
    ),
    spark_conf={
        "spark.sql.shuffle.partitions": "8",
        "spark.sql.adaptive.enabled": "true",
    },
)
```
closes: #646 
ref: #520

### DIFF
--- a/kubeflow/spark/api/spark_client_test.py
+++ b/kubeflow/spark/api/spark_client_test.py
@@ -100,10 +100,10 @@ def test_create_and_connect(test_case: TestCase):
         ),
         (
             FileJob(file_source="s3://bucket/job.py"),
-            {"spark.executor.memory": "4g"},
+            [],
             None,
-            None,
-            NotImplementedError,
+            ValueError("spark_conf must be a dictionary."),
+            ValueError,
         ),
     ],
 )
@@ -167,5 +167,9 @@ def test_submit_job_success(job, options):
         assert name == "spark-job-123"
 
         backend.submit_job.assert_called_once_with(
-            job=job, num_executors=None, resources_per_executor=None, options=options
+            job=job,
+            num_executors=None,
+            resources_per_executor=None,
+            spark_conf=None,
+            options=options,
         )
--- a/kubeflow/spark/backends/kubernetes/backend.py
+++ b/kubeflow/spark/backends/kubernetes/backend.py
@@ -893,6 +893,7 @@ class KubernetesBackend(RuntimeBackend):
         num_executors: int | None = None,
         resources_per_executor: dict[str, str] | None = None,
         options: list | None = None,
+        spark_conf: dict[str, str] | None = None,
     ) -> SparkJob:
         """Submit a SparkApplication for batch execution.
 
@@ -908,6 +909,8 @@ class KubernetesBackend(RuntimeBackend):
 
             options:
                 List of additional Spark configuration options.
+            spark_conf:
+                Spark configuration properties to set on the SparkApplication.
 
         Returns:
             SparkJob information object.
@@ -936,6 +939,7 @@ class KubernetesBackend(RuntimeBackend):
                 resources_per_executor=resources_per_executor,
                 options=options,
                 backend=self,
+                spark_conf=spark_conf,
             )
 
         else:
@@ -948,6 +952,7 @@ class KubernetesBackend(RuntimeBackend):
                 resources_per_executor=resources_per_executor,
                 options=options,
                 backend=self,
+                spark_conf=spark_conf,
             )
 
         # The Name option may override the auto-generated name.
--- a/kubeflow/spark/backends/kubernetes/utils.py
+++ b/kubeflow/spark/backends/kubernetes/utils.py
@@ -136,18 +136,8 @@ def _resolve_executor_resources(
     executor: Executor | None = None,
     num_executors: int | None = None,
     resources_per_executor: dict[str, str] | None = None,
-    spark_conf: dict[str, str] | None = None,
 ) -> tuple[int, int, str]:
     """Resolve executor configuration.
-
-    Precedence rules:
-        - Executor instances:
-          ``spark_conf`` > ``executor.num_instances`` >
-          ``num_executors`` > default.
-        - Executor resources:
-          ``spark_conf`` >
-          ``executor.resources_per_executor`` >
-          ``resources_per_executor`` > default.
 
     Args:
         executor:
@@ -158,9 +148,6 @@ def _resolve_executor_resources(
 
         resources_per_executor:
             Resource requirements.
-
-        spark_conf:
-            Spark configuration properties.
 
     Returns:
         Tuple containing ``(instances, cores, memory)``.
@@ -196,22 +183,6 @@ def _resolve_executor_resources(
         if "memory" in resource_dict:
             memory = _memory_kubernetes_to_spark(
                 resource_dict["memory"],
-            )
-
-    if spark_conf:
-        if "spark.executor.instances" in spark_conf:
-            instances = int(
-                spark_conf["spark.executor.instances"],
-            )
-
-        if "spark.executor.cores" in spark_conf:
-            cores = _validate_cpu_value(
-                spark_conf["spark.executor.cores"],
-            )
-
-        if "spark.executor.memory" in spark_conf:
-            memory = _memory_kubernetes_to_spark(
-                spark_conf["spark.executor.memory"],
             )
 
     return instances, cores, memory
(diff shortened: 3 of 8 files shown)


Goku2099:

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1309,6 +1320,8 @@ def test_get_spark_application_cr_from_func_job(
     assert app.spec.executor.cores == 2
     assert app.spec.executor.memory == "4g"
 
+    assert app.spec.spark_conf == test_case.config["spark_conf"]
+
     print("test execution complete")
```
tariq-hasan, line 1645: For the cases of `test_get_spark_application_cr_from_file_job` and `test_get_spark_application_cr_from_func_job` I think probably a good idea to have dedicated test cases per behavior (`spark_conf` included) similar to what we have for [`test_build_spark_connect_cr`](https://github.com/Goku2099/sdk/blob/c9721032a018b9e0e6d7d4b30cc3c8492c07cb49/kubeflow/spark/backends/kubernetes/utils_test.py#L291-L440).

kubeflow/spark/api/spark_client_test.py
```diff
@@ -170,4 +163,5 @@ def test_submit_job_success(job):
             job=job,
             num_executors=None,
             resources_per_executor=None,
+            spark_conf=None,
         )
```
tariq-hasan, line 175: I think better to combine `test_submit_job_validation` and `test_submit_job_success` and have all test cases grouped under `test_submit_job` for consistency with how unit tests in the packages are structured in general - using the `TestCase` pattern.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -742,6 +744,7 @@ def get_spark_application_cr_from_file_job(
                 num_executors=num_executors,
                 resources_per_executor=resources_per_executor,
             ),
+            spark_conf=spark_conf,
```
tariq-hasan: We can do `spark_conf=spark_conf or None,` to match `arguments=arguments or None,`.

Same in `get_spark_application_cr_from_func_job`

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan, line 189: We need to have a precedence rule here for the case when executor resources are passed in via `spark_conf`.

In the case when executor resources are passed in via `spark_conf` they should have precedence over any other resource specification.

Granted - a user would not necessarily want to use `spark_conf` to pass in executor resources. They'd want to use `spark_conf` for advanced settings. But either way this is a correctness issue so it would need to be addressed.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -97,13 +97,6 @@ def test_create_and_connect(test_case: TestCase):
             ValueError("`job.file_source` must be a non-empty string."),
             ValueError,
         ),
-        (
-            FileJob(file_source="s3://bucket/job.py"),
-            {"spark.executor.memory": "4g"},
-            None,
-            None,
-            NotImplementedError,
-        ),
```
tariq-hasan, line 107: Can we add a test case to have coverage for `spark_conf`?

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan, line 287: Similar to how we validate `cpu` do we also want to validate the type for `spark_conf` key and value?

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -970,6 +975,7 @@ def submit_job(
                 func_args=job.func_args,
                 num_executors=num_executors,
                 resources_per_executor=resources_per_executor,
+                spark_conf=spark_conf,
             )
 
         try:
```
tariq-hasan, line 966: Having `spark_conf` test cases for the `backend` would be nice as well.

tariq-hasan (commented):
Hi @Goku2099! I have added a few comments.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -180,6 +198,22 @@ def _resolve_executor_resources(
...
+
+        if "spark.executor.cores" in spark_conf:
+            cores = _validate_cpu_value(
+                spark_conf["spark.executor.cores"],
+            )
+
+        if "spark.executor.memory" in spark_conf:
+            memory = _memory_kubernetes_to_spark(
+                spark_conf["spark.executor.memory"],
+            )
```
andreyvelich: Why do we need to process setting one by one?
For SparkConnect we directly embed everything to the base_conf: 
https://github.com/Goku2099/sdk/blob/8d59a0ab4f897ba677819b307f28e281ea399901/kubeflow/spark/backends/kubernetes/utils.py#L542-L549

Goku2099: I implemented the explicit handling to satisfy the precedence rule discussed in Tariq's review comment https://github.com/kubeflow/sdk/pull/697#discussion_r3694329877 for spark.executor.instances, spark.executor.cores, and spark.executor.memory. For Spark Connect we currently forward spark_conf directly without interpreting those keys. Would you prefer we keep the same behavior for batch jobs as well, or should batch jobs continue handling these executor-related properties explicitly to preserve the precedence rule?

andreyvelich: I think we should do the same as for `connect()`. @tariq-hasan Do you have any objections with that approach?
We can introduce additional validation in the future if needed.

tariq-hasan: No objections from me.

The precedence would not be exercised in real use cases anyway since one would want to use top-level fields such as `resource_per_executor` instead of passing them through `spark_conf`.

Goku2099: @andreyvelich @tariq-hasan  Updated this to match the Spark Connect behavior by removing the explicit spark_conf handling and updated the tests accordingly. Please take another look.

andreyvelich (commented):
Thanks @Goku2099!
/lgtm
/approve

PR #698, opened 2026-07-31
fix(trainer): copy default ignore_patterns for model initializers

### LINKED ISSUE #612: HuggingFaceModelInitializer and S3ModelInitializer share a mutable default ignore_patterns list across instances
### What happened?

Description

While reviewing kubeflow/trainer/types/types.py, I noticed that both HuggingFaceModelInitializer and S3ModelInitializer define ignore_patterns as:

ignore_patterns: list[str] | None = field(
    default_factory=lambda: constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS
)

At first glance this looks correct because default_factory is used. However, the factory returns the module-level mutable list directly instead of creating a new list for each instance.

INITIALIZER_DEFAULT_IGNORE_PATTERNS is defined as:

INITIALIZER_DEFAULT_IGNORE_PATTERNS = [
    "*.msgpack",
    "*.h5",
    "*.bin",
    "*.pt",
    "*.pth",
]

Since the same list object is returned every time, all initializer instances share the same underlying ignore_patterns list.

Verified locally
from kubeflow.trainer.types.types import HuggingFaceModelInitializer

a = HuggingFaceModelInitializer(storage_uri="hf://user/model1")
b = HuggingFaceModelInitializer(storage_uri="hf://user/model2")

print(a.ignore_patterns is b.ignore_patterns)

Output:

True

Both instances reference the exact same list object.

To verify the impact:

print("Before:")
print("A:", a.ignore_patterns)
print("B:", b.ignore_patterns)

a.ignore_patterns.append("*.custom")

print("After:")
print("A:", a.ignore_patterns)
print("B:", b.ignore_patterns)

Output:

Before:
A: ['*.msgpack', '*.h5', '*.bin', '*.pt', '*.pth']
B: ['*.msgpack', '*.h5', '*.bin', '*.pt', '*.pth']

After:
A: ['*.msgpack', '*.h5', '*.bin', '*.pt', '*.pth', '*.custom']
B: ['*.msgpack', '*.h5', '*.bin', '*.pt', '*.pth', '*.custom']

Even though only a.ignore_patterns was modified, the change is immediately visible in b.ignore_patterns because both instances share the same mutable list.

Existing pattern in the codebase

The same file already uses the correct pattern for CustomTrainer.pip_index_urls:

pip_index_urls: list[str] = field(
    default_factory=lambda: list(constants.DEFAULT_PIP_INDEX_URLS)
)

Creating a new list for every instance avoids sharing mutable state.

Expected behavior

Each HuggingFaceModelInitializer and S3ModelInitializer instance should receive its own independent copy of the default ignore_patterns list.

Suggested fix
ignore_patterns: list[str] | None = field(
    default_factory=lambda: list(constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS)
)

This matches the pattern already used elsewhere in the same file.

Environment

Repository: kubeflow/sdk
kubeflow/trainer/types/types.py
Affected classes:
HuggingFaceModelInitializer
S3ModelInitializer

I'd be happy to submit a PR if this behavior is considered a bug.

Each `HuggingFaceModelInitializer` and `S3ModelInitializer` instance should receive its own independent copy of the default `ignore_patterns` list.

Mutating the `ignore_patterns` of one instance should not affect any other initializer instance.

For example, after creating two independent initializer objects, appending a new ignore pattern to one instance should leave the other instance unchanged.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

sahilleth:
## Summary
- Copy `INITIALIZER_DEFAULT_IGNORE_PATTERNS` in `HuggingFaceModelInitializer` and `S3ModelInitializer` default factories so each instance gets its own list
- Add unit tests verifying mutations on one initializer do not affect another

Fixes #612

## Test plan
- [x] `uv run pytest kubeflow/trainer/types/types_test.py -v`

### DIFF
--- a/kubeflow/trainer/types/types.py
+++ b/kubeflow/trainer/types/types.py
@@ -427,7 +427,7 @@ class HuggingFaceModelInitializer(BaseInitializer):
     """
 
     ignore_patterns: list[str] | None = field(
-        default_factory=lambda: constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS
+        default_factory=lambda: list(constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS)
     )
     access_token: str | None = None
 
@@ -448,7 +448,7 @@ class S3ModelInitializer(BaseInitializer):
     """
 
     ignore_patterns: list[str] | None = field(
-        default_factory=lambda: constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS
+        default_factory=lambda: list(constants.INITIALIZER_DEFAULT_IGNORE_PATTERNS)
     )
     endpoint: str | None = None
     access_key_id: str | None = None
--- a/kubeflow/trainer/types/types_test.py
+++ b/kubeflow/trainer/types/types_test.py
@@ -149,6 +149,27 @@ def test_hugging_face_model_initializer(test_case: TestCase):
 
 
 @pytest.mark.parametrize(
+    "initializer_cls,storage_uri",
+    [
+        (types.HuggingFaceModelInitializer, "hf://user/model"),
+        (types.S3ModelInitializer, "s3://bucket/model"),
+    ],
+)
+def test_model_initializer_ignore_patterns_are_independent(initializer_cls, storage_uri):
+    """Each model initializer instance gets its own ignore_patterns list."""
+    first = initializer_cls(storage_uri=storage_uri)
+    second = initializer_cls(storage_uri=storage_uri)
+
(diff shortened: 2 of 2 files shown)


sahilleth:
Hi maintainers — could you please approve the workflow runs (Unit Test / E2E / Docs)? This is a fork PR and the jobs are currently in `action_required` state. Thanks!

PR #707, opened 2026-08-01
fix: classify HuggingFace dataset dir with a dotted parent as data_dir, not data_files

Atishyy27:
get_args_using_torchtune_config decides dataset.data_files vs dataset.data_dir with `"." in relative_path`, which matches a dot anywhere in the path. so a directory whose parent segment contains a dot, e.g. hf://tatsu-lab/alpaca/my.v2/train (relative_path "my.v2/train"), is misclassified as data_files, and torchtune then tries to load a directory as a file.

fix: check only the final path component (`"." in relative_path.split("/")[-1]`) so the extension heuristic looks at the leaf name, not parent dirs. no change for the existing file/dir cases.

added a test case (directory with a dot in a parent path). before the change it emits data_files=.../my.v2/train (test fails); after, data_dir=.../my.v2/train:

    pytest kubeflow/trainer/backends/kubernetes/utils_test.py -k torchtune

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -527,7 +527,7 @@ def get_args_using_torchtune_config(
         parts = storage_uri_parsed.path.strip("/").split("/")
         relative_path = "/".join(parts[1:]) if len(parts) > 1 else "."
 
-        if relative_path != "." and "." in relative_path:
+        if relative_path != "." and "." in relative_path.split("/")[-1]:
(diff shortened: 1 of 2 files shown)


HarshPopat23:
Hii @Atishyy27 welcome to kubflow,
Nice targeted fix. Could we also add the complementary case for a file under a dotted parent-for example, my.v2/train/data.json-and assert that it remains data_files? This would test both sides of the leaf-component heuristic and guard against future overcorrection.

Atishyy27:
thanks! added that case: `my.v2/train/data.json` now asserts `data_files`, so the dotted-parent fix doesn't overcorrect a real file into `data_dir`. pushed.

PR #664, opened 2026-07-26
fix(spark): handle client_configuration, context, and default namespace  in KubernetesBackend

vishalmore90:
**What this PR does / why we need it**:

Aligns `kubeflow.spark.backends.kubernetes.KubernetesBackend.__init__` with the standard `KubernetesBackendConfig` handling pattern established in `kubeflow.trainer` and `kubeflow.optimizer`.

**Key changes**:
1. **Support `client_configuration`**: Passes `backend_config.client_configuration` to `client.ApiClient` when instantiating `CustomObjectsApi` and `CoreV1Api`.
2. **Kubeconfig context support**: Passes both `config_file` and `context` to `config.load_kube_config(config_file=..., context=...)`.
3. **Dynamic target namespace resolution**: Resolves `namespace` using `common_utils.get_default_target_namespace(context)` when `backend_config.namespace` is `None` instead of statically defaulting to `"default"`.
4. **Unit tests**: Added unit tests in `kubeflow/spark/backends/kubernetes/backend_test.py` covering custom `client_configuration`, `config_file`/`context` loading, and dynamic target namespace resolution.

**Which issue(s) this PR fixes** *(optional, in `Fixes #<issue number>, #<issue number>, ...` format, will close the issue(s) when PR gets merged)*:

Fixes #663

**Checklist:**

- [x] [Docs](https://www.kubeflow.org/docs/) included if any changes are user facing
- [x] Tests added and updated for `KubernetesBackend` initialization logic

PR #719, opened 2026-08-04
fix(spark): include the underlying API error in SparkConnect failure messages

### LINKED ISSUE #761: 403 responses from the Kubernetes API are reported without indicating which RBAC verbs are missing
### What happened?

When a caller lacks RBAC for a Spark Connect session operation, the 403 from the
Kubernetes API is caught by a generic `except Exception` and re-raised as
`Failed to create SparkConnect: <ns>/<name>`.

The user cannot tell whether the problem is permissions or a broken operator, and an
administrator receiving that message has nothing to act on. Determining that
`sparkconnects` requires `create` and `get`, and `pods/log` requires `get`, currently
means reading the SDK source.

The error names the missing verbs, the resource, and the namespace, so the caller knows
it is a permissions problem and what to ask an administrator to grant.

### Environment

- Kubeflow SDK: main
- Kubernetes: v1.35.0 (kind)
- Spark Operator: 2.3.0

### Impacted by this bug?

vikas-saxena02:

When the caller lacks RBAC, the failure is opaque. The 403 is caught by a generic
`except Exception` and re-raised as `Failed to create SparkConnect: <ns>/<name>`. The user
cannot tell that the problem is permissions rather than a broken operator, and an
administrator receiving that message has nothing to act on either. Working out that
`sparkconnects` needs `create` and `get`, and `pods/log` needs `get`, currently means
reading the SDK source.

This PR translates 403s into a message naming the missing verbs, along with a `Role` and
`RoleBinding` that a cluster administrator can apply directly.

**Key changes**:

1. **403 translation in the five session methods**: `_create_session`, `get_session`,
   `list_sessions`, `delete_session`, and `get_session_logs` detect an `ApiException` with
   status 403 and raise a `RuntimeError` whose message names the required verbs, the
   resource, and the namespace. Non-403 errors keep their existing messages.

2. **Log streaming**: `read_pod_logs` wraps every failure as `RuntimeError` but preserves
   the original as `__cause__`, so a 403 raised while streaming is only visible by
   inspecting it. The handler now does, and names `pods/log` in the core API group rather
   than `sparkconnects`.

3. **Docs**: a Permissions section in `sessions.rst` mapping each operation to the verbs it
   requires, with a worked example of the error.

**Example of the new error**:

```
Not permitted to create a SparkConnect session in namespace 'team-analytics'.
Required RBAC verbs on 'sparkconnects' in API group 'sparkoperator.k8s.io': create, get.
Ask a cluster administrator to grant the following, or bind an existing role that
includes these rules to your ServiceAccount:

apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: spark-connect-sparkconnects-create-get
  namespace: team-analytics
rules:
  - apiGroups: ["sparkoperator.k8s.io"]
    resources: ["sparkconnects"]
    verbs: ["create", "get"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
...
```

**Scope**

This started considerably larger. Following review feedback I removed the
`SelfSubjectAccessReview` preflight, the permission cache, the custom
`SparkPermissionError` exception class, and the shared `ApiClient` — 403s now raise a plain
`RuntimeError`, so nothing about the exception contract changes. The PR is 456 insertions
across 6 files, down from 989 across 14, and roughly half of what remains is tests.

The per-call `namespace` argument is also not included. @tariq-hasan suggested the
Trainer pattern in `__init__` as the fix for namespace resolution, which is what
[#664](https://github.com/kubeflow/sdk/pull/664) does. I've left `__init__` alone here so
the two don't conflict. Happy to add per-call namespace targeting as a follow-up if it's
wanted — see the thread below.

**Notes**

- The generated `Role` name includes the verbs, so remediations for different operations on
  the same resource don't overwrite one another when applied in sequence.
- The subject `ServiceAccount` namespace is a separate placeholder, since a caller in one
  namespace may be granted access to another.

Fixes #761 

**Checklist**:

- [x] Docs included (`docs/source/spark/sessions.rst`)
- [x] Tests added


kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -726,40 +919,59 @@ def get_session_logs(
...
+                    f"Timeout to get logs for {constants.SPARK_CONNECT_KIND}: "
+                    f"{target_namespace}/{name}"
                 ) from e
 
             except RuntimeError as e:
                 raise RuntimeError(
-                    f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}"
+                    f"Failed to get logs for {constants.SPARK_CONNECT_KIND}: "
+                    f"{target_namespace}/{name}"
                 ) from e
```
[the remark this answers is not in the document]
vikas-saxena02: code change made and fixed

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -114,6 +117,107 @@ def __init__(self, backend_config: KubernetesBackendConfig):
 
         self.custom_api = client.CustomObjectsApi()
         self.core_api = client.CoreV1Api()
+        self.authorization_api = client.AuthorizationV1Api()
```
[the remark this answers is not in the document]
vikas-saxena02: Added an API client to fix this

examples/spark/README.md
```diff
@@ -8,6 +8,7 @@ This directory contains examples for using the Kubeflow Spark SDK.
 - **spark_advanced_options.py** - Advanced configuration with Driver/Executor objects
 - **demo_existing_sparkconnect.py** - Connect to existing SparkConnect cluster
 - **test_connect_url.py** - Test URL-based connection to Spark Connect
+- **spark_connect_namespace.py** - Namespace targeting and RBAC error handling
```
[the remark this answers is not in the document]
vikas-saxena02: added the missing file

kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -197,44 +316,62 @@ def _create_session(
...
 
         Raises:
+            SparkPermissionError:
+                If the caller lacks RBAC permission to read SparkConnect resources.
             TimeoutError:
                 If getting the SparkConnect resource times out.
             RuntimeError:
                 If the SparkConnect resource cannot be retrieved.
         """
+        target_namespace = self._resolve_namespace(namespace)
```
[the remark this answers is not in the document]
vikas-saxena02: made code change for the same

docs/source/spark/index.rst
```diff
@@ -57,6 +57,62 @@ Quick Example
...
+   client.delete_session("my-session", namespace="team-analytics")
+
+See ``examples/spark/spark_connect_namespace.py`` for a complete, runnable version.
+
+Permissions
+-----------
+
+Creating a session requires RBAC on ``sparkconnects`` in the target namespace, and reading
+session logs requires ``get`` on ``pods/log``. The SDK checks these before submitting
+anything, so a caller without access gets a ``SparkPermissionError`` naming the missing
```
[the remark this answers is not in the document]
vikas-saxena02: Fixed. The Permissions section is now a table mapping each operation to the verbs it needs, and it calls out explicitly that get_session_logs() needs get on sparkconnects to look up the driver pod, not just pods/log. Also documents that connect() needs get alongside create for the readiness poll, which was implicit before.

tariq-hasan:
Hi @vikas-saxena02! Thanks for raising the PR.

> A user should be able to start an interactive Spark session from a Kubeflow Notebook without needing a cluster administrator. Two things get in the way today.
> 
> First, the target namespace is fixed at client construction. `SparkClient` reads it from `KubernetesBackendConfig` and every session method uses it, so working across more than one namespace means building more than one client.

I believe the gap might be due to how the namespace is [initialized](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend.py#L102). Following the Trainer [pattern](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L43-L44) should fix it.

> Second, when the caller lacks RBAC, the failure is opaque. The 403 is caught by a generic `except Exception` and re-raised as `Failed to create SparkConnect: <ns>/<name>`. The user cannot tell that the problem is permissions rather than a broken operator, and an administrator receiving that message has nothing to act on either. Working out that `sparkconnects` needs `create` and `get`, and `pods/log` needs `get`, currently means reading the SDK source.

I'm thinking a simple error message might do here without introducing all the complex machinery in the PR.

> * **Out of scope, but worth flagging**: `DEFAULT_SERVICE_ACCOUNT = "spark-operator-spark"`
>   does not exist in a typical Kubeflow profile namespace. A session created there will
>   pass the RBAC check and then fail at executor creation. Happy to open a follow-up issue.

As it stands the chart installation for the operator creates this [service account](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/templates/spark/_helpers.tpl#L24-L33) in the namespaces that it [watches](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/values.yaml#L510-L522). We plan to remove the hardcoding once https://github.com/kubeflow/spark-operator/issues/3049 lands.

vikas-saxena02:
> Hi @vikas-saxena02! Thanks for raising the PR.
> 
> > A user should be able to start an interactive Spark session from a Kubeflow Notebook without needing a cluster administrator. Two things get in the way today.
> > First, the target namespace is fixed at client construction. `SparkClient` reads it from `KubernetesBackendConfig` and every session method uses it, so working across more than one namespace means building more than one client.
> 
> I believe the gap might be due to how the namespace is [initialized](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend.py#L102). Following the Trainer [pattern](https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/kubernetes/backend.py#L43-L44) should fix it.
> 
> > Second, when the caller lacks RBAC, the failure is opaque. The 403 is caught by a generic `except Exception` and re-raised as `Failed to create SparkConnect: <ns>/<name>`. The user cannot tell that the problem is permissions rather than a broken operator, and an administrator receiving that message has nothing to act on either. Working out that `sparkconnects` needs `create` and `get`, and `pods/log` needs `get`, currently means reading the SDK source.
> 
> I'm thinking a simple error message might do here without introducing all the complex machinery in the PR.
> 
> > * **Out of scope, but worth flagging**: `DEFAULT_SERVICE_ACCOUNT = "spark-operator-spark"`
> >   does not exist in a typical Kubeflow profile namespace. A session created there will
> >   pass the RBAC check and then fail at executor creation. Happy to open a follow-up issue.
> 
> As it stands the chart installation for the operator creates this [service account](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/templates/spark/_helpers.tpl#L24-L33) in the namespaces that it [watches](https://github.com/kubeflow/spark-operator/blob/master/charts/spark-operator-chart/values.yaml#L510-L522). We plan to remove the hardcoding once [kubeflow/spark-operator#3049](https://github.com/kubeflow/spark-operator/issues/3049) lands.

@tariq-hasan Thanks for the review — agreed on the second point, and I'll cut it back.

You're right that the preflight isn't earning its complexity. I checked: the `SelfSubjectAccessReview` path and the 403-translation path call the same formatter with the same verb list, so the message the user sees is identical either way. The preflight only surfaces it one round trip earlier, and a 403 on create means nothing was created anyway.

I'll drop the `SelfSubjectAccessReview`, the permission cache, and `SparkPermissionError`, and keep just the 403 → actionable-message translation. That's the part that was actually missing, and it's a much smaller diff — the user still gets:

```
Not permitted to create a SparkConnect session in namespace 'team-analytics'.
Missing RBAC verbs on 'sparkconnects' in API group 'sparkoperator.k8s.io': create, get.
Ask a cluster administrator to grant the following, or bind an existing role that
includes these rules to your ServiceAccount:

apiVersion: rbac.authorization.k8s.io/v1
kind: Role
...
```

On the namespace — I think there are two separate things here, and my PR description conflated them, which is what invited the confusion.

The Trainer pattern in `__init__` fixes the **default**: you get the namespace you're running in rather than `default`. That's a real gap and I agree it should be fixed. [#664](https://github.com/kubeflow/sdk/pull/664) already does exactly that, which is why I deliberately left `__init__` alone here. Happy to fold it into this PR instead, or rebase on #664 once it lands — whichever you'd prefer.

What `__init__` doesn't give you is targeting more than one namespace from a single client, which is what the per-call `namespace` argument adds. The motivating case is a notebook user working in their own profile namespace who also wants a durable session in a shared project namespace — today that means constructing and tracking two clients. Let me know if you think that's worth keeping; if it's too niche I'm happy to drop it and reduce this PR to just the error message.

On the service account: thanks, that's clearer than what I'd worked out. I'd missed that [kubeflow/spark-operator#3049](https://github.com/kubeflow/spark-operator/issues/3049) already covers it, and good to see the Kubeflow SDK case is explicitly in scope there. I'll drop my note about it from the description.

juliusvonkohout:
@vikas-saxena02 please rebase to master

@nabuskey please take a look

vikas-saxena02:
@tariq-hasan i ahve addressed your review comments, please review again!!

vikas-saxena02:
> @vikas-saxena02 please rebase to master
> 
> @nabuskey please take a look

@juliusvonkohout rebse also done @nabuskey  its ready for review

kubeflow/spark/backends/kubernetes/backend.py
tariq-hasan, line 230: The only change we really need to make to expose the error is to make this change:

```suggestion
        raise RuntimeError(
            f"Failed to get {constants.SPARK_CONNECT_KIND}: {self.namespace}/{name}: {e}"
        ) from e
```

Batch jobs already do [this](https://github.com/vikas-saxena02/kubeflowsdk/blob/0ec5eb33f2b4d3449ee559209229b25a2ff58e0a/kubeflow/spark/backends/kubernetes/backend.py#L1056-L1059). Interactive sessions was the gap.

We should not introduce all the complex machinery in the PR for this simple error message exposure.

vikas-saxena02:
@tariq-hasan Just to check we're talking about the same thing — I've already cut the `SelfSubjectAccessReview` preflight, the permission cache, the custom exception class, and the shared `ApiClient` since that comment. The PR is now 456 insertions across 6 files, down from 989 across 14, and 214 of those insertions are tests.

The remaining non-trivial piece is the helper that generates the `Role`/`RoleBinding` manifest — 72 lines in `utils.py`. If that's the machinery you mean, I'm happy to reduce it to a plain sentence naming the missing verbs:

```
Failed to create SparkConnect: default/my-session: (403)
Requires RBAC verbs on 'sparkconnects' in API group 'sparkoperator.k8s.io': create, get.
```

That's roughly 20 lines instead of 72, and it still closes the gap we agreed on — the user learns the problem is permissions rather than a broken operator, and which verbs to ask for. The manifest is a convenience for whoever grants the access, not the core of the fix, so I'm glad to drop it if you'd prefer the smaller diff.

vikas-saxena02:
@andreyvelich @tariq-hasan @juliusvonkohout can i please get a review on this

vikas-saxena02:
@juliusvonkohout I went through Copilot's five suppressed comments. One was a genuine issue and is fixed; the other four were flagging the code against a PR description I hadn't updated after cutting scope.

**Fixed:** the permissions table in `sessions.rst` implied every `connect()` call needs Kubernetes RBAC, but `connect(base_url=...)` attaches to an existing endpoint over gRPC and never calls the API server. The row is now qualified as create mode, with a note covering the connect-mode case.

**Not applicable:** three comments asked for the per-call `namespace` argument, the `SelfSubjectAccessReview` preflight, and a custom `SparkPermissionError` — all of which I removed following the review feedback above. A fourth pointed at a `examples/spark/spark_connect_namespace.py` that no longer exists for the same reason. Copilot was comparing the diff against the original description, which still described the larger version.

I've rewritten the PR description to match what's actually here now, so it should be easier to review. Also took the chance to move the Permissions section above "When Things Go Wrong", since it's reference material rather than troubleshooting.

nabuskey (commented):
LGTM

vikas-saxena02:
@tariq-hasan please review again and merge if you are happy with it

vikas-saxena02:
> ## Pull request overview
> Copilot reviewed 6 out of 6 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (7)

@juliusvonkohout These are the same five suppressed comments from the previous run. Four are no longer applicable.

**Already fixed:** the permissions table implied every `connect()` call needs Kubernetes RBAC, but `connect(base_url=...)` attaches to an existing endpoint over gRPC and never calls the API server. The row is now qualified as create mode, with a note covering the connect-mode case — see 72c8015. This was the one genuine issue of the five.

**Not applicable:** three comments ask for the per-call `namespace` argument, the `SelfSubjectAccessReview` preflight, and a custom `SparkPermissionError`. All three were removed following the review feedback above, so re-adding them would revert a requested change. A fourth points at an `examples/spark/spark_connect_namespace.py` that no longer exists for the same reason — confirmed absent, with no dangling references in docs or examples.

Those four were comparing the diff against the original PR description, which described the larger version. The description has since been rewritten to match what is actually here.

vikas-saxena02:
@andreyvelich @tariq-hasan @juliusvonkohout please review this if you have sometime this week.

vikas-saxena02:
> LGTM

@tariq-hasan @juliusvonkohout i have received LGTM fro @nabuskey , waiting for a proper LGTM (with a slash) and an apparoval

juliusvonkohout:
> LGTM

/lgtm
/approve

juliusvonkohout:
@andreyvelich

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -986,3 +986,74 @@ def get_spark_application_info_from_cr(
...
+    namespace: str,
+    group: str,
+    resource: str,
+    verbs: Sequence[str],
+) -> str:
+    """Build an error message naming the RBAC rules an operation requires.
+
+    Users who hit a 403 are typically working in a namespace they do not administer, so
+    the message carries a manifest they can hand to a cluster administrator rather than
+    only reporting that access was denied.
```
andreyvelich: Do we really need that in addition to simple error message? Usually, cluster admins should read an instruction guide which RBAC rules are required to use `SparkClient()`.
We didn't introduce this rule for `TrainerClient()`, and I didn't get much complains from the end-users.

@kramaranya @robert-bell @Fiona-Waters @tariq-hasan @Sridhar1030 @abhijeet-dhumal Any thoughts on this?

andreyvelich: /hold

tariq-hasan: I don't think the PR changes belong in the SDK.

I think we'd want to decouple client-side semantics from server-side concerns.

The SDK is a thin layer of abstraction that's just meant to collapse control-plane logic and expose API semantics that's idiomatic for the data scientist.

This PR would be adding in changes in the wrong layer so there are overall design concerns with these changes.

This would also set a precedent for changes that don't belong in the SDK.

I'd already mentioned to add the `: {e}` that already passes through the control-plane error. That's the simplest  change that does not break the current design and also avoids introducing all the heavy machinery in the PR (that would also be a maintenance concern).

vikas-saxena02: Fair question, and I'm happy to drop the manifest.

The part I'd argue for keeping is naming the verbs. Without it the failure is
`Failed to create SparkConnect: <ns>/<name>` — the user can't tell a permissions problem
from a broken operator, and there's nothing to search for. With it they at least know
what to ask for:

```
Not permitted to create a SparkConnect session in namespace 'team-analytics'.
Required RBAC verbs on 'sparkconnects' in API group 'sparkoperator.k8s.io': create, get.
```

That's ~20 lines instead of ~70, and it drops the `Role`/`RoleBinding` generation entirely.

On the docs point — I agree that's where the authoritative list belongs, and this PR adds
it to `sessions.rst`. My reason for putting it in the error too is that the person hitting
it is usually not the person who reads the install guide: a notebook user gets the 403,
and an admin gets forwarded a message with no actionable content in it.

Worth noting a Kubeflow Profile namespace grants no RBAC for `sparkconnects` at all —
`kubeflow-edit` predates the CRD — so every notebook user hits this on a stock install,
not just misconfigured ones. That's what prompted the PR.

Happy to go either way. If the consensus is a plain message, I'll cut it back to the two
lines above.

vikas-saxena02: @andreyvelich  @tariq-hasan Stepping back on intent — the goal here was user experience rather than anything architectural. This is what a notebook user gets from the released SDK today:

```
Failed to create SparkConnect: default/spark-connect-aebe5a09
```
Screenshot 2026-08-21 at 10 29 31 pm


That's the entire message which is not helpful at all. Underneath it is a 403 saying
`sparkconnects.sparkoperator.k8s.io is forbidden: User "system:serviceaccount:...:default-editor"
cannot create resource "sparkconnects"` — but the generic `except Exception` swallows it, so
nothing reaches the user. There's no way to tell a permissions problem from a broken operator,
and nothing to search for.

I agree with the layering point, and the `: {e}` passthrough you suggested fixes the actual
problem — the API server's own message names the identity, verb, resource and namespace, and
message construction stays server-side. That's already in my branch that I am trying to merge.

So I'll drop `format_missing_permissions_error`, the constants, and the `_permission_error`
helper, leaving the passthrough plus tests.

One open question: should the `Permissions` section in `sessions.rst` stay? It's currently the
only place the required verbs are written down, and it's the docs half of what @andreyvelich
was pointing at. Happy to drop it if it's out of scope here.

Last but not the least, Happy to drop the manifest generation, the constants, and the `_permission_error` helper — the layering argument makes sense and I don't want to set a precedent for message construction in the SDK but I do want to get a helpful error message to the end user.

tariq-hasan: The control plane should already expose the permissions error in some form.

```
Not permitted to create a SparkConnect session in namespace 'team-analytics'.
Required RBAC verbs on 'sparkconnects' in API group 'sparkoperator.k8s.io': create, get.
```

That's why we'd want to wrap the error `: {e}` while keeping the client-side concerns decoupled.

Client-side SDK cannot make assumptions about server-side error handling. That is fundamentally a server-side concern.

If this is a permissions issue should the changes be against the manifests repo though?

https://github.com/kubeflow/community-distribution/blob/master/applications/spark/spark-operator/base/aggregated-roles.yaml

vikas-saxena02: @tariq-hasan You're right that this belongs in the manifests repo — thanks for the pointer, that's exactly the file.

`sparkconnects` isn't in any of the three roles. `kubeflow-spark-edit` and `kubeflow-spark-view`
both list only `sparkapplications` and `scheduledsparkapplications`, so `default-editor` can't
create or even list a Spark Connect session on a stock install. The aggregation labels are fine;
the resource is just missing, presumably because `SparkConnect` is a newer CRD. I'll raise it
against community-distribution with a patch.

On this PR: agreed on the layering, cutting back to the `: {e}` passthrough. Dropping
`format_missing_permissions_error`, the constants, and the `_permission_error` helper.

Should the `Permissions` section in `sessions.rst` stay or go? Happy either way.

robert-bell: I agree with @tariq-hasan. The error message shouldn't try to interpret the server-side message. It should just include it in the exception. Probably both in the message with ": {e}" and reraising from. 

The manifest is overkill and may end up being more confusing than helpful. The user can't do anything with it themselves and the admin is likely to already have roles they assign to users.

vikas-saxena02: @tariq-hasan @robert-bell @andreyvelich Cut back as discussed — the PR is now 77 insertions across two files.

`backend.py` keeps the `: {e}` passthrough on the session error messages, plus one fix in
`get_session_logs`: `read_pod_logs` wraps failures in its own `RuntimeError`, so plain `{e}`
would surface that rather than the API server's message. It now reads through `__cause__`.
No message construction, nothing assumed about server-side error semantics.

Everything else is gone — `format_missing_permissions_error`, the RBAC constants, the
`_permission_error` helper, and their tests.

The remaining question is `sessions.rst`. It documents which verbs each operation needs, which
isn't written down anywhere today. Happy to drop it if it's out of scope, or keep it as the
docs half of what @andreyvelich was pointing at.

Separately: `sparkconnects` is missing from `kubeflow-spark-edit` in the manifests repo — only
`sparkapplications` and `scheduledsparkapplications` are listed — so `default-editor` can't
create a session on a stock install. Raising that against community-distribution.

andreyvelich: Why do we need to include Exception message in the return statement?
Since we propagate the entire traceback (`raise RuntimeError() from e`), you should be able to see the results with: `print(traceback.format_exc())`.
Also, if you just invoke `connect()` without `try except` , it should print the full traceback message.

vikas-saxena02: Closing this — the premise was wrong and I should have checked it earlier.

`raise ... from e` already chains the exception, so an uncaught `connect()` surfaces the full `ApiException` including the API server's message. What I was looking at was my own `try/except` printing `str(e)`, which drops the chain. That's a property of how I caught it, not of the SDK.

Thanks both for the review — the layering argument stands on its own, and it's a useful distinction to have had spelled out.

The real gap turned out to be in the manifests, as @tariq-hasan suggested: `sparkconnects` is missing from `kubeflow-spark-edit` and `kubeflow-spark-view`, so `default-editor` can't create a session on a stock install. Raised as kubeflow/community-distribution#3586

vikas-saxena02:
The real gap turned out to be in the manifests, as @tariq-hasan suggested: sparkconnects is missing from kubeflow-spark-edit and kubeflow-spark-view, so default-editor can't create a session on a stock install. Raised as https://github.com/kubeflow/community-distribution/pull/3586

PR #721, opened 2026-08-05
chore: expand API reference documentation for client classes

### LINKED ISSUE #260: Enhance API Reference

Expand the API reference documentation for all three clients `TrainerClient`, `OptimizerClient`, `ModelRegistryClient` to include complete method documentation with description, params, returns, return type, raises, example.

### Why is this needed?

The current API docs are auto-generated from docstrings and only show method signatures with parameter descriptions. Users need to see how methods are actually used in practice. Complete API documentation is essential for developer experience.

### Love this feature?

prakharsingh-74:
Description

This PR improves the API reference documentation for Kubeflow SDK clients by expanding docstrings with clearer descriptions, parameters, return types, error details, and practical usage examples.

These documentation-only changes improve developer experience without modifying any runtime behavior or public APIs.

Fixes: #260


prakharsingh-74:
@andreyvelich, if you're free, can you review this PR?

andreyvelich (commented):
/ok-to-test
/retest

kubeflow/trainer/api/trainer_client.py
```diff
@@ -224,11 +273,18 @@ def get_job_events(self, name: str) -> list[types.Event]:
...
-            RuntimeError: Failed to get a TrainJob events.
+            TimeoutError: Timeout occurred while getting the TrainJob events.
+            RuntimeError: Failed to get the TrainJob events.
+
+        Examples:
+            >>> from kubeflow.trainer import TrainerClient
+            >>> client = TrainerClient()
+            >>> events = client.get_job_events("s8d44aa4fb6d")
+            >>> for event in events:
+            ...     print(f"[{event.time}] {event.message}")
```
andreyvelich: `Event` has no `time` attribute — the field is `event_time` (`kubeflow/trainer/types/types.py:321`), so this example raises `AttributeError`.

```suggestion
            ...     print(f"[{event.event_time}] {event.message}")
```

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -243,10 +287,17 @@ def get_job_events(self, name: str) -> list[Event]:
...
-            RuntimeError: Failed to get an OptimizationJob events.
+            TimeoutError: Timeout occurred while getting the OptimizationJob events.
+            RuntimeError: Failed to get the OptimizationJob events.
+
+        Examples:
+            >>> from kubeflow.optimizer import OptimizerClient
+            >>> client = OptimizerClient()
+            >>> events = client.get_job_events("opt-12345")
+            >>> for event in events:
+            ...     print(f"[{event.time}] {event.message}")
```
andreyvelich: Same as in `trainer_client.py` — the field is `event_time`, not `time`.

```suggestion
            ...     print(f"[{event.event_time}] {event.message}")
```

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -65,23 +68,40 @@ def optimize(
...
+            ValueError: If input arguments are invalid.
+            TimeoutError: Timeout occurred while creating the Experiment.
+            RuntimeError: Failed to create the Experiment.
+
+        Examples:
+            >>> from kubeflow.trainer import TrainJobTemplate, CustomTrainer
+            >>> from kubeflow.optimizer import OptimizerClient, Search, TrialConfig
+            >>> def train_fn():
+            ...     pass
+            >>> template = TrainJobTemplate(runtime="torch-distributed", trainer=CustomTrainer(func=train_fn))
```
andreyvelich: This example would fail at runtime. The backend injects every `search_space` key into the trainer's `func_args` (`optimizer/backends/kubernetes/backend.py:114`) and the generated code calls `train_fn(**params)`, so a zero-parameter `train_fn` makes every Trial die with `TypeError`. The parameter names must match the search-space keys.

Also wraps the 110-char line below the 100 limit.

```suggestion
            >>> def train_fn(learning_rate, num_epochs):
            ...     pass
            >>> template = TrainJobTemplate(
            ...     runtime="torch-distributed", trainer=CustomTrainer(func=train_fn)
            ... )
```

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -65,23 +68,40 @@ def optimize(
...
-        """Create an OptimizationJob for hyperparameter tuning.
+        """Create and submit an OptimizationJob for hyperparameter tuning.
 
         Args:
             trial_template: The TrainJob template defining the training script.
             trial_config: Optional configuration to run Trials.
-            objectives: List of objectives to optimize.
             search_space: Dictionary mapping parameter names to Search specifications using
                 Search.uniform(), Search.loguniform(), Search.choice(), etc.
+            objectives: Optional list of objectives to optimize.
```
andreyvelich: Every neighbouring arg gained a documented default, but this one has the most consequential one: the backend defaults to `[Objective()]` (`backend.py:90`), i.e. minimize the metric named `loss`.

```suggestion
            objectives: Optional list of objectives to optimize. Defaults to minimizing the
                "loss" metric.
```

kubeflow/trainer/api/trainer_client.py
```diff
@@ -64,44 +66,71 @@ def __init__(
...
+
+        This method lists available runtimes, preferring namespace-scoped over
+        cluster-scoped runtimes in case of duplicate names.
 
         Returns:
-            A list of training runtimes from both namespace-scoped and
-            cluster-scoped resources. If duplicates exist, the
-            namespace-scoped runtime is preferred. Returns an empty list
-            if no runtimes are found.
+            list[types.Runtime]: A list of training runtimes from namespace-scoped
```
andreyvelich: Please drop the `Type:` prefix from `Returns:`/`Yields:` here and at the other 17 sites in this PR (`trainer_client.py` 98, 168, 198, 221, 251, 276, 312; `optimizer_client.py` 82, 118, 142, 173, 197, 235, 290; `model_registry_client.py` 138, 178, 204, 230, 256, 280, 305, 326, 343).

Three reasons:
- CLAUDE.md: *"Types go in function signatures, NOT in docstrings."*
- `docs/source/conf.py:86` sets `typehints_document_rtype = False` to deliberately suppress the duplicate "Return type" field, but `napoleon_use_rtype = True` (line 77) turns the docstring prefix back into an `:rtype:` — so the type reappears twice in the rendered page.
- It leaks internal module aliases: `list[types.Runtime]` renders literally, but `kubeflow.trainer.types` isn't a path users import.

Every other file in the repo (e.g. `kubeflow/spark/api/spark_client.py`) uses the untyped form.

```suggestion
            A list of training runtimes from namespace-scoped
```

kubeflow/trainer/api/trainer_client.py
```diff
@@ -64,44 +66,71 @@ def __init__(
...
 
         Args:
-            runtime: Reference to one of existing runtimes.
+            runtime: Reference to an existing training runtime.
 
-        Raises:
-            ValueError: Input arguments are invalid.
-            RuntimeError: Failed to get Runtime.
+        Returns:
+            None
```
andreyvelich: This is inaccurate: `LocalProcessBackend.get_runtime_packages` returns `local_runtime.trainer.packages`, a `list[str]` (`backends/localprocess/backend.py:69`) — only the Kubernetes and Container backends return `None`. It's also a type in a docstring. The real fix is a `-> None` annotation on the backend method, which is out of scope for a docs PR.

```suggestion
```

kubeflow/trainer/api/trainer_client.py
```diff
@@ -276,8 +338,16 @@ def delete_job(self, name: str):
         Args:
             name: Name of the TrainJob.
 
+        Returns:
+            None
+
```
andreyvelich: `Returns: None` renders as a content-free "Returns: None" field in the built docs. Better expressed as a `-> None` return annotation on the method (separate PR).

```suggestion
```

kubeflow/optimizer/api/optimizer_client.py
```diff
@@ -226,9 +262,17 @@ def delete_job(self, name: str):
         Args:
             name: Name of the OptimizationJob.
 
+        Returns:
+            None
+
```
andreyvelich: Same as `trainer_client.py:341` — please drop this block.

```suggestion
```

kubeflow/hub/api/model_registry_client.py
```diff
@@ -111,37 +113,44 @@ def register_model(
...
+            metadata: Additional metadata dictionary. Defaults to None.
+            storage_config: Storage credentials configuration for the model artifact.
+                See `StorageConfig` for details. Defaults to None.
 
         Returns:
-            Registered model.
+            RegisteredModel: The registered model object.
+
+        Raises:
+            Exception: If registration fails in the underlying registry store.
```
andreyvelich: Documenting bare `Exception` gives callers nothing to act on. The delegate raises `StoreError`, and every other method in this file already documents it by full name.

```suggestion
            model_registry.exceptions.StoreError: If registration fails in the underlying store.
```

andreyvelich (commented):
Thanks a lot for taking this on @prakharsingh-74 — the API reference genuinely needed this, and switching the examples away from markdown fences is a real improvement (those fences were emitting docutils warnings in the Sphinx build).

I've left inline suggestions on specific lines. Note that this review was produced with the help of AI tooling.

kubeflow/hub/api/model_registry_client.py
```diff
@@ -111,37 +113,44 @@ def register_model(
...
-        and has to be stored prior to registration.
+        """Register a model in the model registry.
 
-        Most models can be registered using their URI, along with an optional
-        `storage_config` describing how KServe should fetch the model at
-        inference time. URI builder utilities are recommended when referring to
-        specialized storage; for example `utils.s3_uri_from` when using S3
-        object storage data connections.
+        This registers a model version and its artifact in the registry. The model data
+        must be stored in remote storage (e.g., S3, GCS) before registration.
```
andreyvelich, line 119: The rewrite dropped the URI-builder guidance, which isn't recoverable from the signature — and the new example below hand-writes the exact raw `s3://` URI that this paragraph steered users away from. `model_registry.utils.s3_uri_from` is real and exists for this purpose.

```suggestion
        This registers a model version and its artifact in the registry. The model data
        must be stored in remote storage (e.g., S3, GCS) before registration.

        Most models can be registered using their URI, along with an optional
        `storage_config` describing how KServe should fetch the model at inference time.
        URI builder utilities are recommended when referring to specialized storage; for
        example `utils.s3_uri_from` when using S3 object storage data connections.
```

kubeflow/trainer/api/trainer_client.py
```diff
@@ -224,11 +273,18 @@ def get_job_events(self, name: str) -> list[types.Event]:
...
 
         Returns:
-            A list of Event objects associated with the TrainJob.
+            list[types.Event]: A list of Event objects associated with the TrainJob.
 
         Raises:
-            TimeoutError: Timeout to get a TrainJob events.
-            RuntimeError: Failed to get a TrainJob events.
+            TimeoutError: Timeout occurred while getting the TrainJob events.
+            RuntimeError: Failed to get the TrainJob events.
```
andreyvelich, line 279: Missing the exception a local-dev user is most likely to hit: both `LocalProcessBackend.get_job_events` (`backends/localprocess/backend.py:213`) and `ContainerBackend.get_job_events` (`backends/container/backend.py:807`) are `raise NotImplementedError()`.

```suggestion
            TimeoutError: Timeout occurred while getting the TrainJob events.
            RuntimeError: Failed to get the TrainJob events.
            NotImplementedError: If the client uses the LocalProcess or Container backend.
```

kubeflow/trainer/api/trainer_client.py
```diff
@@ -64,44 +66,71 @@ def __init__(
...
-            A runtime object. If both namespace-scoped and
-            cluster-scoped runtimes exist with the same name, the
-            namespace-scoped runtime is returned.
+            types.Runtime: A runtime object. If both namespace-scoped and
+            cluster-scoped runtimes exist with the same name, the namespace-scoped
+            runtime is returned.
+
+        Raises:
+            TimeoutError: Timeout occurred while retrieving the runtime.
+            RuntimeError: Failed to get the runtime.
```
andreyvelich, line 104: These two are correct for the Kubernetes backend, but the new block omits `ValueError`, which `LocalProcessBackend.get_runtime` raises for an unknown name (`backends/localprocess/backend.py:60`) and `ContainerBackend.get_runtime` propagates.

```suggestion
            TimeoutError: Timeout occurred while retrieving the runtime.
            RuntimeError: Failed to get the runtime.
            ValueError: If the runtime does not exist (LocalProcess and Container backends).
```

kubeflow/hub/api/model_registry_client.py
```diff
@@ -111,37 +113,44 @@ def register_model(
...
             author: Author of the model. Defaults to the client author.
             owner: Owner of the model. Defaults to the client author.
-            version_description: Description of the model version.
-            metadata: Additional version metadata.
-            storage_config: Storage credentials for the model artifact. Groups
-                           `storage_key`, `storage_path`, and
-                           `service_account_name` used by KServe's
-                           StorageInitializer. See `StorageConfig` for details.
+            version_description: Description of the model version. Defaults to None.
+            metadata: Additional metadata dictionary. Defaults to None.
```
andreyvelich: Losing "version" here matters: `metadata` is forwarded as `custom_properties` to the model *version*, not to the `RegisteredModel` this method returns.

```suggestion
            metadata: Additional model version metadata. Defaults to None.
```

kubeflow/hub/api/model_registry_client.py
```diff
@@ -179,17 +195,24 @@ def update_model(self, model: RegisteredModel) -> RegisteredModel:
...
 
         Returns:
-            Updated model version.
+            ModelVersion: The updated model version object.
 
         Raises:
-            TypeError: If model_version is not a ModelVersion instance.
-            model_registry.exceptions.StoreError: If model version does not have an ID.
+            TypeError: If the input is not a ModelVersion instance.
+            model_registry.exceptions.StoreError: If the model version does not have an ID or store fails.
```
andreyvelich: Over the 100-char limit (106). `E501` is in the ruff ignore list so CI won't catch these, but CLAUDE.md still specifies 100.

Same issue at `model_registry_client.py:234`, `optimizer_client.py:44` and `:230`, `trainer_client.py:163` and `:307`.

```suggestion
            model_registry.exceptions.StoreError: If the model version does not have an ID
                or the store operation fails.
```

andreyvelich (commented):
Thanks for this @prakharsingh-74!
/lgtm
/approve

PR #726, opened 2026-08-06
feat(optimizer): add trainer-native OptimizationJob backend

Sridhar1030:
Part of kubeflow/trainer#3794 (KEP-3562).

Adds a trainer-native backend for OptimizerClient that creates OptimizationJob CRs (trainer.kubeflow.org/v1alpha1, KEP-2605) instead of Katib Experiments. Selected via `TrainerNativeBackendConfig`; the Katib backend is untouched.

**Blocked on trainer PRs** (hence draft): CRD kubeflow/trainer#3552, controller kubeflow/trainer#3828, trial history kubeflow/trainer#3856. Unit tests (44) run against a mocked Kubernetes API; E2E impossible until the controller merges.

### DIFF
--- a/kubeflow/optimizer/__init__.py
+++ b/kubeflow/optimizer/__init__.py
@@ -18,6 +18,9 @@ from kubeflow.common.types import KubernetesBackendConfig
 # Import the Kubeflow Optimizer client.
 from kubeflow.optimizer.api.optimizer_client import OptimizerClient
 
+# Import the Kubeflow Optimizer backend configs.
+from kubeflow.optimizer.backends.trainer_native.types import TrainerNativeBackendConfig
+
 # Import the Kubeflow Optimizer types.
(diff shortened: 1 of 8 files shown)


andreyvelich (commented):
Thanks for this work @Sridhar1030!
Before integrating OptimizationJob into Kubeflow SDK, can we include Phase 2 to the @aniket2405's KEP with the client APIs: https://github.com/kubeflow/trainer/tree/master/proposals/2605-optimization-job-crd

We should talk about API design before implementing them.

/hold

Sridhar1030:
Thanks @andreyvelich, agreed on design first. I've opened kubeflow/trainer#3881 adding Phase 2 (client APIs) to the KEP. It proposes `TrainerClient().optimize()` rather than the separate backend config used in this PR. Keeping this on hold as the working prototype, I'll rework it once the design settles.

PR #727, opened 2026-08-06
fix(spark): stop PodTemplateOverride from mutating its own template

rjgoyln:

`PodTemplateOverride` writes back into the template dict it was handed, leaving an empty `containers` list behind. That list then travels: applying the same option to a second resource replaces the containers already declared there, so a sidecar disappears with nothing logged.

```
resource2 containers BEFORE: [{"name": "sidecar", "image": "fluentd:latest"}]
resource2 containers AFTER : []
```

Building an options list once and reusing it across sessions is what the docstring demonstrates, so this is reachable from the API's own examples.

Copying inside the merge keeps the guarantee in the code that would otherwise break it rather than in a caller who has to remember it. The trainer and optimizer backends already copy user input before mutating it for the same reason.

`Labels`, `Annotations`, `NodeSelector` and `Name` build a fresh dict and update into it, so none of them share the defect.

**Tests**:

Both new tests fail on `main` — the second on `assert [] == ['sidecar']`.

None; found by reading the code.

**Checklist:**

---

AI assistance disclosure: drafted with Claude Code (Opus 5); I reviewed the change and verified the reproduction and tests locally.


Goku2099:
- @rjgoyln `PodTemplateOverride` is being removed entirely in #743, so this fix won't have anything left to apply to.

/close

Issue #720, opened 2026-08-04
Refactor Spark options API to clearly distinguish SparkApplication and SparkConnect options


Goku2099:
## Summary

 The current Spark options API exposes options through a common entry point even though some options are only applicable to SparkConnect (for example PodTemplateOverride). This can make it difficult for users to determine which options are supported by submit_job() versus connect().

Refactor the options API and documentation to provide a clearer, more consistent experience while preserving existing functionality.

Ref - https://github.com/kubeflow/sdk/pull/685#pullrequestreview-4859037596

Goku2099:
cc @andreyvelich @tariq-hasan

Ebraam-Ashraf:
hi @Goku2099 @andreyvelich can i work on it ?

Goku2099:
Hi @Ebraam-Ashraf, thanks for your interest!
/assign @Ebraam-Ashraf

Goku2099:
Hi @Ebraam-Ashraf shraf, thanks for taking this up. This issue is important for our upcoming release. If possible, could you please prioritize it and get started soon? It would be really helpful if we could get this done at the earliest. Let me know if you need any clarification or help from my side.

Ebraam-Ashraf:
okay @Goku2099  I will finish it today

PR #730, opened 2026-08-07
feat(docs): Spark SDK docs for batch job submission and lifecycle APIs

Goku2099:
## What this PR does

`docs/source/spark/index.rst` covered interactive sessions well, but batch job submission and lifecycle management (submit_job(), FileJob, FuncJob, list_jobs/get_job/wait_for_job_status/get_job_logs/delete_job) are now implemented and had nowhere to go without turning index.rst into an unmaintainable wall of text.

This PR splits the Spark docs the same way `docs/source/train/` already splits Trainer's docs — a slim landing page plus one guide per capability — instead of cramming everything into a single file.

## New structure

spark/
├── index.rst # landing page: overview, "Two Ways to Run Spark", guide cards
├── sessions.rst # connect(), session management (moved out of index.rst, content unchanged)
├── batch-jobs.rst # submit_job(), FileJob, FuncJob
├── lifecycle.rst # SparkJobStatus, list/get/wait/logs/delete, common patterns
├── options.rst # Labels/Annotations/NodeSelector/Toleration/Name (shared by sessions + batch jobs)
└── api.rst # unchanged

## Notes on content

- `PodTemplateOverride` has been removed everywhere — it's no longer part of either `connect()` or `submit_job()`.
- Batch job examples use `https://` as the flagship `file_source` scheme (matches `examples/spark` and the batch-jobs blog post). `s3a://`, `gs://`, and `hdfs://` are documented as supported but not yet covered by a working example/notebook.
- `batch-jobs.rst` documents the real `FuncJob` constraints from `_validate_func_job`: top-level importable functions only (no lambdas, decorators, `async def`, or interactively-defined functions), imports must live inside the function body, and `func_args` is limited to JSON-like primitive values that must bind to the function's signature.
- `lifecycle.rst` calls out that `wait_for_job_status()` raises `RuntimeError` immediately if a job reaches `FAILED` and `FAILED` isn't in the target `status` set, and that `get_job_logs()` reads from the driver pod only (executor logs aren't wired in yet).
- `sessions.rst` documents the automatic `kubectl port-forward` behavior when `connect()` creates a session from outside the cluster.
- This PR also sets `toc_object_entries = False` in the Sphinx configuration. This is an intentional site-wide documentation change that disables per-object entries in the local table of contents for autodoc pages (Trainer, Spark, Pipelines, Hub, etc.). The goal is to keep API reference sidebars focused on the documentation structure instead of expanding every documented class and function.

## Testing

- [x] `make html` (or equivalent Sphinx build) — no broken `:doc:` references
- [x] Spot-checked code examples against `kubeflow/spark/client.py` and the
      Kubernetes backend implementation

 ### Blocker #720

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -136,6 +136,9 @@ html_title = "Kubeflow SDK"
 html_static_path = ["_static"]
 html_css_files = ["custom.css"]
 
+# Don't let every autodoc'd class/function bloat the sidebar's local TOC
+toc_object_entries = False
+
 # Furo theme options with top navigation
 html_theme_options = {
     "sidebar_hide_name": False,
--- a/docs/source/spark/api.rst
+++ b/docs/source/spark/api.rst
@@ -10,3 +10,19 @@ Spark Client
    :members:
    :undoc-members:
    :show-inheritance:
+
+Types
+-----
+
+.. automodule:: kubeflow.spark.types.types
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+Kubernetes Options
+-------------------
+
+.. automodule:: kubeflow.spark.options.kubernetes
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/spark/batch-jobs.rst
+++ b/docs/source/spark/batch-jobs.rst
@@ -104,7 +104,7 @@ This is passed through as ``spec.mainApplicationFile`` on the generated
 ``SparkApplication``:
 
 .. code-block:: yaml
-  
+
    mainApplicationFile: local:///opt/spark/app/etl.py
 
 Additional dependencies (Python packages, JARs, archives) are not configurable
@@ -124,7 +124,7 @@ through ``submit_job()`` yet — the SDK never sets ``spec.deps`` on the
      - ``str``
      - Local or remote path to the Spark application file (required)
    * - ``args``
-     - ``List[str]``
+     - ``list[str] | None``
      - Command-line arguments passed to the application
 
 Function Mode
--- a/docs/source/spark/batch-jobs.rst
+++ b/docs/source/spark/batch-jobs.rst
@@ -1,0 +1,260 @@
+Batch Jobs
+==========
+
+Submit existing Spark applications as managed, long-running Kubernetes workloads.
+
+Overview
+--------
+
+Batch jobs are built for production workflows — scheduled ETL, CI/CD pipelines, and
+anything that should run to completion without an interactive session attached.
+``submit_job()`` translates a job definition into a ``SparkApplication`` resource
+managed by the Spark Operator.
+
+.. code-block:: python
+
+   from kubeflow.spark import FileJob, SparkClient
+
+   client = SparkClient()
+
+   job_name = client.submit_job(
+       job=FileJob(
+           file_source="https://raw.githubusercontent.com/<repo>/<branch>/daily_pipeline.py",
+           args=["--date", "2024-01-15", "--output", "s3a://bucket/output/"],
+       ),
+       spark_conf={"spark.sql.adaptive.enabled": "true"},
+   )
+
+.. note::
+   The job script itself is fetched from ``https://raw.githubusercontent.com/...``
+   here — that's the URI scheme with a working, tested example today. The job's
+   *output* can still go anywhere Spark can write, including ``s3a://`` — see the
+   note on remote sources below for why the script source and the data paths are
+   different concerns.
+
+Once submitted, use the :doc:`lifecycle` APIs to monitor, wait on, and clean up the
+job.
+
+Two Submission Modes
+---------------------
+
+``submit_job()`` uses function overloading — the type of ``job`` you pass determines
+the mode:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 60
+
+   * - Job Type
+     - Mode
+     - Use Case
+   * - ``job=FileJob(...)``
+     - File mode
+     - Existing scripts, CI/CD pipelines
+   * - ``job=FuncJob(...)``
+     - Function mode
+     - Inline transformations, notebooks
+
+File Mode
+---------
+
+``FileJob`` submits an existing Spark application, referenced by a local or remote
+file source.
+
+.. code-block:: python
+
+   from kubeflow.spark import FileJob
+
+   job_name = client.submit_job(
+       job=FileJob(
+           file_source="https://raw.githubusercontent.com/<repo>/<branch>/daily_pipeline.py",
+           args=["--date", "2024-01-15"],
+       )
+   )
+
+**Remote sources** — ``file_source`` can point at ``s3a://``, ``gs://``,
+``hdfs://``, or ``https://``. The SDK doesn't validate or fetch the file itself —
+The SDK passes the URI through to the Spark runtime, which is responsible for resolving and accessing it:
+
+.. code-block:: python
+
+   client.submit_job(job=FileJob(file_source="https://raw.githubusercontent.com/<repo>/<branch>/etl.py"))
+
+.. note::
+   ``https://`` is the scheme with a working, tested example today (see
+   ``examples/spark`` in the ``kubeflow/sdk`` repo). ``s3a://``, ``gs://``, and
+   ``hdfs://`` are supported the same way in principle — the SDK just passes the
+   URI to Spark — but they haven't been verified end-to-end and may need
+   additional filesystem connector packages that aren't guaranteed to be
+   preinstalled in every Spark image. Until there's an example notebook
+   covering that setup, start with ``https://`` if you want something that's
+   known to work.
+
+**Local sources** — when ``file_source`` uses a local URI
+(``local:///opt/spark/app/etl.py``), the file must already be available to the
+``SparkApplication`` — for example through a mounted PersistentVolumeClaim (PVC) or
+a pre-built container image. SparkClient does not package or upload local files
+automatically:
+
+.. code-block:: python
+
+   client.submit_job(job=FileJob(file_source="local:///opt/spark/app/etl.py"))
+
+.. code-block:: yaml
+
+   mainApplicationFile: local:///opt/spark/app/etl.py
+
+Additional dependencies (Python packages, JARs, archives, and other resources) are
+managed using the Spark Operator's native dependency mechanisms (``deps.files``,
+``deps.pyFiles``, ``deps.jars``, etc.).
+
+``FileJob`` fields:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 60
+
+   * - Field
+     - Type
+     - Description
+   * - ``file_source``
+     - ``str``
+     - Local or remote path to the Spark application file (required)
+   * - ``args``
+     - ``List[str]``
+     - Command-line arguments passed to the application
+   * - ``main_class``
+     - ``str``
+     - Main class for JVM-based (Java/Scala) applications
+
+Function Mode
+-------------
+
+``FuncJob`` lets you submit a Python function directly, without authoring a
+standalone application file yourself.
+
+**In short:** write a normal Python function, put any imports it needs
+*inside* the function body, and pass it to ``FuncJob``. The SDK turns it into
+a runnable script for you. The details below explain exactly which functions
+qualify and why.
+
+.. code-block:: python
+
+   from kubeflow.spark import FuncJob
+
+   def etl_pipeline(date: str, output_path: str):
+       from pyspark.sql import SparkSession
+       import pyspark.sql.functions as F
+
+       spark = SparkSession.builder.getOrCreate()
+       df = spark.read.parquet(f"s3a://data/raw/{date}/")
+
+       result = (
+           df.filter(df.status == "valid")
+           .groupBy("category")
+           .agg(F.sum("amount").alias("total"))
+       )
+
+       result.write.parquet(output_path)
+
+   job_name = client.submit_job(
+       job=FuncJob(
+           func=etl_pipeline,
+           func_args={"date": "2024-01-15", "output_path": "s3a://data/processed/"},
+       )
+   )
+
+**Function requirements.** The SDK reads the function's source directly via
+``inspect.getsource()`` and serializes it — so ``func`` must be a plain,
+top-level function defined in an importable ``.py`` module. Lambdas, decorated
+functions, async functions, and anything defined interactively (a REPL or
+notebook cell) aren't supported — the SDK validates this at submit time and
+raises a ``ValueError`` with a specific reason if ``func`` doesn't qualify. If
+you're prototyping in a notebook, the function still needs to live in a module
+you import from, not be defined inline in a cell.
+
+**func_args must be simple, JSON-like values.** Keys must be strings, and
+values are limited to ``None``, ``str``, ``int``, ``bool``, finite floats, and
+lists/dicts built from those — no arbitrary Python objects (DataFrames,
+custom classes, open file handles, etc.), since ``func_args`` has to be
+serialized alongside the function. The SDK also checks that ``func_args``
+actually matches your function's signature before submitting, so a typo'd
+keyword argument fails fast locally instead of surfacing as a cryptic error
+inside the driver pod.
+
+**Imports must live inside the function body.** Only the function's source is
+serialized, not its surrounding module — so the function needs to be
+self-contained. Any imports it needs (``pyspark.sql``, ``pyspark.sql.functions``,
+etc.) have to be placed inside the function itself, as in the example above,
+so they're available when the generated script runs:
+
+.. code-block:: python
+
+   def transform():
+       from pyspark.sql import SparkSession  # required: import inside the function
+
+       spark = SparkSession.builder.getOrCreate()
+       # ...
+
+Under the hood, the SDK serializes the function into a generated script, writes
+it to a shared ``emptyDir`` volume through an init container, and points the
+driver at the generated file — this is the one case where the SDK does the
+packaging step for you. Once submitted, a ``FuncJob`` converges on the same
+``SparkJob`` shape as a ``FileJob``, so every :doc:`lifecycle` call treats them
+identically.
+
+``FuncJob`` fields:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 20 60
+
+   * - Field
+     - Type
+     - Description
+   * - ``func``
+     - ``Callable``
+     - Top-level, importable Python function (no lambdas, decorators, or
+       ``async def``) to run as the Spark application
+   * - ``func_args``
+     - ``dict``
+     - Keyword arguments passed to ``func``. String keys; values limited to
+       JSON-like primitives (``None``, ``str``, ``int``, ``bool``, finite
+       ``float``, or lists/dicts of the same) — must match ``func``'s
+       signature
+
+Resource and Spark Configuration
+----------------------------------
+
+Batch jobs use the same resource and Spark configuration model as interactive
+sessions:
+
+.. code-block:: python
+
+   client.submit_job(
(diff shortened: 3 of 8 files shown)


Goku2099:

andreyvelich (commented):
Thanks @Goku2099!
Can you also ensure to add the Spark pages to the website outline:

Image

docs/source/spark/index.rst
```diff
@@ -33,176 +37,114 @@ To use Spark with the Kubeflow SDK, install the Spark dependencies:
...
-   result.show()
-
-**Aggregate data:**
-
-.. code-block:: python
+.. grid:: 2
+   :gutter: 3
 
-   df = spark.range(100)
+   .. grid-item-card:: Options Reference
```
andreyvelich, line 125: The four new pages aren't in any toctree, so they don't appear in the sidebar. `sphinx-build` warns:

```
docs/source/spark/batch-jobs.rst: WARNING: document isn't included in any toctree
docs/source/spark/lifecycle.rst:  WARNING: document isn't included in any toctree
docs/source/spark/options.rst:    WARNING: document isn't included in any toctree
docs/source/spark/sessions.rst:   WARNING: document isn't included in any toctree
```

Add them to the Spark toctree in `docs/source/index.rst` (same as `train/*`):

```rst
   spark/index
   spark/sessions
   spark/batch-jobs
   spark/lifecycle
   spark/options
   spark/api
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+     - Description
+   * - ``file_source``
+     - ``str``
+     - Local or remote path to the Spark application file (required)
+   * - ``args``
+     - ``List[str]``
+     - Command-line arguments passed to the application
+   * - ``main_class``
+     - ``str``
+     - Main class for JVM-based (Java/Scala) applications
```
andreyvelich: `FileJob` only has `file_source` and `args` — there is no `main_class` field, and `get_spark_application_cr_from_file_job()` hardcodes `type="Python"`, so JVM apps aren't supported yet:

```suggestion
   * - ``args``
     - ``list[str] | None``
     - Command-line arguments passed to the application
```

docs/source/spark/sessions.rst
```diff
@@ -0,0 +1,236 @@
...
+=====================
+
+Connect to Spark interactively from a notebook or script using Spark Connect.
+
+Overview
+--------
+
+An interactive session gives you a standard PySpark ``SparkSession`` backed by a
+Spark Connect server running on Kubernetes. Sessions are self-managing: the SDK
+provisions the server and executors for you, and cleans them up when you call
```
andreyvelich: There is no auto-cleanup — no `atexit`/`__del__` hook and `stop()` isn't wrapped, so the `SparkConnect` resource outlives the process. Users must call `delete_session()`:

```suggestion
Spark Connect server running on Kubernetes. The SDK provisions the server and
executors for you, but does not delete them automatically — call
``client.delete_session(name)`` when you are done, otherwise the ``SparkConnect``
resource keeps running after your script exits.
```

docs/source/spark/options.rst
```diff
@@ -0,0 +1,119 @@
...
+
+.. code-block:: python
+
+   client.submit_job(
+       job=FileJob(file_source="https://raw.githubusercontent.com/<repo>/<branch>/etl.py"),
+       options=[Name("daily-etl-2026-06-18")],
+   )
+
+Composing Options
+------------------
```
andreyvelich, line 123: `PodTemplateOverride` is still exported from `kubeflow.spark` and still applied for `connect()` (see `examples/spark/spark_advanced_options.py`), so dropping it from the docs loses a supported option:

```suggestion
Pod Template Override
----------------------

Override the driver or executor pod template for custom volumes, sidecars, or
security contexts. Sessions only — passing it to ``submit_job()`` raises
``ValueError``:

.. code-block:: python

   from kubeflow.spark import PodTemplateOverride, SparkClient

   client = SparkClient()

   spark = client.connect(
       options=[
           PodTemplateOverride(
               role="executor",
               template={"spec": {"securityContext": {"runAsUser": 1000}}},
           ),
       ],
   )

Composing Options
------------------
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+
+   client.submit_job(job=FileJob(file_source="local:///opt/spark/app/etl.py"))
+
+.. code-block:: yaml
+
+   mainApplicationFile: local:///opt/spark/app/etl.py
+
+Additional dependencies (Python packages, JARs, archives, and other resources) are
+managed using the Spark Operator's native dependency mechanisms (``deps.files``,
+``deps.pyFiles``, ``deps.jars``, etc.).
```
andreyvelich: `submit_job()` has no `deps` parameter and neither CR builder sets `spec.deps`, so this reads as a supported path that isn't reachable from the SDK:

```suggestion
Additional dependencies (Python packages, JARs, archives) are not configurable
through ``submit_job()`` yet — the SDK never sets ``spec.deps`` on the
``SparkApplication``. Bake them into the Spark image or mount them via a PVC.
```

andreyvelich (commented):
Thanks for taking this on, @Goku2099 — splitting the Spark docs the way `train/` is split is the right call, and the lifecycle/`FuncJob` constraint write-ups are genuinely useful.

I left inline comments with suggestions where the docs drifted from the current `SparkClient` implementation. This review was done with the help of AI tools.

docs/source/spark/sessions.rst
```diff
@@ -0,0 +1,236 @@
...
+   )
+
+``spark_conf`` maps directly to Spark configuration properties and is applied when
+the session is created.
+
+**Per-role driver/executor overrides:** ``connect()`` also accepts ``driver``
+and ``executor`` objects for settings that apply specifically to the driver or
+to executors (rather than to the session as a whole). These are for advanced
+cases — most sessions only need ``num_executors``, ``resources_per_executor``,
+and ``spark_conf`` above. See :doc:`api` for the current field list.
```
andreyvelich, line 132: `api.rst` only has `automodule:: kubeflow.spark.api.spark_client`, so `Driver`/`Executor` fields aren't rendered there and this link is a dead end. Please add `automodule` entries for `kubeflow.spark.types.types` and `kubeflow.spark.options.kubernetes` to `docs/source/spark/api.rst` (that also fixes the `FileJob`/`FuncJob`/`SparkJob` references on the other pages).

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+   different concerns.
+
+Once submitted, use the :doc:`lifecycle` APIs to monitor, wait on, and clean up the
+job.
+
+Two Submission Modes
+---------------------
+
+``submit_job()`` uses function overloading — the type of ``job`` you pass determines
+the mode:
```
andreyvelich: `submit_job()` has no `@overload`s — `_validate_job()` does an `isinstance` check and raises `TypeError` otherwise:

```suggestion
``submit_job()`` dispatches on the type of ``job`` you pass — anything other than
``FileJob`` or ``FuncJob`` raises ``TypeError``:
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+
+   * - Job Type
+     - Mode
+     - Use Case
+   * - ``job=FileJob(...)``
+     - File mode
+     - Existing scripts, CI/CD pipelines
+   * - ``job=FuncJob(...)``
+     - Function mode
+     - Inline transformations, notebooks
```
andreyvelich: This contradicts the constraint documented below (and `_validate_func_job()`): functions defined in a notebook cell are rejected.

```suggestion
     - Python functions kept in an importable module
```

docs/source/spark/lifecycle.rst
```diff
@@ -0,0 +1,187 @@
...
+       FAILED = "Failed"
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - SDK Status
+     - SparkApplication States
+   * - ``CREATED``
+     - SUBMITTED
```
andreyvelich: `_SDK_STATE_BY_OPERATOR_STATE` also maps `""` (no state reported yet) to `CREATED`:

```suggestion
   * - ``CREATED``
     - SUBMITTED, or no state reported yet
```

docs/source/spark/lifecycle.rst
```diff
@@ -0,0 +1,187 @@
...
+   * - ``RUNNING``
+     - RUNNING, SUCCEEDING, SUSPENDING, SUSPENDED, RESUMING
+   * - ``COMPLETED``
+     - COMPLETED
+   * - ``FAILED``
+     - FAILED, SUBMISSION_FAILED, FAILING, PENDING_RERUN, INVALIDATING, UNKNOWN
+
+Additional status categories may be introduced in future releases.
```
andreyvelich: Worth stating the actual fallback from `SparkJobStatus.from_operator_state()`:

```suggestion
Any SparkApplication state the SDK doesn't recognize maps to ``FAILED`` and logs a
warning, so newly introduced operator states are handled conservatively.
```

docs/source/spark/options.rst
```diff
@@ -0,0 +1,119 @@
...
+       options=[
+           Toleration(key="spot-instance", operator="Exists", effect="NoSchedule"),
+       ],
+   )
+
+Custom Name
+-----------
+
+Set a custom session or job name via the ``Name`` option. If not specified, a name
+is auto-generated (``spark-connect-{uuid}`` for sessions):
```
andreyvelich: `generate_job_name()` uses a different prefix — worth naming it here:

```suggestion
is auto-generated (``spark-connect-{uuid}`` for sessions, ``spark-job-{uuid}`` for
batch jobs):
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+a pre-built container image. SparkClient does not package or upload local files
+automatically:
+
+.. code-block:: python
+
+   client.submit_job(job=FileJob(file_source="local:///opt/spark/app/etl.py"))
+
+.. code-block:: yaml
+
+   mainApplicationFile: local:///opt/spark/app/etl.py
```
andreyvelich, line 108: This YAML block has no lead-in — add one so it's clear what it shows:

```suggestion
This is passed through as ``spec.mainApplicationFile`` on the generated
``SparkApplication``:

.. code-block:: yaml

   mainApplicationFile: local:///opt/spark/app/etl.py
```

docs/source/spark/options.rst
andreyvelich: No need to add `PodTemplateOverride` to the docs, since we are going to remove it soon: https://github.com/kubeflow/sdk/issues/720

Goku2099:
@andreyvelich I have addressed all of the review comments. Could you please take another look when you have a chance?

docs/source/spark/sessions.rst
```diff
@@ -0,0 +1,236 @@
+Interactive Sessions
+=====================
+
+Connect to Spark interactively from a notebook or script using Spark Connect.
+
+Overview
+--------
+
+Spark Connect server running on Kubernetes. The SDK provisions the server and
```
andreyvelich, line 10: My earlier suggestion was anchored one line short, so applying it swallowed the opening sentence — the paragraph now starts mid-thought. Restoring it:

```suggestion
An interactive session gives you a standard PySpark ``SparkSession`` backed by a
Spark Connect server running on Kubernetes. The SDK provisions the server and
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+
+.. code-block:: python
+
+   client.submit_job(job=FileJob(file_source="local:///opt/spark/app/etl.py"))
+
+This is passed through as ``spec.mainApplicationFile`` on the generated
+``SparkApplication``:
+
+.. code-block:: yaml
+
```
andreyvelich: This line has two trailing spaces, which fails the `trailing-whitespace` pre-commit hook:

```suggestion

```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+   :widths: 20 20 60
+
+   * - Field
+     - Type
+     - Description
+   * - ``file_source``
+     - ``str``
+     - Local or remote path to the Spark application file (required)
+   * - ``args``
+     - ``List[str]``
```
andreyvelich: `FileJob.args` is `list[str] | None`, and the repo targets 3.10 with lowercase builtins:

```suggestion
     - ``list[str] | None``
```

docs/source/spark/api.rst
```diff
@@ -10,3 +10,19 @@ Spark Client
...
+
+.. automodule:: kubeflow.spark.types.types
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+Kubernetes Options
+-------------------
+
+.. automodule:: kubeflow.spark.options.kubernetes
```
andreyvelich, line 25: Thanks — this is the right fix. Heads-up that it also surfaces pre-existing malformed docstrings: the docs build goes from 26 to 39 warnings, including two `ERROR: Unexpected indentation` from the `PodTemplateOverride` docstring, plus `Definition list ends without a blank line` from `Driver`, `Executor`, `Annotations`, `NodeSelector`, `Toleration`, and `Name`. Non-fatal, so CI stays green. Fixing those docstrings in `kubeflow/spark/types/types.py` and `kubeflow/spark/options/kubernetes.py` is arguably out of scope here — a follow-up issue is fine.

docs/source/conf.py
```diff
@@ -136,6 +136,9 @@ def setup(app):
 html_static_path = ["_static"]
 html_css_files = ["custom.css"]
 
+# Don't let every autodoc'd class/function bloat the sidebar's local TOC
+toc_object_entries = False
```
andreyvelich, line 140: This is a site-wide change — it removes per-object entries from the local TOC on every autodoc page (Trainer, Optimizer, Hub, Pipelines), not just Spark. If the goal is only to keep the new Spark API page's sidebar manageable, could you either call that out explicitly in the PR description so it gets a deliberate review, or drop it and handle sidebar noise separately?

andreyvelich (commented):
Thanks for turning these around so quickly, @Goku2099 — the toctree fix is confirmed (the four `isn't included in any toctree` warnings are gone in a local build) and the rest of the content fixes look right.

A few follow-ups inline, including one regression that's my fault: the suggestion I anchored on `sessions.rst` was one line short and ate the opening sentence. This review was done with the help of AI tools.

docs/source/conf.py
andreyvelich: @Goku2099 Any reason you want to add it?

Goku2099: Good point. I added toc_object_entries = False to keep the new Spark API page manageable after adding the requested automodule entries. Since this changes the local TOC behavior for all autodoc pages, I'll update the PR description to explicitly call it out so it gets a deliberate review.

docs/source/spark/api.rst
Goku2099: Thanks for the heads-up! I agree this is outside the scope of the current PR. I've created a follow-up issue https://github.com/kubeflow/sdk/issues/736 to track fixing the malformed docstrings and the associated Sphinx warnings.

andreyvelich (commented):
Looks great, thank you @Goku2099!
/lgtm
/assign @Shekharrajak @tariq-hasan @RobuRishabh 
/cc @nabuskey @vara-bonthu

docs/source/spark/options.rst
```diff
@@ -0,0 +1,120 @@
...
+
+   client = SparkClient()
+
+   spark = client.connect(
+       num_executors=5,
+       resources_per_executor={"cpu": "4", "memory": "16Gi", "nvidia.com/gpu": "1"},
+       options=[
+           NodeSelector({"node-type": "spark-gpu", "workload": "ml"}),
+       ],
+   )
```
tariq-hasan: We can't currently leverage GPUs with the `connec API`.

```suggestion
   spark = client.connect(
       num_executors=5,
       resources_per_executor={"cpu": "4", "memory": "16Gi"},
       options=[
           NodeSelector({"node-type": "spark-gpu", "workload": "ml"}),
       ],
   )
```

docs/source/spark/index.rst
tariq-hasan, line 1: I'm thinking it would be good to mention the default resources - 1 CPU and 512Mi - since the docs do not provide any other info about them.

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+This is passed through as ``spec.mainApplicationFile`` on the generated
+``SparkApplication``:
+
+.. code-block:: yaml
+
+   mainApplicationFile: local:///opt/spark/app/etl.py
+
+Additional dependencies (Python packages, JARs, archives) are not configurable
+through ``submit_job()`` yet — the SDK never sets ``spec.deps`` on the
+``SparkApplication``. Bake them into the Spark image or mount them via a PVC.
```
tariq-hasan: Both the image baking and PVC mounting are actually out of scope for the Spark SDK. I suppose this is from the operator perspective but the docs are targeted for a SDK user and such a user cannot leverage either option directly from the Spark SDK.

```suggestion
```

docs/source/spark/batch-jobs.rst
```diff
@@ -0,0 +1,260 @@
...
+
+.. note::
+   ``https://`` is the scheme with a working, tested example today (see
+   ``examples/spark`` in the ``kubeflow/sdk`` repo). ``s3a://``, ``gs://``, and
+   ``hdfs://`` are supported the same way in principle — the SDK just passes the
+   URI to Spark — but they haven't been verified end-to-end and may need
+   additional filesystem connector packages that aren't guaranteed to be
+   preinstalled in every Spark image. Until there's an example notebook
+   covering that setup, start with ``https://`` if you want something that's
+   known to work.
```
tariq-hasan, line 91: I think we'd want to update this as well - https://github.com/Goku2099/sdk/blob/1c840f7ea2f6121119a889b735e36d46e6e54f77/examples/spark/README.md - with the batch job examples but it could be a separate PR.

docs/source/spark/index.rst
```diff
@@ -33,176 +37,114 @@ To use Spark with the Kubeflow SDK, install the Spark dependencies:
 
 For full setup instructions, see `the Spark installation guide <https://www.kubeflow.org/docs/components/spark-operator/getting-started/>`_.
```
tariq-hasan, line 39: I think probably a good idea to note as a prerequisite that the `spark-operator-spark` service account must exist in the target namespace and that the `SparkApplication` RBAC must be bound to it or that every `submit_job()` fails - especially given that this requirement is critical at this time until https://github.com/kubeflow/spark-operator/issues/3049 is fixed.

docs/source/spark/options.rst
tariq-hasan: We'd want to rewrite the node selection section as we can't currently leverage GPUs with the `connect` API.

tariq-hasan (commented):
Hi @Goku2099! I just have a few comments. Otherwise lgtm. Thanks!

docs/source/spark/batch-jobs.rst
Goku2099: I agree. To keep this PR focused, I created a tracking issue for updating the Spark examples README with the batch job examples: #737 . I'll address it in a follow-up PR.

Goku2099:
@tariq-hasan  Thanks for the review! I've addressed all the review comments and pushed the requested changes. Could you please take another look when you have a chance?

tariq-hasan:
Thanks @Goku2099!
/lgtm

andreyvelich (commented):
/lgtm
/approve

PR #603, opened 2026-07-17
fix(spark): validate hostname is present in Spark Connect URL

Ishan-AI-coder:
Fixes #598 

## What this PR does

`validate_spark_connect_url()` in `kubeflow/spark/backends/kubernetes/utils.py`
checked the URL scheme and port, but never checked that a hostname was
present. This meant a malformed URL like `sc://:15002` passed validation
and only failed later, further down the connection path, with a much
less clear error.

This PR adds a `parsed.hostname` check alongside the existing port check.

## Change

```python
parsed = urlparse(url)
if parsed.scheme != "sc":
    raise ValueError(f"Invalid scheme '{parsed.scheme}'. Expected 'sc://'")
if not parsed.hostname:
    raise ValueError("Host is required in Spark Connect URL")
if not parsed.port:
    raise ValueError("Port is required in Spark Connect URL")
return True
```

## Testing

- Added `test_missing_hostname` to `TestValidateSparkConnectUrl` in
  `utils_test.py`, following the existing `test_missing_port` pattern.
- Verified `parsed.hostname` correctly distinguishes the empty-host case
  from all legitimate hostname shapes (plain hosts, FQDNs, IPv6 literals
  like `sc://[::1]:15002`) — none of these are affected by this change.
- Ran the full local test suite:
  - `utils_test.py`: 35 passed
  - `backend_test.py`: 33 passed (no regressions)

## Checklist

- [x] Tests added/updated
- [x] DCO sign-off on commit
- [x] Followed existing code style in the file

PR #732, opened 2026-08-07
fix: [KEP-107] reject Spark Connect URLs with missing hostname

### LINKED ISSUE #598: validate_spark_connect_url accepts URLs with a missing hostname (e.g. sc://:15002)
###  Description
`validate_spark_connect_url()` in `kubeflow/spark/backends/kubernetes/utils.py` only checks that the scheme is `sc` and that a port is present. It does not check that a hostname is present, so a malformed URL with an empty host passes validation and fails later with a less clear error further down the connection path. ## Steps to Reproduce ```python from kubeflow.spark.backends.kubernetes.utils import validate_spark_connect_url print(validate_spark_connect_url("sc://:15002")) # -> True, but no host given ```

### Expected Behavior
 `ValueError` should be raised for a URL with no hostname, similar to how a missing port is already handled.

### Actual Behavior 
Returns `True` for `sc://:15002`, deferring the failure to wherever the URL is actually used to connect.

### Suggested Fix 
Add a check for `parsed.hostname` alongside the existing `parsed.port` check: ```python if not parsed.hostname: raise ValueError("Host is required in Spark Connect URL") ```

### Environment 
- Repository: `kubeflow/sdk` 
- Branch: `main` 
- Component: `SparkClient` (`kubeflow.spark`)

### Impacted by this bug?

arntri:

`validate_spark_connect_url` only checked the URL scheme and port, so a URL with an empty hostname such as `sc://:15002` passed validation. Such a URL is not usable as a Spark Connect endpoint, and the failure surfaced later and less clearly at connection time.

This adds a hostname check between the existing scheme and port checks, so the error is raised at validation time with a message consistent with the surrounding checks. A `missing hostname` test case was added to `test_validate_spark_connect_url`.

Fixes #598

### DIFF
--- a/kubeflow/spark/backends/kubernetes/utils.py
+++ b/kubeflow/spark/backends/kubernetes/utils.py
@@ -394,6 +394,8 @@ def validate_spark_connect_url(url: str) -> bool:
     parsed = urlparse(url)
     if parsed.scheme != "sc":
         raise ValueError(f"Invalid scheme '{parsed.scheme}'. Expected 'sc://'")
+    if not parsed.hostname:
+        raise ValueError("Hostname is required in Spark Connect URL")
     if not parsed.port:
         raise ValueError("Port is required in Spark Connect URL")
     return True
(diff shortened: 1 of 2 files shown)


arntri:
Apologies for the duplicate. I missed that #603 already addresses this issue.
Closing in favour of that PR.

PR #742, opened 2026-08-08
fix(trainer): validate TorchTune loss and dataset source enums

### LINKED ISSUE #741: TorchTune loss and dataset source raise AttributeError instead of ValueError when not an enum
### What happened?

In the TorchTune fine-tuning path, passing a plain string where a `Loss` or `DataFormat`
enum is expected raises `AttributeError` from inside the SDK instead of the `ValueError`
that the surrounding code clearly intends. `dtype` in the same function already behaves
correctly, so the three fields are inconsistent.

```python
from kubeflow.trainer.backends.kubernetes import utils
from kubeflow.trainer.types import types

# dtype: guarded, reports the problem clearly
utils.get_args_using_torchtune_config(types.TorchTuneConfig(dtype="bf16"))
# ValueError: Invalid dtype: bf16.

# loss: no guard at all
utils.get_args_using_torchtune_config(
    types.TorchTuneConfig(loss="torchtune.modules.loss.CEWithChunkedOutputLoss")
)
# AttributeError: 'str' object has no attribute 'value'

# dataset source: guard exists but its own error message crashes
utils.get_args_from_dataset_preprocess_config(
    types.TorchTuneInstructDataset(source="json")
)
# AttributeError: 'str' object has no attribute 'value'
```

**1. `loss` is never type-checked.** `kubeflow/trainer/backends/kubernetes/utils.py`
compares as follows:

```python
# dtype - guarded
if fine_tuning_config.dtype:
    if not isinstance(fine_tuning_config.dtype, types.DataType):
        raise ValueError(f"Invalid dtype: {fine_tuning_config.dtype}.")
    args.append(f"dtype={fine_tuning_config.dtype.value}")

# loss - not guarded
if fine_tuning_config.loss:
    args.append(f"loss={fine_tuning_config.loss.value}")
```

**2. The `ValueError` for an invalid dataset source is unreachable.** In
`get_args_from_dataset_preprocess_config`:

```python
if dataset_preprocess_config.source:
    if not isinstance(dataset_preprocess_config.source, types.DataFormat):
        raise ValueError(f"Invalid data format: {dataset_preprocess_config.source.value}.")
```

The branch is entered only when `source` is *not* a `DataFormat`, but the message then
evaluates `.value` on that non-enum object. Building the exception raises `AttributeError`
first, so the intended `ValueError` can never be raised.

This is reproducible in pure Python without a cluster, since it only involves the SDK's
argument-building logic.

Both cases should raise `ValueError` at the point where the configuration is converted to
trainer args, matching the existing `dtype` behaviour, so the user sees which field is
wrong instead of an `AttributeError` from inside the SDK.

- `TorchTuneConfig(loss="...")` -> `ValueError: Invalid loss: ...`
- `TorchTuneInstructDataset(source="json")` -> `ValueError: Invalid data format: json.`

Passing a string instead of the enum member is an easy mistake to make when writing a
fine-tuning configuration by hand, and `DataFormat.JSON.value` is literally `"json"`, so
the invalid value often looks correct at a glance.

### Environment

Kubernetes version:
```bash
$ kubectl version
Not applicable - reproducible without a cluster.
```

Kubeflow Trainer version:
```bash
Not applicable - reproducible without a cluster.
```

Kubeflow Python SDK version:
```bash
$ pip show kubeflow
Version: 0.5.0rc0
```
Also present on `main` at commit 97ec647.

### Impacted by this bug?

Yigtwxx:

`TorchTuneConfig.loss` and `TorchTuneInstructDataset.source` both raised `AttributeError`
from inside the SDK when they were given a plain string instead of an enum member, even
though `dtype` in the same code path already raised a clear `ValueError`. This PR makes
all three consistent.

Two distinct causes, both in `kubeflow/trainer/backends/kubernetes/utils.py`:

1. `loss` had no type check at all in `get_args_using_torchtune_config`, so
   `loss.value` was evaluated directly on whatever was passed in.
2. In `get_args_from_dataset_preprocess_config` the `ValueError` for an invalid
   `source` was unreachable: the branch is entered only when `source` is *not* a
   `DataFormat`, but the message then evaluated `.value` on that non-enum object, so
   constructing the exception raised `AttributeError` before it could be raised.

Behaviour, reproduced with a small script against `main` (`97ec647`) and after the fix:

| Input | Before | After |
| --- | --- | --- |
| `TorchTuneConfig(dtype="bf16")` | `ValueError: Invalid dtype: bf16.` | unchanged |
| `TorchTuneConfig(loss="torchtune.modules.loss.CEWithChunkedOutputLoss")` | `AttributeError: 'str' object has no attribute 'value'` | `ValueError: Invalid loss: torchtune.modules.loss.CEWithChunkedOutputLoss.` |
| `TorchTuneInstructDataset(source="json")` | `AttributeError: 'str' object has no attribute 'value'` | `ValueError: Invalid data format: json.` |

Passing a bare string is an easy mistake when a fine-tuning configuration is written by
hand, and `DataFormat.JSON.value` is literally `"json"`, so the invalid value tends to
look correct at a glance.

Scope is deliberately limited to type validation. No value-range or membership checks are
added, no public signature changes, and the only behaviour that changes is on inputs that
already failed.

Tests: three cases added to the existing parametrized tests in `utils_test.py`, following
the `TestCase` pattern already used there.

- `invalid loss raises ValueError`
- `invalid source raises ValueError`
- `source is serialized by value` — `get_args_from_dataset_preprocess_config` previously
  had no coverage for `source` at all, which is why the unreachable branch went unnoticed.

Both failure cases fail on `main` with `assert <class 'AttributeError'> is <class
'ValueError'>` and pass with the fix, so they are genuine regression tests.

Verification run locally: `uv run pytest -q kubeflow` (691 passed), `uv lock --check`,
`uv run ruff check .`, `uv run ruff format --check kubeflow`, `uv run ty check kubeflow/hub`,
and `uv run pre-commit run` on the changed files — all clean.

Fixes #741

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/kubernetes/utils.py
+++ b/kubeflow/trainer/backends/kubernetes/utils.py
@@ -512,6 +512,9 @@ def get_args_using_torchtune_config(
 
     # Override the loss if it is provided.
     if fine_tuning_config.loss:
+        if not isinstance(fine_tuning_config.loss, types.Loss):
+            raise ValueError(f"Invalid loss: {fine_tuning_config.loss}.")
+
         args.append(f"loss={fine_tuning_config.loss.value}")
 
     # Override the data dir or data files if it is provided.
@@ -594,7 +597,7 @@ def get_args_from_dataset_preprocess_config(
     # Override the dataset source field if it is provided.
     if dataset_preprocess_config.source:
         if not isinstance(dataset_preprocess_config.source, types.DataFormat):
-            raise ValueError(f"Invalid data format: {dataset_preprocess_config.source.value}.")
+            raise ValueError(f"Invalid data format: {dataset_preprocess_config.source}.")
 
         args.append(f"dataset.source={dataset_preprocess_config.source.value}")
 
--- a/kubeflow/trainer/backends/kubernetes/utils_test.py
+++ b/kubeflow/trainer/backends/kubernetes/utils_test.py
@@ -858,6 +858,29 @@ def test_get_args_from_peft_config(test_case: TestCase):
                 "dataset.train_on_input=True",
             ],
         ),
+        TestCase(
+            name="source is serialized by value",
+            expected_status=SUCCESS,
+            config={
+                "dataset_preprocess_config": types.TorchTuneInstructDataset(
+                    source=types.DataFormat.JSON,
+                ),
+            },
+            expected_output=[
+                f"dataset={constants.TORCH_TUNE_INSTRUCT_DATASET}",
+                "dataset.source=json",
+            ],
+        ),
+        TestCase(
+            name="invalid source raises ValueError",
+            expected_status=FAILED,
+            config={
+                "dataset_preprocess_config": types.TorchTuneInstructDataset(
+                    source="json",
+                ),
+            },
+            expected_error=ValueError,
+        ),
     ],
 )
 def test_get_args_from_dataset_preprocess_config(test_case: TestCase):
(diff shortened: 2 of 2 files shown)


Yigtwxx:
Friendly ping — this has been open since Aug 8 with all checks green and no review yet.
The diff is small (type validation only, no behaviour change on valid input) and the
linked issue #741 has just been picked up by another contributor, so getting it reviewed
would also avoid duplicated work there.

@astefanutti @Electronic-Waste @szaher whenever you have a moment, thanks!

PR #658, opened 2026-07-26
chore(spark): add SparkClient API delegation coverage

lntutor:
## Summary

- add unit coverage for `SparkClient.connect()` against existing Spark Connect servers, including URL validation and token propagation
- cover create-mode forwarding of session configuration to the Kubernetes backend
- cover public session-management delegation, including `delete_session()` returning `None` and `get_session_logs()` forwarding `follow` by keyword

Part of #609. This is intentionally limited to the Spark API layer; it does not overlap with #610, #614, or #621.

## Validation

- `uv run pytest -q kubeflow/spark/api/spark_client_test.py` — 15 passed
- `uv run ruff check kubeflow/spark/api/spark_client_test.py`
- `uv run ruff format --check kubeflow/spark/api/spark_client_test.py`
- `make verify`

I also ran `make test-python`; it entered the unrelated Trainer container-backend test section and did not complete locally, so I stopped that single runner rather than leave it consuming resources. The focused Spark suite is clean and CI will run the repository gate.

PR #750, opened 2026-08-10
chore(spark): expand unit test coverage for SparkClient job methods

Neha-Nimsatkar:

Adds proper unit test coverage for the `SparkClient` job management methods in `kubeflow/spark/api/spark_client_test.py` — `submit_job`, `get_job`, `list_jobs`, `delete_job`, `wait_for_job_status`, and `get_job_logs`.

Tests now follow the same `@pytest.mark.parametrize` + `TestCase` dataclass pattern already used in `backend_test.py`, so we get consistent coverage of backend delegation and input validation across the client.

## Motivation

A few of the public `SparkClient` methods didn't have solid tests around argument forwarding, type contracts, or validation before hitting the backend. This was making it easy for delegation bugs to slip through unnoticed. This PR closes that gap and brings the test style in line with the rest of the codebase.

## What changed

- **General**: Rewrote the test suites to use `@pytest.mark.parametrize` with the `TestCase` dataclass — prints test names and checks against `expected_output`.
- **`test_submit_job`**: Combined validation and submission scenarios into one parametrized test. Added checks that `num_executors`, `resources_per_executor`, and `spark_conf` get forwarded correctly, plus `backend.submit_job.assert_not_called()` for invalid input.
- **`test_get_job`**: Covers job lookup delegation and the `RuntimeError` raised when a job isn't found.
- **`test_list_jobs`**: Fixed status filtering to expect sets (`{SparkJobStatus.RUNNING}`, `{SparkJobStatus.COMPLETED, SparkJobStatus.FAILED}`), plus a failure case test.
- **`test_delete_job`**: Covers deletion delegation and the nonexistent-job error path.
- **`test_wait_for_job_status`**:
  - Mock now returns a `SparkJob` instead of a bool.
  - Added `backend.wait_for_job_status.assert_not_called()` check for validation failures.
  - Extended matrix: default vs. custom timeouts, default vs. explicit polling intervals, single vs. multiple status sets, timeout edge cases (`< 0`, `= 0`), polling interval edge cases (`<= 0`, `>= timeout`).
- **`test_get_job_logs`**: Added cases for default `follow`, plus explicit `follow=True` / `follow=False`.

## Verification

- `uv run pytest kubeflow/spark/api/spark_client_test.py -v -s` — all passing
- `uv run ruff format kubeflow/spark/api/spark_client_test.py`
- `uv run ruff check --fix kubeflow/spark/api/spark_client_test.py`
- `uv run pre-commit run --all-files`
- `make verify`

## Fixes

Fixes #658

## Checklist

- [x] Tested locally with pytest, verified with `make verify`
- [x] Formatted with Ruff, passes pre-commit
- [ ] Docs — N/A, this is test-only


kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +174,73 @@ def test_submit_job_success(job, options):
             spark_conf=None,
             options=options,
         )
+
+
+def test_get_job():
```
tariq-hasan: The new tests should use the `@pytest.mark.parametrize` marker with the `TestCase` dataclass from `kubeflow.spark.test.common` - similar to [`backend_test.py`](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend_test.py) and [`utils_test.py`](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/utils_test.py).

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/test/common.py#L33-L42

tariq-hasan (changes requested):

kubeflow/spark/api/spark_client_test.py
Neha-Nimsatkar: Thanks for the pointer @tariq-hasan! I've refactored all new job management test methods (get_job, list_jobs, delete_job, wait_for_job_status, and get_job_logs) to use @pytest.mark.parametrize with the TestCase dataclass from kubeflow.spark.test.common.

All unit tests, ruff linting, and pre-commit checks are passing cleanly. Please have a look!

Neha-Nimsatkar:
Hi @tariq-hasan, I've updated the tests to use @pytest.mark.parametrize with the TestCase dataclass pattern from kubeflow.spark.test.common, and all unit tests and pre-commit hooks are passing locally.

The E2E Test matrix hit an infrastructure disconnect across the self-hosted runners ("self-hosted runner lost communication with the server"). Could you please click "Re-run failed jobs" when you get a chance? Thanks!

tariq-hasan:
/retitle chore(spark): expand unit test coverage for SparkClient job methods

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
             spark_conf=None,
             options=options,
         )
```
tariq-hasan, line 173: Please combine `test_submit_job_validation` and `test_submit_job_success` into a single `test_submit_job` with the `pytest.mark.parametrize` marker and the `TestCase` dataclass usage.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
...
+            config={
+                "job_name": "spark-job-123",
+                "status": "Completed",
+                "timeout": 60,
+                "polling_interval": 2,
+            },
+        ),
+    ],
+)
+def test_wait_for_job_status(test_case: TestCase):
```
tariq-hasan, line 325: For `test_wait_for_job_status` please test this behavior as well.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/api/spark_client.py#L285-L288

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
...
         )
+
+
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="get job by name",
+            expected_status=SUCCESS,
+            config={"job_name": "spark-job-123"},
```
tariq-hasan, line 182: Please ensure `name` and `expected_status` get used as well - otherwise they would be redundant. The same applies for other tests as well.

Also unhappy paths where errors are raised should be tested in addition to happy paths.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
...
+    [
+        TestCase(
+            name="list all jobs",
+            expected_status=SUCCESS,
+            config={"status": None},
+        ),
+        TestCase(
+            name="list jobs filtered by status",
+            expected_status=SUCCESS,
+            config={"status": "Running"},
```
tariq-hasan: We should use `SparkJobStatus.RUNNING`.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
...
+    """Test wait_for_job_status delegation to KubernetesBackend."""
+    with patch("kubeflow.spark.api.spark_client.KubernetesBackend") as mock_backend:
+        backend = mock_backend.return_value
+        backend.wait_for_job_status.return_value = True
+
+        client = SparkClient()
+        result = client.wait_for_job_status(
+            name=test_case.config["job_name"],
+            status=test_case.config["status"],
+            timeout=test_case.config["timeout"],
```
tariq-hasan: We're not really using polling interval here but the test case has polling interval so the test passes by accident. We need to test both default and overrides.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +173,158 @@ def test_submit_job_success(job, options):
...
+
+@pytest.mark.parametrize(
+    "test_case",
+    [
+        TestCase(
+            name="wait for job status success",
+            expected_status=SUCCESS,
+            config={
+                "job_name": "spark-job-123",
+                "status": "Completed",
```
tariq-hasan: We should use enum here - `SparkJobStatus.COMPLETED`.

tariq-hasan (changes requested):

Neha-Nimsatkar:
Thanks for the detailed review, @tariq-hasan!

I have addressed all the feedback and updated the test suite:

* Used `SparkJobStatus` enums (`RUNNING` and `COMPLETED`) instead of hardcoded strings.
* Added assertions for `expected_status` and error handling across test cases so `name` and status definitions are actively validated.
* Added unhappy path coverage for `get_job`, `list_jobs`, and `delete_job`.
* Updated `wait_for_job_status` to explicitly test both default (`2s`) and overridden polling intervals, as well as parameter validation for invalid timeout values.
* Verified all 21 unit tests, Ruff linting/formatting, and pre-commit checks pass locally.

kubeflow/spark/api/spark_client_test.py
tariq-hasan: Let's address this as well.

In addition please add a case with all of num_executors, resources_per_executor and spark_conf set, asserting each reaches the backend.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+    [
+        TestCase(
+            name="list all jobs",
+            expected_status=SUCCESS,
+            config={"status": None},
+        ),
+        TestCase(
+            name="list jobs filtered by status",
+            expected_status=SUCCESS,
+            config={"status": SparkJobStatus.RUNNING},
```
tariq-hasan: Let's fix the type contracts everywhere in the tests. For example this one should be

```suggestion
            config={"status": {SparkJobStatus.RUNNING}},
```

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+            },
+            expected_error=ValueError,
+        ),
+    ],
+)
+def test_wait_for_job_status(test_case: TestCase):
+    """Test wait_for_job_status delegation and validation."""
+    with patch("kubeflow.spark.api.spark_client.KubernetesBackend") as mock_backend:
+        backend = mock_backend.return_value
+        backend.wait_for_job_status.return_value = True
```
tariq-hasan: `backend.wait_for_job_status.return_value` should return a mock SparkJob not a boolean.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+            expected_status=FAILED,
+            config={
+                "job_name": "spark-job-123",
+                "status": SparkJobStatus.COMPLETED,
+                "timeout": -1,
+            },
+            expected_error=ValueError,
+        ),
+    ],
+)
```
tariq-hasan, line 445: Let's also add cases for `polling_interval <= 0`,  `polling_interval >= timeout` and `timeout=0`.

And cases for default and custom timeout as well.

The ideal test case matrix would cover
- default status invocation
- default timeout invocation
- default polling interval
- explicit polling interval
- single status {RUNNING}
- multiple statuses {COMPLETED, FAILED}
- invalid timeout < 0
- timeout = 0
- polling_interval <= 0
- polling_interval >= timeout
- successful wait
- backend timeout

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+def test_wait_for_job_status(test_case: TestCase):
+    """Test wait_for_job_status delegation and validation."""
```
tariq-hasan, line 447: For invalid inputs let's add something along the lines of

```
with pytest.raises(ValueError):
    client.wait_for_job_status(...)

backend.wait_for_job_status.assert_not_called()
```

The idea is that validation must occur before delegation to the polling function and this invariant needs to be asserted in unit tests.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+                name=test_case.config["job_name"],
+                follow=test_case.config["follow"],
+            )
+        )
+
+        assert logs == test_case.expected_output
+        backend.get_job_logs.assert_called_once_with(
+            name=test_case.config["job_name"],
+            follow=test_case.config["follow"],
+        )
```
tariq-hasan: Let's print test name throughout all the tests. You can check [kubeflow/spark/backends/kubernetes/backend_test.py](https://github.com/kubeflow/sdk/blob/main/kubeflow/spark/backends/kubernetes/backend_test.py) for the general pattern.

In addition let's use `expected_output` from the `TestCase` dataclass more comprehensively in the tests.

kubeflow/spark/api/spark_client_test.py
```diff
@@ -173,3 +171,222 @@ def test_submit_job_success(job, options):
...
+            expected_output=["log line 1", "log line 2"],
+        ),
+        TestCase(
+            name="get job logs follow",
+            expected_status=SUCCESS,
+            config={"job_name": "spark-job-123", "follow": True},
+            expected_output=["log stream 1", "log stream 2"],
+        ),
+    ],
+)
```
tariq-hasan, line 512: Let's have a default follow invocation that does not use the follow parameter in its input.

tariq-hasan (changes requested):
Hi @Neha-Nimsatkar! I have added a few comments.

Neha-Nimsatkar:
Hi @tariq-hasan , 
* All suggested changes are in commit 38faf0d:
* Full test matrix added (timeouts, polling intervals, invalid inputs)
* Added validation assertions (assert_not_called)
* Fixed type contracts and mock returns
All CI checks are green and comments are resolved. Could you please take a look when you have time? 
Thankyou ...

PR #751, opened 2026-08-10
fix(spark): avoid mutable default argument in wait_for_job_status

Neha-Nimsatkar:
Fixes a mutable default parameter (`status: set = {SparkJobStatus.COMPLETED}`) in `KubernetesBackend.wait_for_job_status` by setting the default to `None` and instantiating the set inside the method body.

In Python, default arguments are evaluated at function definition time rather than execution time. Using a mutable `set` as a default parameter means all calls to `wait_for_job_status()` without an explicit `status` argument share the same `set` reference in memory. If `status` is ever mutated, those changes leak across subsequent calls. Updating the signature to `status: set[SparkJobStatus] | None = None` and initializing `{SparkJobStatus.COMPLETED}` inside the method ensures a fresh set is created for every call.

Fixes #

**Checklist:**

### DIFF
--- a/kubeflow/spark/backends/kubernetes/backend.py
+++ b/kubeflow/spark/backends/kubernetes/backend.py
@@ -1130,7 +1130,7 @@ class KubernetesBackend(RuntimeBackend):
     def wait_for_job_status(
         self,
         name: str,
-        status: set[SparkJobStatus] = {SparkJobStatus.COMPLETED},
+        status: set[SparkJobStatus] | None = None,  # Change from {SparkJobStatus.COMPLETED} to None
         timeout: int = 600,
         polling_interval: int = 2,
     ) -> SparkJob:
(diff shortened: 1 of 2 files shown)


Goku2099:
Nice catch on the mutable default. One thing though — SparkClient.wait_for_job_status in spark_client.py still has the same {SparkJobStatus.COMPLETED} default and always forwards status=status to the backend, so the backend fix alone doesn't actually change anything for callers going through the public API — the hazard's one layer up.

Goku2099:
@Neha-Nimsatkar Can we apply the same None sentinel fix across all three: spark_client.py, base.py's abstract method, and backend.py? Keeps the interface consistent top to bottom instead of just patching the one layer.

andreyvelich (commented):
We don't need this as we discussed before for TrainerClient: https://github.com/kubeflow/sdk/pull/320#pullrequestreview-3849789318
/close

PR #752, opened 2026-08-10
fix(spark): reject bool type as CPU in _validate_cpu_value

Neha-Nimsatkar:
### What this PR does / why we need it:
- Scopes CPU validation in `_validate_cpu_value` strictly to rejecting `bool` types (`True`/`False`), which previously passed as 1 core due to Python's `isinstance(True, int)` subclassing behavior.
- Removes `float` CPU support and reverts resource dictionary type hints to align with backend `SparkV1alpha1ServerSpec` core allocation semantics.
- Raises `TypeError` when an unsupported argument type (e.g. `bool`, `list`, `dict`) is passed.
- Updates unit tests in `utils_test.py` and `types_test.py`.

### Which issue(s) this PR fixes:
N/A

### Checklist:
- [x] Pre-commit checks passed
- [x] Unit tests passed
- [x] DCO Signed-off


tariq-hasan:
Hi @Neha-Nimsatkar! Thanks for raising the PR.

You'd need to fix the errors in pre-commit to pass the CI.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -278,13 +282,21 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
+                cores = int(milli_cpu) / 1000
+            except ValueError:
+                raise ValueError(f"Invalid CPU milli-value: {cpu!r}")
 
         else:
-            cores = float(cpu)
+            try:
+                cores = float(cpu)
+            except ValueError:
+                raise ValueError(f"Invalid CPU string value: {cpu!r}")
```
tariq-hasan: Do we need these new try-except blocks?

Neha-Nimsatkar: Thanks @tariq-hasan! You're right—those try-except blocks were redundant. I've removed them, cleaned up the implementation, and verified that all pre-commit formatting and unit tests pass.

HarshPopat23 (changes requested):
I reviewed the latest head (`d596f10`). Rejecting `bool` is a valid correctness fix because `bool` subclasses `int` in Python, so the current implementation incorrectly accepts `True` as one CPU core.

I still see three points that should be addressed before merging:

### Blocking 1: Public API types do not support the new float input

The private helper now accepts `float`, but the user-facing APIs still declare resource dictionaries as `dict[str, str]`. This includes `SparkClient.connect()`, `SparkClient.submit_job()`, `Driver.resources`, `Executor.resources_per_executor`, and the relevant resolver/builder methods.

As a result, `resources_per_executor={"cpu": 1.5}` works dynamically but is rejected by static type checkers and IDEs. Could we either update the public resource-value type annotations and documentation to make numeric CPU values an official part of the API, or keep the public contract string-based and scope this PR to the boolean-validation fix?

### Blocking 2: Please clarify the float semantics

`1.5` is normalized with `math.ceil()` and becomes `2`. The generated CR therefore receives two Spark cores; it does not preserve a fractional 1.5-CPU Kubernetes request.

The Spark Operator distinguishes the integer `cores` field, which maps to `spark.executor.cores`, from the string `coreRequest` field used for physical Kubernetes CPU requests. The current title and description could therefore be interpreted as fractional CPU-request support even though this change only accepts a float representation and rounds it to an integer core count.

Could we document that fractional values are rounded upward? If the intention is to support fractional Kubernetes CPU allocation, this would need to target `coreRequest` instead.

### Blocking 3: Propagate the new exception contract

Changing unsupported types from `ValueError` to `TypeError` is semantically reasonable, but that exception propagates through public methods. The resolver, spec-builder, and `SparkClient.connect()` docstrings still document only `ValueError` for invalid resource configuration.

Could we update the relevant `Raises` sections so users can reliably handle this new exception type?

Please also add at least one test through a public construction path, such as `get_spark_connect_executor_spec(resources_per_executor={"cpu": 1.5})`, and assert that the resulting CR field is `cores == 2`. A similar driver-resource test would cover the advanced API path.

The explicit boolean rejection itself looks correct.

Neha-Nimsatkar:
Thanks @HarshPopat23 for the thorough review!

I have updated the PR to address all three points:
1. **Public API Types:** Updated `resources_per_executor` type annotations to `dict[str, str | int | float]` across the relevant helper and builder functions.
2. **Float Semantics:** Documented in `_validate_cpu_value` that fractional CPU values are rounded upward (`math.ceil`) for Spark executor cores.
3. **Exception Contract & Tests:** Documented `TypeError` in the relevant `Raises` sections and added public construction path unit tests in `utils_test.py` (`test_get_spark_connect_executor_spec_float_cpu` and `test_get_spark_connect_executor_spec_bool_cpu_raises_type_error`).

All unit tests, Ruff linter checks, and pre-commit hooks are passing cleanly. Please let me know if any further tweaks are needed!

Goku2099:
Non-blocking nit: `dict[str, str | int | float] | None` is now repeated across multiple signatures in this file. Might be worth pulling it into a shared alias in constants.py (e.g. `ResourceDict = dict[str, str | int | float]`) so future changes to accepted types only need one edit. Could be a fast-follow rather than blocking this PR.. @tariq-hasan thoughts?

Neha-Nimsatkar:
Thanks @Goku2099! Extracting ResourceDict = dict[str, str | int | float] into constants.py makes total sense to clean up duplicate signatures across the file. I'll be happy to open a quick fast-follow PR for that right after this one is merged!

Goku2099:
Thanks for the PR!
/lgtm
cc @tariq-hasan

tariq-hasan:
@Neha-Nimsatkar @Goku2099 @HarshPopat23 Looking into the PR changes a bit more I think we should keep this PR focused only on rejecting `bool` type as CPU.

The reason is because the CR construction step currently converts resources into an integer number of cores and not directly CPU request and limit.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/backends/kubernetes/utils.py#L466-L470

There is an issue raised in the spark-operator project to support request and limit directly on the `SparkConnect` CRD - similar to `SparkApplication`-  https://github.com/kubeflow/spark-operator/issues/2957. Once this is completed we can revisit how to support non-integer resources directly through CPU request and limit on the SDK - to ensure the semantics of resources on the SDK matches the expectation that this be treated as K8s-native resource allocation instead of Spark-native interpretation of task parallelism.

@Neha-Nimsatkar Please remove the support for float CPU values for now from the PR so we can move forward with the changes.

Neha-Nimsatkar:
> @Neha-Nimsatkar @Goku2099 @HarshPopat23 Looking into the PR changes a bit more I think we should keep this PR focused only on rejecting `bool` type as CPU.
> 
> The reason is because the CR construction step currently converts resources into an integer number of cores and not directly CPU request and limit.
> 
> https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/backends/kubernetes/utils.py#L466-L470
> 
> There is an issue raised in the spark-operator project to support request and limit directly on the `SparkConnect` CRD - similar to `SparkApplication`- [kubeflow/spark-operator#2957](https://github.com/kubeflow/spark-operator/issues/2957). Once this is completed we can revisit how to support non-integer resources directly through CPU request and limit on the SDK - to ensure the semantics of resources on the SDK matches the expectation that this be treated as K8s-native resource allocation instead of Spark-native interpretation of task parallelism.
> 
> @Neha-Nimsatkar Please remove the support for float CPU values for now from the PR so we can move forward with the changes.

Thanks for the feedback @tariq-hasan! That makes complete sense regarding the backend integer cores mapping and the Spark Operator CRD tracking.

I have updated the PR to remove float CPU support and scoped the changes strictly to rejecting bool types (along with updating the unit tests). All linting, pre-commit checks, and test suites are passing locally. Please take a look when you have a chance!

Neha-Nimsatkar:
cc @tariq-hasan Hi Tariq, the `E2E Test (1.34.0, master)` job hit an infrastructure timeout ("self-hosted runner lost communication") while all other versions passed cleanly. Could you please click "Re-run failed jobs" when you get a chance? Thanks!

kubeflow/spark/types/types.py
```diff
@@ -87,7 +87,7 @@ class Driver:
     """
 
     image: str | None = None
-    resources: dict[str, str] | None = None
+    resources: dict[str, str | int] | None = None
```
tariq-hasan: I think best to follow the approach from other clients.

```suggestion
    resources: dict | None = None
```

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/trainer/types/types.py#L61

These should change as well.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/api/spark_client.py#L67

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/api/spark_client.py#L178

kubeflow/spark/types/types.py
```diff
@@ -121,7 +121,7 @@ class Executor:
     """
 
     num_instances: int | None = None
-    resources_per_executor: dict[str, str] | None = None
+    resources_per_executor: dict[str, str | int] | None = None
```
tariq-hasan: Same here.

```suggestion
    resources_per_executor: dict | None = None
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -136,7 +136,7 @@ def _resolve_driver_resources(
 def _resolve_executor_resources(
     executor: Executor | None = None,
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan: This will need the same treatment for int as in `validate_cpu_value`.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -473,7 +483,7 @@ def get_spark_connect_driver_spec(
 def get_spark_connect_executor_spec(
     executor: Executor | None = None,
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -511,7 +521,7 @@ def build_spark_connect_cr(
     namespace: str,
     spark_version: str | None = None,
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -688,7 +698,7 @@ def get_spark_job_driver_spec(
 
 def get_spark_job_executor_spec(
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -802,7 +812,7 @@ def get_spark_application_cr_from_file_job(
     main_file: str,
     arguments: list[str] | None = None,
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -865,7 +875,7 @@ def get_spark_application_cr_from_func_job(
     func: Callable,
     func_args: dict[str, Any] | None = None,
     num_executors: int | None = None,
-    resources_per_executor: dict[str, str] | None = None,
+    resources_per_executor: dict[str, str | int] | None = None,
```
tariq-hasan:
```suggestion
    resources_per_executor: dict | None = None,
```

kubeflow/spark/backends/kubernetes/utils.py
tariq-hasan:
```suggestion
def _validate_cpu_value(cpu: str) -> int:
```

tariq-hasan: This needs to have the same checks as for cpu.

tariq-hasan (changes requested):

Neha-Nimsatkar:
Thanks for the review and guidance, @tariq-hasan!

I have addressed all the feedback and pushed the updates:

* Simplified `resources` and `resources_per_executor` annotations to `dict | None = None` across `types.py` and `utils.py` to match the convention in `kubeflow.trainer`.
* Updated the two occurrences in `spark_client.py` (lines 67 and 178) as well.
* Ran pre-commit hooks and confirmed all Spark unit tests are passing locally.

Ready for another look whenever you have time!

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -279,13 +283,19 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
         else:
-            cores = float(cpu)
+            try:
+                cores = float(cpu)
+            except ValueError as e:
+                raise ValueError(f"Invalid CPU value '{cpu}'.") from e
 
     else:
-        raise ValueError(f"Invalid CPU type '{type(cpu)}'. Expected str or int.")
+        raise TypeError(f"Invalid CPU type '{type(cpu).__name__}'. Expected str or int.")
```
tariq-hasan, line 298: I'm thinking it's best to revert this change as it is unrelated to the original bug related to bool type.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -490,8 +500,8 @@ def get_spark_connect_executor_spec(
         API ExecutorSpec model.
 
     Raises:
-        ValueError:
-            If the configured executor resources are invalid.
+        ValueError: If the configured executor resources are invalid.
+        TypeError: If an unsupported CPU type (such as bool) is passed.
```
tariq-hasan, line 504: Same here.

```suggestion
        ValueError:
            If the configured executor resources are invalid.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -542,8 +552,8 @@ def build_spark_connect_cr(
         SparkConnect CR as typed Pydantic model.
 
     Raises:
-        ValueError:
-            If the provided driver or executor resource configuration is invalid.
+        ValueError: If the provided driver or executor resource configuration is invalid.
+        TypeError: If an unsupported CPU type (such as bool) is passed.
```
tariq-hasan, line 556:
```suggestion
        ValueError:
            If the provided driver or executor resource configuration is invalid.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -114,8 +114,8 @@ def _resolve_driver_resources(
         Tuple of (cores, memory).
 
     Raises:
-        ValueError:
-            If the configured CPU or memory values are invalid.
+        ValueError: If the configured CPU or memory values are invalid.
+        TypeError: If an unsupported CPU type (such as bool) is passed.
```
tariq-hasan, line 118:
```suggestion
        ValueError:
            If the configured CPU or memory values are invalid.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -154,8 +154,8 @@ def _resolve_executor_resources(
         Tuple containing ``(instances, cores, memory)``.
 
     Raises:
-        ValueError:
-            If the configured CPU or memory values are invalid.
+        ValueError: If the configured CPU or memory values are invalid.
+        TypeError: If an unsupported CPU type (such as bool) is passed.
```
tariq-hasan, line 158:
```suggestion
        ValueError:
            If the configured CPU or memory values are invalid.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -257,11 +257,15 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
         Integer CPU core value.
 
     Raises:
-        ValueError: If CPU value is invalid.
+        ValueError: If CPU value is invalid or non-positive.
+        TypeError: If an unsupported type (such as bool) is passed.
```
tariq-hasan, line 261:
```suggestion
        ValueError: If CPU value is invalid.
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -257,11 +257,15 @@ def _validate_cpu_value(cpu: str | int | None) -> int:
...
     """
     if cpu is None:
         raise ValueError("CPU value cannot be None")
 
+    if isinstance(cpu, bool):
+        raise TypeError("Invalid CPU type 'bool'. Expected str or int.")
+
```
tariq-hasan, line 268:
```suggestion
    if isinstance(cpu, bool):
        raise ValueError("Invalid CPU type 'bool'. Expected str or int.")

```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1892,4 +1901,10 @@ def test_get_spark_application_info_from_cr(
         assert job.creation_timestamp == creation_timestamp
         assert job.num_executors == 5
 
+
+def test_get_spark_connect_executor_spec_bool_cpu_raises_type_error():
+    """Verify that boolean CPU values raise TypeError via public construction path."""
+    with pytest.raises(TypeError):
+        get_spark_connect_executor_spec(resources_per_executor={"cpu": True})
+
```
tariq-hasan, line 1909: We can remove this.

```suggestion
```

tariq-hasan (changes requested):

Neha-Nimsatkar:
Hi @tariq-hasan,
I have addressed all the feedback:
* Reverted TypeError back to ValueError across the validation logic and docstrings.
* Removed the extra try/except on float CPU parsing.
* Removed the boolean TypeError test case from utils_test.py.

All threads are resolved and ready for your review whenever you get time. Thank you...

PR #762, opened 2026-08-15
fix(trainer): resolve LocalProcessBackend Windows compatibility issues

### LINKED ISSUE #757: LocalProcessBackend crashes on Windows due to bash PATH resolution conflict (WSL relay vs Git Bash)
### What happened?

When running a training job locally via LocalProcessBackend on Windows, the job crashes with a misleading error unrelated to the actual training code:

The RPC call contains a handle that differs from the declared handle type.
Error code: Bash/Service/0x8007072c

Root Cause

LocalProcessBackend` invokes the training command via subprocess using the bare string "bash":

https://github.com/kubeflow/sdk/blob/main/kubeflow/trainer/backends/localprocess/utils.py (`get_local_train_job_script`):

return "bash", "-c", command

On Windows, `"bash"` resolves to **whichever `bash.exe` is first in PATH**. It is common for Windows systems to have multiple `bash.exe` binaries:

- `C:\Windows\System32\bash.exe` — a WSL relay launcher
- `C:\Program Files\Git\usr\bin\bash.exe` — Git Bash

If the WSL relay is earlier in PATH (a common default) and the WSL distro is not in a valid/running state, the subprocess call fails immediately with the cryptic RPC error above — even if a fully working Git Bash install is present on the same machine.

 Reproduction Steps

1. On Windows, ensure both WSL (in any state) and Git Bash are installed.
2. Confirm PATH order causes `bash` to resolve to the WSL relay:
   (powershell)
   where.exe bash
   # C:\Windows\System32\bash.exe   <- resolves first
   # C:\Program Files\Git\usr\bin\bash.exe
   
3. Follow the official `examples/pytorch/image-classification/mnist.ipynb` notebook using `LocalProcessBackend`:
   ```python
   from kubeflow.trainer import CustomTrainer, TrainerClient, LocalProcessBackendConfig

   client = TrainerClient(backend_config=LocalProcessBackendConfig())
   for runtime in client.list_runtimes():
       if runtime.name == "torch-distributed":
           torch_runtime = runtime
           break

   job_name = client.train(trainer=CustomTrainer(func=train_fn), runtime=torch_runtime)
   for log in client.get_job_logs(job_name, follow=True):
       print(log, end="")
   ```
4. Observe the job fails immediately with:
   ```
   The RPC call contains a handle that differs from the declared handle type.
   Error code: Bash/Service/0x8007072c
   ```

## Verification

Confirmed the two `bash.exe` binaries behave differently when invoked directly:

```powershell
# WSL relay - fails
> bash --version
<3>WSL (10 - Relay) ERROR: CreateProcessCommon:818: execvpe(/bin/bash) failed: No such file or directory

# Git Bash - works
> & "C:\Program Files\Git\usr\bin\bash.exe" --version
GNU bash, version 5.3.9(1)-release (x86_64-pc-cygwin)


## Expected Behavior

`LocalProcessBackend` should resolve a working shell deterministically rather than relying on unqualified PATH lookup — for example, using `shutil.which("bash")` with validation, or falling back through known-good bash locations on Windows.

## Environment

- OS: Windows
- Kubeflow SDK version: 0.4.1
- Backend: LocalProcessBackend
- Tested against: kubeflow/trainer main branch (current with v2.3.0 timeframe)

## Impact

Any Windows user following the official example notebooks with `LocalProcessBackend` who has multiple `bash.exe` installations on PATH (common with Git for Windows installed) will hit this crash, with an error message that gives no indication the actual cause is bash resolution.


LocalProcessBackend should successfully start the training process on Windows when a valid Bash installation is available. If multiple Bash installations are present, it should resolve a usable Bash executable reliably or provide a clear error indicating the required Bash configuration, rather than failing with the misleading Bash/Service/0x8007072c error

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

kingabdulsalamsas111-lgtm:

Resolves both issues discussed in #757:

1. **Bash resolution** — `LocalProcessBackend` called `subprocess.Popen` with the bare string `"bash"`, which on Windows could resolve to a broken WSL relay launcher instead of a working Git Bash install. Now validates that the resolved bash actually runs before using it.

2. **Environment wiping** — `LocalJob` called `subprocess.Popen(..., env=self.env)` with an explicit `env` dict. On Windows, passing any explicit `env=` replaces the entire parent environment instead of extending it, wiping PATH for the training subprocess regardless of the user's system PATH. Now merges `self.env` into `os.environ`.

While verifying these two fixes end-to-end on Windows, I also had to fix two supporting issues that were blocking the training subprocess from starting at all:

3. **Path normalization** — Windows paths (`venv_dir`, `python_bin`) use backslashes, which bash interprets as escape characters, corrupting paths inside the generated bash script. Now normalized to forward slashes.

4. **Scripts/ vs bin/** — Python's `venv` module creates a `Scripts/` folder on Windows, not `bin/` (the Unix convention the code assumed). The activate script and binary paths now detect this dynamically.

**Verification**:

I tested this locally on Windows with `LocalProcessBackend` + the `torch-distributed` runtime, before and after the fix:

- **Before**: the job crashed immediately with a misleading, unrelated error (`RPC call... handle differs...`) caused by bash resolving to a broken WSL relay.
- **After**: I confirmed the backend successfully resolves a working bash, creates and activates the venv, and installs dependencies (`torch`, `torchvision`) end-to-end via pip — verified across multiple runs with full logs.

Fixes #757

**Checklist:**

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -106,6 +106,7 @@ class LocalProcessBackend(RuntimeBackend):
         # create temp dir
         venv_dir = tempfile.mkdtemp(prefix=trainjob_name)
         logger.debug(f"operating in {venv_dir}")
+        venv_dir = venv_dir.replace("\\", "/")
 
         # get local runtime trainer
         runtime.trainer = local_utils.get_local_runtime_trainer(
--- a/kubeflow/trainer/backends/localprocess/constants.py
+++ b/kubeflow/trainer/backends/localprocess/constants.py
@@ -78,8 +78,9 @@ LOCAL_EXEC_JOB_TEMPLATE = textwrap.dedent(
     set -e
     $OS_PYTHON_BIN -m venv --without-pip $PYENV_LOCATION
     echo "Operating inside $PYENV_LOCATION"
-    source $PYENV_LOCATION/bin/activate
-    $PYENV_LOCATION/bin/python -m ensurepip --upgrade --default-pip
+    if [ -d "$PYENV_LOCATION/Scripts" ]; then VENV_BIN="$PYENV_LOCATION/Scripts"; else VENV_BIN="$PYENV_LOCATION/bin"; fi
+    source $VENV_BIN/activate
+    $VENV_BIN/python.exe -m ensurepip --upgrade --default-pip
     $DEPENDENCIES_SCRIPT
     $ENTRYPOINT
     $CLEANUP_SCRIPT
--- a/kubeflow/trainer/backends/localprocess/job.py
+++ b/kubeflow/trainer/backends/localprocess/job.py
@@ -83,7 +83,7 @@ class LocalJob(threading.Thread):
                 text=True,
                 encoding="utf-8",
                 bufsize=1,
-                env=self.env,
+                env={**os.environ, **self.env},
             )
             # set job status
             self._status = constants.TRAINJOB_RUNNING
--- a/kubeflow/trainer/backends/localprocess/utils.py
+++ b/kubeflow/trainer/backends/localprocess/utils.py
@@ -5,6 +5,7 @@ from pathlib import Path
 import re
 import shutil
 from string import Template
+import subprocess
 import textwrap
 from typing import Any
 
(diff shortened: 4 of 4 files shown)


kingabdulsalamsas111-lgtm:
@robert-bell  @andreyvelich  PR #762 is ready for review, with all Windows LocalProcessBackend issues fixed and verified end-to-end.

PR #766, opened 2026-08-19
feat(spark): add get_job_ui_url() to reach a submitted job's Spark UI

### LINKED ISSUE #733: SparkClient never turns on the Spark UI or metrics for a submitted job
Hi, I was digging through the SparkClient code and noticed something.

Spark itself comes with a built in dashboard that shows you what a job is doing while it runs and after it finishes, which steps are done, which one is stuck, how long each part took. It also supports sending out Prometheus metrics through a sidecar. Both of these are just fields on the SparkApplication spec, `sparkUIOptions` and `monitoring`, that the operator already understands and supports.

I checked both places where SparkClient builds a SparkApplication, `get_spark_application_cr_from_file_job` and `get_spark_application_cr_from_func_job` in `kubeflow/spark/backends/kubernetes/utils.py`, and neither field ever gets set. I also checked the public SparkClient API in `spark_client.py` and there is nothing there about the UI or metrics either.

So right now if you submit a job with `submit_job()` and it hangs or runs slow, the only thing you can look at is raw logs through `get_job_logs()`. There is no way to actually see the job, even though Spark already has that dashboard built and ready, this SDK just never connects it.

I would like to work on this. My plan is to add an option to `submit_job()` that turns on the UI (something like `enable_ui`) and wires up `spark_ui_options` with a sensible default, plus a small helper like `get_job_ui_url()` that reuses the same port forward approach already used for Spark Connect in `get_connect_url()`. I think monitoring can be a separate follow up since it is really its own feature (the Prometheus sidecar) and not just a missing wire.

Let me know if this sounds like the right direction before I start on a PR.

RudraDudhat2509:

Fixes #733, full description and root cause are there. tldr the Spark
Operator already creates a UI service for every job by default,
SparkClient just had no way to find or reach it. get_job_ui_url() looks
it up via the sparkoperator.k8s.io/app-name label (verified against the
actual operator source, not guessed) and port forwards to it, same
pattern as the existing get_connect_url().

Fixes #733

**Checklist:**

### DIFF
--- a/docs/source/spark/lifecycle.rst
+++ b/docs/source/spark/lifecycle.rst
@@ -140,6 +140,22 @@ Lifecycle APIs
    exists — if it has been deleted (for example, due to TTL-based cleanup),
    logs may no longer be available.
 
+**View the job's Spark UI:**
+
+.. code-block:: python
+
+   ui_url, port_forward_process = client.get_job_ui_url(job_name)
+   print(f"Spark UI: {ui_url}")
+   # port_forward_process is None when running in-cluster; outside the
+   # cluster it's the kubectl port-forward process backing ui_url, keep a
+   # reference to it for as long as you need the UI reachable.
+
+.. note::
+   The Spark Operator creates a web UI Service for every job by default, so
+   this works without any extra configuration on ``submit_job()``. It only
+   fails if the operator was deployed with ``--enable-ui-service=false``, or
+   the job has already been cleaned up.
+
 **Delete a job:**
 
 .. code-block:: python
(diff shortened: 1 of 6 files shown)


HarshPopat23 (commented):
Thanks @RudraDudhat2509  for adding this-`get_job_ui_url()` is directly relevant to SparkClient observability.

I noticed two correctness concerns:

1. Selecting `services.items[0]` by only the `app-name` label may return the wrong Service, and custom UI labels may remove that label. Could we use `status.driverInfo.webUIServiceName` and `webUIPort`, which Spark Operator already provides?

2. Checking only whether `kubectl port-forward` is still running does not confirm that the UI is reachable. Could we reuse the port-readiness check from `get_connect_url()` and terminate the process on timeout?

These cases should be covered before approval.

cc: @andreyvelich

RudraDudhat2509:
@HarshPopat23 both good catches, fixed in 860c24c. switched to status.driverInfo.webUIServiceName/webUIPort since the operator writes and reads those itself, and reused _wait_for_connect_port() + terminate on timeout like get_connect_url does. 287 tests pass, lmk if anything else

HarshPopat23:
Thanks @RudraDudhat2509 
from my side
/lgtm

PR #769, opened 2026-08-22
fix(trainer): unpack train_func_parameters as kwargs in localprocess backend

SahilKumar75:
What happened?

The LocalProcess backend generates the training function call by embedding train_func_parameters directly into the generated script as a single positional argument:

    func_code = f"{func_code}\n{train_func.__name__}({train_func_parameters})\n"

Since train_func_parameters is a dict, this produces a call like train_func({'lr': 0.01, 'num_epochs': 5}) instead of train_func(lr=0.01, num_epochs=5). Any training function that takes named parameters raises a TypeError as soon as it is run through LocalProcessBackend.

The Kubernetes backend already handles this correctly with double star unpacking:

    func_call = f"{train_func.__name__}(**{train_func_parameters})"

What did you expect to happen?

CustomTrainer func_args should be passed as keyword arguments on the LocalProcess backend too, consistent with the Kubernetes backend.

Fix

Apply the same ** unpacking used by the Kubernetes backend to the LocalProcess backend's generated call. Added kubeflow/trainer/backends/localprocess/utils_test.py (did not previously exist) covering both the parameterized and no argument cases, asserting the generated script is valid, executable Python.

make test-python and make verify pass locally.

### DIFF
--- a/kubeflow/trainer/backends/localprocess/utils.py
+++ b/kubeflow/trainer/backends/localprocess/utils.py
@@ -216,7 +216,7 @@ def get_command_using_train_func(
     if train_func_parameters is None:
         func_code = f"{func_code}\n{train_func.__name__}()\n"
     else:
-        func_code = f"{func_code}\n{train_func.__name__}({train_func_parameters})\n"
+        func_code = f"{func_code}\n{train_func.__name__}(**{train_func_parameters})\n"
 
     with open(func_file, "w") as f:
         f.write(func_code)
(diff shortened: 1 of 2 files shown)


SahilKumar75:
The required checks are passing. Could a Kubeflow approver please review this change and add LGTM if it is ready?

SahilKumar75:
@szaher, when convenient, could you please review this change? The focused tests and required checks are passing, and the PR is ready for LGTM.

PR #770, opened 2026-08-23
fix(trainer): preserve local job status in list_jobs

### LINKED ISSUE #703: LocalProcessBackend.list_jobs() returns TrainJob objects with Unknown status instead of actual job status
### What happened?

While looking through the `LocalProcessBackend` implementation, I noticed that `list_jobs()` and `get_job()` don't return consistent `TrainJob` objects.

`get_job()` calculates the job status using `__get_job_status()` before creating the `TrainJob`, but `list_jobs()` creates the same object without setting the `status` field.

Since `TrainJob.status` defaults to `Unknown`, every job returned by `list_jobs()` reports `Unknown` even if the underlying job has already completed, failed, or is running.

 **What did you expect to happen?**

I expected `list_jobs()` to return the same status information as `get_job()`, since both APIs represent the same `TrainJob`.

 **How to reproduce**

Looking at the implementation:

`get_job()` does:

status = self.__get_job_status(_job)

return types.TrainJob(
    ...
    status=status,
)
```

But `list_jobs()` constructs the object without setting `status`:


types.TrainJob(
    name=_job.name,
    creation_timestamp=_job.created,
    runtime=runtime,
    num_nodes=1,
    steps=[...],
)


I also reproduced this with a small unit test by creating a local job whose step status is `TRAINJOB_COMPLETE`. Calling `list_jobs()` still returns a `TrainJob` with `status == "Unknown"` instead of `"Complete"`.

 **Why this matters**

This makes the behavior inconsistent between `list_jobs()` and `get_job()`. A caller using `list_jobs()` cannot rely on the returned status, even though the backend already has enough information to determine it.



I expected list_jobs() to return the actual status of each TrainJob, consistent with get_job(). If a job has already completed, failed, or is running, list_jobs() should report that status instead of always returning Unknown.

### Environment

Kubernetes version:
```bash
$ kubectl version

```
Kubeflow Trainer version:
```bash

```
Kubeflow Python SDK version:
```bash
$ pip show kubeflow

```

### Impacted by this bug?

SahilKumar75:

`LocalProcessBackend.list_jobs()` currently omits the overall status when constructing each `TrainJob`. As a result, listed jobs report the default `Unknown` status even when they are complete, failed, running, or created.

This change calculates each job status using the same logic as `get_job()` and includes it in the returned `TrainJob`. It also adds a regression test confirming that a completed local job is listed with the `Complete` status.

Fixes #703

**Checklist:**

- [x] [Docs](https://www.kubeflow.org/docs/) are not needed because this corrects existing behavior without changing the public interface

### DIFF
--- a/kubeflow/trainer/backends/localprocess/backend.py
+++ b/kubeflow/trainer/backends/localprocess/backend.py
@@ -152,6 +152,7 @@ class LocalProcessBackend(RuntimeBackend):
         for _job in self.__local_jobs:
             if runtime and _job.runtime.name != runtime.name:
                 continue
+            status = self.__get_job_status(_job)
             result.append(
                 types.TrainJob(
                     name=_job.name,
@@ -162,6 +163,7 @@ class LocalProcessBackend(RuntimeBackend):
                         types.Step(name=s.step_name, pod_name=s.step_name, status=s.job.status)
                         for s in _job.steps
                     ],
+                    status=status,
                 )
             )
         return result
--- a/kubeflow/trainer/backends/localprocess/backend_test.py
+++ b/kubeflow/trainer/backends/localprocess/backend_test.py
@@ -22,7 +22,10 @@ import pytest
 
 from kubeflow.trainer.backends.localprocess.backend import LocalProcessBackend
 from kubeflow.trainer.backends.localprocess.constants import LOCAL_RUNTIME_IMAGE
+from kubeflow.trainer.backends.localprocess.job import LocalJob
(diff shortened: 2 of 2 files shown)


SahilKumar75:
The title has been corrected to the repository format. Could a Kubeflow approver please review this change and add LGTM if it is ready?

SahilKumar75:
@andreyvelich, when convenient, could you please review this change? The title and description are corrected, the PR is mergeable, and it is ready for LGTM.

PR #776, opened 2026-08-30
feat(spark)!: delegate SparkApplication driver service account to the Spark Operator

### LINKED ISSUE #616: KEP-107: Remove `SparkJob` service account defaulting in Spark package

The Spark SDK currently defaults the service account for the `SparkApplication` driver to `spark-operator-spark`. This is required because the native K8s scheduler in Spark requires that driver pod is attached to a  `ServiceAccount` with pod creation permission (so that the driver can create executors). In turn, the `SparkApplication` controller requires that the driver spec contain a `ServiceAccount` with the same permission (which it then passes to Spark via `spark-submit`). The default `spark-operator` Helm installation creates a per-namespace `ServiceAccount` called `spark-operator-spark` with this permission and all the examples CRs in the repo showcase a driver spec with this `ServiceAccount` included.

An issue has been created in the spark-operator repo to enable a fallback mechanism when a default `ServiceAccount` is not passed in the driver spec for `SparkApplication`: https://github.com/kubeflow/spark-operator/issues/3049. Once that PR lands the SDK can remove the hardcoded `spark-operator-spark` service account from the `constants` module and the associated wiring during the driver spec construction step.

### Why is this needed?

Kubeflow SDK is a distinct plane of abstraction decoupled from upstream operator logic. The core API semantics of the SDK layer are primarily focused on data science use cases as opposed to cluster admin use cases.

### Love this feature?

adibmbrk:

Removes the hardcoded `spark-operator-spark` default service account for the `SparkApplication` driver spec. The Spark Operator now supports a fallback service account mechanism for the driver ([kubeflow/spark-operator#3092](https://github.com/kubeflow/spark-operator/pull/3092), merged), so the SDK no longer needs to set this default itself. `get_spark_job_driver_spec` now leaves `service_account` unset unless a caller explicitly provides one via `Driver.service_account`, keeping the SDK decoupled from operator cluster-admin concerns.

Fixes #616

**Checklist:**

### DIFF
--- a/docs/source/spark/index.rst
+++ b/docs/source/spark/index.rst
@@ -52,11 +52,28 @@ using ``num_executors`` and ``resources_per_executor``.
 
 .. note::
 
-   Batch job submission relies on the Spark Operator's fallback ServiceAccount
-   (``spark-operator-spark`` with the standard Helm install) having the
-   required SparkApplication RBAC permissions in the target namespace.
-   ``SparkClient`` does not set a ServiceAccount on the driver spec, so this
-   operator-configured fallback is always used for batch jobs.
+   ``SparkClient`` does not set a ServiceAccount on the driver spec. Both
+   interactive sessions and batch jobs need the driver to run as a
+   ServiceAccount with permission to create executor pods in the target
+   namespace. That ServiceAccount is provisioned and selected by the platform,
+   not by the SDK.
+
+   On a Kubeflow Platform install, the Profile controller creates a
+   ``default-editor`` ServiceAccount in every profile namespace and binds it to
+   the ``kubeflow-edit`` ClusterRole, which covers pods and services
+   in-namespace. Set ``controller.defaultServiceAccount=default-editor`` in the
+   Spark Operator Helm chart so the operator falls back to it. On a standalone
+   Spark Operator install, set it to a ServiceAccount that exists in every
+   namespace where you submit jobs — for example the ``<release>-spark``
+   account the chart creates in the namespaces listed under
+   ``spark.jobNamespaces``.
+
+   When ``controller.defaultServiceAccount`` is unset, the driver runs as the
+   namespace's ``default`` ServiceAccount and the job fails with a
+   ``403 Forbidden`` error when the driver tries to create executor pods.
+
+   Interactive sessions can override the ServiceAccount per session with
+   ``Driver(service_account=...)``; batch jobs have no per-job override.
 
 Two Ways to Run Spark
 -----------------------
--- a/docs/source/spark/index.rst
+++ b/docs/source/spark/index.rst
@@ -52,13 +52,11 @@ using ``num_executors`` and ``resources_per_executor``.
 
 .. note::
 
-   Batch job submission requires the ``spark-operator-spark`` ServiceAccount to
-   exist in the target namespace, with the required SparkApplication RBAC
-   permissions bound to it. Otherwise, ``submit_job()`` requests will fail.
-
-   This is a current Spark Operator requirement and is expected to be simplified
-   once `kubeflow/spark-operator#3049 <https://github.com/kubeflow/spark-operator/issues/3049>`_
-   is resolved.
+   Batch job submission relies on the Spark Operator's fallback ServiceAccount
+   (``spark-operator-spark`` with the standard Helm install) having the
+   required SparkApplication RBAC permissions in the target namespace.
+   ``SparkClient`` does not set a ServiceAccount on the driver spec, so this
+   operator-configured fallback is always used for batch jobs.
 
 Two Ways to Run Spark
 -----------------------
--- a/examples/spark/README.md
+++ b/examples/spark/README.md
@@ -35,7 +35,7 @@ Install spark dependencies:
 uv pip install kubeflow[spark]
 ```
 
-The Spark examples run against a Kubernetes cluster with the Spark Operator installed. Batch job submission relies on the Spark Operator's fallback ServiceAccount (`spark-operator-spark` by default with the standard Helm install) having the required SparkApplication RBAC permissions in the target namespace; the SDK does not set a ServiceAccount on the driver spec, so this operator-configured fallback is always used. See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
+The Spark examples run against a Kubernetes cluster with the Spark Operator installed. The SDK does not set a ServiceAccount on the driver spec, so the Spark Operator must be configured with a fallback ServiceAccount that can create executor pods in the target namespace (`controller.defaultServiceAccount` in the Helm chart). See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
 
 ## Running Examples
 
--- a/examples/spark/README.md
+++ b/examples/spark/README.md
@@ -35,7 +35,7 @@ Install spark dependencies:
 uv pip install kubeflow[spark]
 ```
 
-The Spark examples run against a Kubernetes cluster with the Spark Operator installed. Batch job submission requires a `spark-operator-spark` ServiceAccount in the target namespace with the required SparkApplication RBAC permissions. See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
+The Spark examples run against a Kubernetes cluster with the Spark Operator installed. Batch job submission relies on the Spark Operator's fallback ServiceAccount (`spark-operator-spark` by default with the standard Helm install) having the required SparkApplication RBAC permissions in the target namespace; the SDK does not set a ServiceAccount on the driver spec, so this operator-configured fallback is always used. See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
 
 ## Running Examples
 
--- a/kubeflow/spark/backends/kubernetes/utils.py
+++ b/kubeflow/spark/backends/kubernetes/utils.py
@@ -670,8 +670,9 @@ def get_spark_job_driver_spec(
 ) -> models.SparkV1beta2DriverSpec:
     """Build DriverSpec for SparkApplication.
 
-    The service account is intentionally left unset so that the Spark
-    Operator's configured fallback service account is used.
+    If no service account is specified on the driver, it is intentionally
+    left unset so that the Spark Operator's configured fallback service
+    account is used.
 
     Returns:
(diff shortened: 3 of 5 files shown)


kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -682,7 +686,7 @@ def get_spark_job_driver_spec(
     return models.SparkV1beta2DriverSpec(
         cores=cores,
         memory=memory,
-        service_account=constants.DEFAULT_SERVICE_ACCOUNT,
+        service_account=driver.service_account if driver else None,
     )
```
tariq-hasan: `service_account=driver.service_account if driver else None,` should be removed.

```suggestion
    )
```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1256,15 +1261,16 @@ def test_get_spark_job_driver_spec(test_case: TestCase) -> None:
...
+    spec = get_spark_job_driver_spec(driver=driver)
 
     assert test_case.expected_status == SUCCESS
 
     assert spec.cores == constants.DEFAULT_DRIVER_CPU
     assert spec.memory == _memory_kubernetes_to_spark(
         constants.DEFAULT_DRIVER_MEMORY,
     )
-    assert spec.service_account == constants.DEFAULT_SERVICE_ACCOUNT
+    assert spec.service_account == (driver.service_account if driver else None)
```
tariq-hasan:
```suggestion
    assert spec.service_account is None
```

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1256,15 +1261,16 @@ def test_get_spark_job_driver_spec(test_case: TestCase) -> None:
 
     print("Executing test:", test_case.name)
 
-    spec = get_spark_job_driver_spec()
+    driver = test_case.config["driver"]
+    spec = get_spark_job_driver_spec(driver=driver)
```
tariq-hasan: This change is not necessary.

```suggestion
    spec = get_spark_job_driver_spec()
```

docs/source/spark/index.rst
```diff
@@ -52,13 +52,11 @@ using ``num_executors`` and ``resources_per_executor``.
...
-   permissions bound to it. Otherwise, ``submit_job()`` requests will fail.
-
-   This is a current Spark Operator requirement and is expected to be simplified
-   once `kubeflow/spark-operator#3049 <https://github.com/kubeflow/spark-operator/issues/3049>`_
-   is resolved.
+   Batch job submission relies on the Spark Operator's fallback ServiceAccount
+   (``spark-operator-spark`` with the standard Helm install) having the
+   required SparkApplication RBAC permissions in the target namespace.
+   ``SparkClient`` does not set a ServiceAccount on the driver spec, so this
+   operator-configured fallback is always used for batch jobs.
```
tariq-hasan: Please add the following in the docs instead of what is in lines 55 to 59.

_Both interactive workloads and batch job submission require a service account with the necessary RBAC permissions for a driver to create executor pods. The Kubeflow profile controller creates this service account - `default-editor` - in every profile namespace and binds it to `kubeflow-edit`, which covers pods and services in-namespace. The platform controller provisions this resouce on namespace creation, and it's what `controller.defaultServiceAccount` in the Spark Operator Helm chart should be set to in a Kubeflow Platform install. The data scientist inherits profile identity rather than hand-writing a service account._

examples/spark/README.md
```diff
@@ -35,7 +35,7 @@ Install spark dependencies:
 uv pip install kubeflow[spark]
 ```
 
-The Spark examples run against a Kubernetes cluster with the Spark Operator installed. Batch job submission requires a `spark-operator-spark` ServiceAccount in the target namespace with the required SparkApplication RBAC permissions. See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
+The Spark examples run against a Kubernetes cluster with the Spark Operator installed. Batch job submission relies on the Spark Operator's fallback ServiceAccount (`spark-operator-spark` by default with the standard Helm install) having the required SparkApplication RBAC permissions in the target namespace; the SDK does not set a ServiceAccount on the driver spec, so this operator-configured fallback is always used. See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
```
tariq-hasan: Please also update this documentation like the one above.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1245,9 +1245,14 @@ def test_read_pod_logs(test_case: TestCase) -> None:
...
-            name="default spark job driver spec",
+            name="default spark job driver spec leaves service account unset",
             expected_status=SUCCESS,
-            config={},
+            config={"driver": None},
+        ),
+        TestCase(
+            name="spark job driver spec honors service account override",
+            expected_status=SUCCESS,
+            config={"driver": Driver(service_account="custom-sa")},
```
tariq-hasan: This change is not necessary.

tariq-hasan:
/retitle feat(spark)!: delegate SparkApplication driver service account to the Spark Operator

examples/spark/README.md
```diff
@@ -35,7 +35,7 @@ Install spark dependencies:
...
```
tariq-hasan:
```suggestion
The Spark examples run against a Kubernetes cluster with the Spark Operator installed. The SDK does not set a ServiceAccount on the driver spec, so the Spark Operator must be configured with a fallback ServiceAccount that can create executor pods in the target namespace (`controller.defaultServiceAccount` in the Helm chart). See the [Spark SDK docs](https://sdk.kubeflow.org/en/latest/spark/index.html) for prerequisites.
```

docs/source/spark/index.rst
tariq-hasan:
```suggestion
   ``SparkClient`` does not set a ServiceAccount on the driver spec. Both
   interactive sessions and batch jobs need the driver to run as a
   ServiceAccount with permission to create executor pods in the target
   namespace. That ServiceAccount is provisioned and selected by the platform,
   not by the SDK.

   On a Kubeflow Platform install, the Profile controller creates a
   ``default-editor`` ServiceAccount in every profile namespace and binds it to
   the ``kubeflow-edit`` ClusterRole, which covers pods and services
   in-namespace. Set ``controller.defaultServiceAccount=default-editor`` in the
   Spark Operator Helm chart so the operator falls back to it. On a standalone
   Spark Operator install, set it to a ServiceAccount that exists in every
   namespace where you submit jobs — for example the ``<release>-spark``
   account the chart creates in the namespaces listed under
   ``spark.jobNamespaces``.

   When ``controller.defaultServiceAccount`` is unset, the driver runs as the
   namespace's ``default`` ServiceAccount and the job fails with a
   ``403 Forbidden`` error when the driver tries to create executor pods.

   Interactive sessions can override the ServiceAccount per session with
   ``Driver(service_account=...)``; batch jobs have no per-job override.
```

kubeflow/spark/backends/kubernetes/utils_test.py
tariq-hasan: We do not need the second test case.

```suggestion
            config={},
```

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -670,6 +670,10 @@ def get_spark_job_driver_spec(
 ) -> models.SparkV1beta2DriverSpec:
     """Build DriverSpec for SparkApplication.
 
+    The service account is intentionally left unset so that the Spark
+    Operator's fallback service account (configured cluster-wide by the
+    operator installation) is used unless overridden via `driver`.
+
```
tariq-hasan:
```suggestion
    The service account is intentionally left unset so that the Spark
    Operator's configured fallback service account is used.

```

tariq-hasan (commented):
Hi @adibmbrk! I have added a few comments.

kubeflow/spark/backends/kubernetes/utils_test.py
```diff
@@ -1245,7 +1245,7 @@ def test_read_pod_logs(test_case: TestCase) -> None:
     "test_case",
     [
         TestCase(
-            name="default spark job driver spec",
+            name="default spark job driver spec leaves service account unset",
```
Goku2099, line 1248: Looks good for the default case. Can we also add a case that passes `Driver(service_account="something")` and asserts it comes through untouched?

Goku2099 (commented):
Thanks for the PR!

kubeflow/spark/backends/kubernetes/utils_test.py
adibmbrk: Addressed in a739415

adibmbrk:
Hi @tariq-hasan @Goku2099! Thank you for the review. I have addressed all your comments. Let me know if further changes are needed.

kubeflow/spark/backends/kubernetes/utils_test.py
tariq-hasan: Actually we can do without this test case. It will bring back the redundant `driver.service_account` pass-through. Let's remove the last commit.

adibmbrk: Done with [3fa6863](https://github.com/kubeflow/sdk/pull/776/commits/3fa6863d6b4daa58d20fd557bfe0c94b85e5622e)

tariq-hasan:
Thanks!
/lgtm

tariq-hasan:
@andreyvelich I am wondering how we should move this PR forward. This change depends on https://github.com/kubeflow/spark-operator/issues/3049 which has been completed recently.

As there is an upstream dependency that has not been released yet I am thinking if we'd need a release note for the breaking change that would basically only work for the next operator release - v2.6.0.

kubeflow/spark/backends/kubernetes/utils.py
```diff
@@ -670,6 +670,9 @@ def get_spark_job_driver_spec(
 ) -> models.SparkV1beta2DriverSpec:
     """Build DriverSpec for SparkApplication.
 
+    The service account is intentionally left unset so that the Spark
+    Operator's configured fallback service account is used.
+
```
andreyvelich: I don't think we need it.

```suggestion

```

adibmbrk:
Hi @tariq-hasan, the E2E check has failed in this PR, I believe it's a flake, would you be able to trigger the tests once again to confirm.

PR #777, opened 2026-08-30
chore(spark): align e2e harness with upstream Helm chart defaults

### LINKED ISSUE #611: KEP-107: Improve Spark E2E Robustness

The goal of this issue is to make the Spark E2E tests more robust and align closely with upstream artifacts to enable better detection of drifts.

The following is a consolidated list of target end states:
- Remove the vendored `SparkConnect` CRD and pull all the latest upstream artifacts to test Spark SDK against the latest version.
- Remove the `default` service account from the RBAC.
- Enable "cluster watcher" support for both `SparkConnect` and `SparkApplication`.

### Why is this needed?

Test robustness ensures the Spark package works in realistic production settings.

### Love this feature?

adibmbrk:

Makes the Spark E2E harness mirror a default `spark-operator` Helm install so it catches upstream drift instead of masking it:

- Remove the vendored SparkConnect CRD; use the chart's own `hook.upgradeCrd=true` pre-install hook instead.
- Remove the hand-rolled RBAC (`ensure_sparkconnect_rbac`) — it granted permissions beyond what the default chart install provides. Examples now run as subprocesses against the CI runner's kubeconfig instead of in-cluster Jobs, which no longer need bespoke ServiceAccount permissions.
- Drop in-cluster (Job-based) example execution entirely: removes `hack/Dockerfile.spark-e2e-runner`, `test/e2e/spark/run_in_cluster.py`, the `connect_existing_session.py` example, and the `SPARK_E2E_RUN_IN_CLUSTER`/`SPARK_E2E_RUNNER_IMAGE` env vars/CI steps. This mode existed to let Job pods (running as the `default` SA) connect via an in-cluster URL; removing the `default`-SA RBAC (per #611) breaks that path, so rather than provision a new dedicated SA for it, in-cluster execution is removed and all examples now run as subprocesses against the ambient kubeconfig.
- Extend the cluster watcher to poll `SparkApplication` in addition to `SparkConnect` so batch-job test failures get the same diagnostics as session tests.
- Bump the K8s version matrix to `1.32.11`, `1.33.7`, `1.34.3`, `1.35.0` and the `SPARK_OPERATOR_VERSION` fallback to `2.5.0`.

Fixes #611

**Checklist:**


hack/e2e-setup-cluster.sh
```diff
@@ -136,6 +136,7 @@ install_spark_operator() {
         --set image.repository=kubeflow/spark-operator/controller
         --set "image.tag=$SPARK_OPERATOR_IMAGE_TAG"
         --set "spark.jobNamespaces[0]=$NAMESPACE"
+        --set "hook.upgradeCrd=true"
     )
```
tariq-hasan: We should add the `controller.defaultServiceAccount` flag to the args list now that https://github.com/kubeflow/spark-operator/issues/3049 has been completed.

The permissions of this service account should match the permissions of the `default-editor` service account according to the justification laid out in this comment: https://github.com/kubeflow/spark-operator/issues/3049#issuecomment-5381830822.

hack/e2e-setup-cluster.sh
tariq-hasan: For `SPARK_OPERATOR_VERSION` let's use the latest Helm chart version.

We can make a decision for whether we want to test fidelity against an operator release or whether we want to test drift against the latest HEAD of the operator. In this specific case though e2e as a drift test is more beneficial since SDK and operator releases tend to be tightly coupled and operator release in general precedes an SDK release.

.github/workflows/test-spark-examples.yaml
tariq-hasan: We should update the kind version as well.

test/e2e/spark/run_in_cluster.py
```diff
@@ -1,231 +0,0 @@
...
-            namespace,
-            "--ignore-not-found=true",
-            "--wait=false",
-        ],
-        capture_output=True,
-        timeout=30,
-    )
-
-    success = succeeded and not failed
-    return success, logs, job_desc
```
tariq-hasan, line 231: We should not remove `run_in_cluster.py`, `Dockerfile.spark-e2e-runner` and `connect_existing_session.py` as in-cluster testing is important.

.github/workflows/test-spark-examples.yaml
```diff
@@ -48,46 +48,26 @@ jobs:
...
-          SPARK_E2E_RUNNER_IMAGE: spark-e2e-runner:local
         timeout-minutes: 15
 
       - name: Collect logs on failure
         if: failure()
         run: |
           kubectl get pods -n spark-test
           kubectl get sparkconnect -n spark-test 2>/dev/null || true
+          kubectl get sparkapplication -n spark-test 2>/dev/null || true
           kubectl logs -n spark-test -l app.kubernetes.io/component=server --tail=100 || true
```
tariq-hasan: Let's repurpose this workflow file to follow the canonical workflow for Trainer repo.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/.github/workflows/test-e2e.yaml#L1-L82

I don't think we'd want a hand-crafted cluster creation setup for Spark SDK - we should delegate this process to the corresponding upstream spark operator project. The Makefile already supports cluster lifecycle management so let's centralize and reuse that logic instead of client-side ownership for server-side cluster creation process - given that this cluster setup is already part of the spark operator e2e and we'd want to ensure reproducibility on both the client-side and server-side layers as well as ensure consistency between the client-side components.

https://github.com/kubeflow/spark-operator/blob/cfc9cbaa571330df17140f1bc8e6560e566f33e2/Makefile#L332-L359

.github/workflows/test-spark-examples.yaml
tariq-hasan, line 1: Let's repurpose this workflow file to follow the approach taken by the Trainer repo.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/.github/workflows/test-e2e.yaml#L1-L82

I don't think we'd want a hand-crafted cluster creation setup for Spark SDK - we should delegate this process to the corresponding upstream spark operator project. The Makefile already supports cluster lifecycle management so let's centralize and reuse that logic instead of making the SDK own setup for a control plane it doesn't ship - given that this cluster setup is already part of the spark operator e2e and we'd want to ensure reproducibility on both the client-side and server-side layers as well as ensure consistency between the client-side components.

https://github.com/kubeflow/spark-operator/blob/cfc9cbaa571330df17140f1bc8e6560e566f33e2/Makefile#L332-L359

tariq-hasan (commented):
Hi @adibmbrk! Thanks for raising the PR. I have added a few comments.

PR #788, opened 2026-09-07
fix(trainer): don't report Complete when a container status is Unknown

### LINKED ISSUE #787: bug(trainer): container backend reports Complete when one container status is Unknown
### What happened?

Follow-up to #562, agreed there: that PR fixed the empty and all-unknown cases, and the mixed case was left out of its scope.

`aggregate_status_from_containers()` in `kubeflow/trainer/backends/container/utils.py` filters `Unknown` out before checking for completion:

```python
known_statuses = [s for s in container_statuses if s != UNKNOWN]
if known_statuses and all(s == constants.TRAINJOB_COMPLETE for s in known_statuses):
```

So ["Complete", "Unknown"] returns Complete, asserted today by the "complete with unknown is complete" case in utils_test.py.

Unknown is not a "no signal to weigh" value. get_container_status() returns it whenever the adapter raises during inspect, and container_status_to_trainjob_status() also falls through to it for paused/dead/restarting. It means "not determined", so dropping it from the check treats an undetermined node as if it agreed with the others.

On a two-node job where node-0 exited 0 and node-1 can't be inspected, wait_for_job_status(status={"Complete"}) returns on that poll and the caller proceeds as if training finished, node-1's outcome never determined.

`Complete` only when every container is known to be complete. `["Complete", "Unknown"]` should aggregate to `Unknown`, so `wait_for_job_status()` keeps polling until node-1 becomes readable or the timeout fires.

Scoped to `Complete` only, `Unknown` outranks it, and every other status keeps its current precedence. `Failed` and `Running` already short-circuit first, and `["Created", "Unknown"]` stays `Created`, so the only behavior that changes is the one case above.

### Environment

Kubeflow SDK version: main (0.5.0rc0)
Python: 3.12.5
OS: macOS 13.6.7
Backend: container (Docker/Podman local execution)

### Impacted by this bug?

brightyorcerf:

`aggregate_status_from_containers()` filtered `Unknown` out of the status set before checking for completion, so `["Complete", "Unknown"]` aggregated to `Complete`.

`Unknown` isn't an absent signal — `get_container_status()` returns it whenever the adapter raises during inspect, and `container_status_to_trainjob_status()` falls through to it for paused/dead/restarting. Filtering it out treated an undetermined container as if it agreed with the others: on a two-node job where node-0 exited 0 and node-1 couldn't be inspected, `wait_for_job_status(status={"Complete"})` returned on that poll with node-1's outcome never determined.

This checks every status instead. Scoped to `Complete` only — the empty-list guard from #562 is preserved, `Failed` and `Running` still short-circuit above it, and `["Created", "Unknown"]` still returns `Created`. Enumerating every multiset of `{Failed, Running, Complete, Created, Unknown}` up to size 4: 120 combinations unchanged, 6 changed, all of them sets drawn from `{Complete, Unknown}` with at least one of each.

Follow-up to #562 

One note on cross-backend consistency, since this makes the container backend stricter than its siblings: `kubernetes` reads the TrainJob CR's conditions (`kubernetes/backend.py:764-772`) so an unreadable pod never blocks `Complete` there, and `localprocess` falls through to `COMPLETE` for anything unrecognized (`localprocess/backend.py:268-281`). The container backend is the only one that synthesizes job status from raw container inspects, so it's the only one where "couldn't determine" is a real signal rather than an artifact of a controller that already decided. Happy to align the others in a follow-up if you'd prefer uniformity.

Fixes #787

**Verification**:

- `uv run pytest kubeflow/ -q` → 696 passed; `ruff check` / `ruff format --check` clean.
- The new `wait_for_job_status` case is a real regression test, verified both ways: reverting `aggregate_status_from_containers` to its current `main` body fails it, with the fix it passes.
- Not covered locally: no real Docker/Podman daemon, so the multi-node path is exercised through `MockContainerAdapter` only.

**Checklist:**


SahilKumar75:
Thanks for the update. I tested PR #788 locally. All 696 tests passed, including 56 container backend tests, and Ruff checks passed. The status aggregation fix looks correct from my testing.

brightyorcerf:
> Thanks for the update. I tested PR #788 locally. All 696 tests passed, including 56 container backend tests, and Ruff checks passed. The status aggregation fix looks correct from my testing.

appreciate your help!

PR #789, opened 2026-09-07
chore: share embedded Python script generation

### LINKED ISSUE #638: Share embedded function execution script template between Trainer and Spark
## Summary

Trainer and Spark currently maintain separate templates for embedding Python function source code into the container at runtime.

The Spark implementation uses a quoted heredoc delimiter to avoid shell expansion and command substitution, while Trainer uses a similar template with a plain `EOM` delimiter. The overall execution flow is conceptually the same, but the implementations have diverged.

## Proposal

Move the embedded function execution script template into a common module so both Trainer and Spark reuse the same implementation.

The shared helper should:
- Generate the embedded Python execution script.
- Be reusable by both Trainer and Spark.
- Preserve the robustness of the quoted heredoc delimiter (or another agreed-upon implementation).
- Reduce duplicated logic across the two clients.

## Benefits

- Consistent behavior between Trainer and Spark.
- Less duplicated code.
- Easier maintenance for future improvements.

ref- https://github.com/kubeflow/sdk/pull/629#discussion_r3634877066

pritesh0089:

Trainer and Spark currently maintain separate shell templates for embedding Python function source into a file at runtime.

This PR:
- moves the quoted-heredoc script generation into `kubeflow.common`
- reuses the shared helper from both Trainer and Spark
- keeps Trainer-specific entrypoint handling local to Trainer
- keeps Spark-specific init-container and volume handling local to Spark
- moves the heredoc delimiter to a shared constant so Spark's existing collision validation uses the same source of truth

The existing Spark reserved-delimiter collision check remains Spark-specific for now. I asked in #638 whether maintainers would prefer moving that validation into the common helper so Trainer receives the same protection as well.

Validation performed locally:
- full Python test suite: 696 passed
- pre-commit hooks passed on all changed files
- `uv lock --check`
- Ruff lint and format checks
- `ty` type check
- `uv build`

AI assistance disclosure: ChatGPT was used for code exploration, implementation guidance, and test/debugging support. The changes were reviewed, understood, and verified locally by the contributor.

Fixes #638

**Checklist:**


kubeflow/spark/backends/kubernetes/backend.py
```diff
@@ -836,11 +836,12 @@ def _validate_func_job(
...
         if any(
-            line.strip() == constants.FUNC_JOB_SCRIPT_DELIMITER for line in func_source.splitlines()
+            line.strip() == common_constants.EMBEDDED_PYTHON_SCRIPT_DELIMITER
+            for line in func_source.splitlines()
         ):
             raise ValueError(
                 "`job.func` source contains the reserved heredoc delimiter "
-                f"{constants.FUNC_JOB_SCRIPT_DELIMITER!r}, which is not supported."
+                f"{common_constants.EMBEDDED_PYTHON_SCRIPT_DELIMITER!r}, which is not supported."
             )
```
tariq-hasan, line 845: I'm thinking it would be ideal for this validation to be shared between both Spark and Trainer by moving it into `get_embedded_python_script` in `common/utils.py`.

At the same time though `kubeflow/trainer/constants/constants.py` passes in format placeholders `"{func_code}"` and `"{func_file}"` so the validation will not happen for Trainer.

Attempting to make the validation work for Trainer will mean that we need to move command building from the constants module entirely into `get_command_using_train_func`.

@andreyvelich I'm wondering if we should move the validation to the common module and modify Trainer's command building to make it work or if this should be parked as a follow-up issue and we keep this validation Spark-only for now.

kubeflow/common/constants.py
```diff
@@ -20,3 +20,6 @@
 
 # Unknown indicates that the value can't be identified.
 UNKNOWN = "Unknown"
+
+# Delimiter used when embedding Python source code in a shell heredoc.
+EMBEDDED_PYTHON_SCRIPT_DELIMITER = "__KUBEFLOW_FUNC_JOB_SCRIPT__"
```
tariq-hasan: `__KUBEFLOW_FUNC_JOB_SCRIPT__` is Spark-specific vocabulary. Trainer does not have a notion of file job and func job. Better to use `__KUBEFLOW_EMBEDDED_SCRIPT__` as the common heredoc delimiter.

pritesh0089: Updated the shared delimiter to __KUBEFLOW_EMBEDDED_SCRIPT__ and adjusted the affected tests. Thanks!

PR #796, opened 2026-09-12
feat(spark): convert Spark batch job lifecycle examples to Jupyter notebooks

### LINKED ISSUE #635: KEP-107: Convert Spark SDK Python examples to Jupyter notebooks
## Summary

Convert the Spark SDK examples from standalone Python scripts into Jupyter notebooks, similar to the Trainer examples.

## Motivation

Many Spark users work primarily in Jupyter notebooks, especially within Kubeflow Notebooks. Notebook-based examples provide a more interactive and reproducible experience, allowing users to execute each step incrementally while following the workflow.

This would also align the Spark SDK examples with the Trainer examples and provide a foundation for future Kubeflow Notebooks v2 documentation.

## Proposed work

- Convert existing Spark SDK examples into Jupyter notebooks.
- Preserve the existing example workflows while organizing them into executable notebook cells.
- Include explanatory markdown cells describing each step.
- Ensure the notebooks are reproducible and easy to run in a Kubeflow environment.

## Benefits

- Improved onboarding experience for new users.
- Interactive, step-by-step examples.
- Better alignment with Trainer examples.
- Useful reference material for future Kubeflow Notebooks documentation.

## Context

This idea came from the discussion https://github.com/kubeflow/sdk/pull/629#discussion_r3634829461, where notebook-based examples similar to the Trainer examples were suggested as a better user experience.

### LINKED ISSUE #801: [Spark] SparkConnect session hangs on remote query execution during headless CI notebook run
### Description

While executing Spark example notebooks headlessly via Papermill in PR #796, `demo_existing_sparkconnect.ipynb` consistently hits a `CellTimeoutError` after 600 seconds.

The remote Spark Connect consumer session hangs when executing `test_spark.range(100).count()`, suggesting that the remote SparkConnect server or its executors are not fully ready or accessible from the CI runner host when the remote session connects.

### Environment

- **Workflow:** `.github/workflows/test-spark-examples.yaml`
- **Kubernetes:** KinD (`v1.32.0`)
- **Runner Environment:** GitHub Actions `ubuntu-latest`
- **Execution Engine:** Papermill with `ipykernel`

### Failure Logs

```text
nbclient.exceptions.CellTimeoutError: A cell timed out while it was being executed, after 600 seconds.
The message was: Cell execution timed out.
Here is a preview of the cell contents:
-------------------
count = test_spark.range(100).count()
print(f"spark.range(100).count() = {count}")
assert count == 100, f"Expected 100, got {count}"

test_spark.stop()
print("Consumer session disconnected.")
-------------------
```

CI Run: https://github.com/kubeflow/sdk/actions/runs/35433120916/job/105871103840?pr=796

### Context & Steps Already Attempted in PR #796

1. Sanitized `subprocess.run` calls without `shell=True` to eliminate process-level blocking issues.
2. Switched from non-deterministic `sessions[-1]` to unique UUID-based session naming to prevent targeting stale sessions.
3. Added explicit `client.delete_session(...)` teardown calls to clean up prior sessions.
4. Verified that standalone and batch notebooks execute properly until reaching remote Spark Connect session evaluation.

### Possible Causes to Investigate

- SparkConnect server pod/service readiness polling before initiating the remote client session.
- Port-forwarding lifecycle or network connectivity between the GitHub Actions runner host and the KinD-hosted SparkConnect pod.
- Resource constraints on the single-node KinD runner delaying Spark executor scheduling.

The remote SparkConnect session should connect to the running SparkConnect pod/service, execute `test_spark.range(100).count()`, verify the count equals 100, and terminate cleanly without hitting the cell execution timeout. All example notebooks in PR #796 should run headlessly to completion via Papermill in the CI pipeline without manual intervention.

### Environment

Kubernetes version:
KinD v1.32.0 (GitHub Actions CI runner)

Spark Operator image:
ghcr.io/kubeflow/spark-operator/controller:local (built from kubeflow/spark-operator master)

Kubeflow Python SDK version:
Current master branch / PR #796 (`uv sync --extra spark --group notebooks`)

### Impacted by this bug?

Neha-Nimsatkar:
## Description

**TL;DR:** Migrates Spark examples from standalone scripts to Jupyter notebooks, removes the legacy test harness, and fixes a CI timeout caused by out-of-cluster DNS resolution (#801).

Fixes #635
Fixes #801

This PR completes the migration of the Spark SDK examples from standalone Python scripts to interactive Jupyter notebooks (`.ipynb`), aligning with the conventions established in `kubeflow/trainer`. It removes the legacy in-cluster test harness and introduces a host-level Papermill-based E2E CI workflow that headlessly validates the notebooks against a live KinD cluster.

## What changed

**Notebooks added & converted (`examples/spark/`)**

Batch job lifecycle:
- `batch_job_lifecycle.ipynb`
- `batch_func_job_lifecycle.ipynb`
- `batch_job_options.ipynb`
- `batch_failed_job.ipynb`

Spark Connect lifecycle & sessions:
- `connect_existing_session.ipynb`
- `demo_existing_sparkconnect.ipynb`
- `spark_advanced_options.ipynb`
- `spark_connect_simple.ipynb`

**Out-of-cluster connection handling & bug fixes (#801)**
- **Resolved `CellTimeoutError` during host execution:** Migrating from in-cluster jobs to host-level Papermill runner execution caused `build_service_url(info)` to construct internal DNS (`*.svc.cluster.local:15002`), which cannot be resolved from the host runner. Updated `connect_existing_session.ipynb` to check for `KUBERNETES_SERVICE_HOST`:
  - Uses direct internal service DNS when running in-cluster.
  - Falls back to `setup_client.backend.get_connect_url(info)` to leverage built-in localhost port-forwarding when executed on external hosts/CI.
- **SDK attribute fix:** Used the public `setup_client.backend` property to resolve the previous `AttributeError: 'SparkClient' object has no attribute '_backend'`.

**KEP-107 & SDK spec alignment**
- **Session naming options:** Aligned session initialization to pass custom identifiers via `options=[Name(...)]` rather than direct kwargs.
- **Compute specifications:** Restructured `Driver` and `Executor` arguments to pass standard resource dictionaries (`resources={"cpu": "1", "memory": "512m"}`) and `resources_per_executor`.
- **Dataclass field correction:** Replaced deprecated `info.service_url` with `info.service_name`.
- **KinD-friendly resource constraints:** Scaled down advanced options (swapped unavailable GPU limits for CPU/memory, updated node selectors to `kubernetes.io/os: linux`, and removed non-provisioned custom `service_account`) so all sessions schedule and pass reliably on single-node KinD CI environments.

**Security, teardown & review feedback**
- **Subprocess sanitization:** Sanitized `subprocess.run` calls in `demo_existing_sparkconnect.ipynb` by passing arguments as an explicit list without `shell=True`.
- **Deterministic session targeting:** Replaced non-deterministic list indexing (`sessions[-1]`) with unique UUID-based session names (`f"...-{uuid.uuid4().hex[:8]}"`).
- **Resource cleanup:** Added explicit `client.delete_session(...)` teardown calls across all demo blocks in `spark_advanced_options.ipynb`.

**Legacy test harness removal & workflow modernization**
- **Removed legacy test harness:** Completely deleted `test/e2e/spark/test_spark_examples.py`, `run_in_cluster.py`, `cluster_watcher.py`, `test/e2e/spark/README.md`, and `hack/Dockerfile.spark-e2e-runner`.
- **Removed deprecated standalone scripts:** Deleted old script equivalents (`batch_*.py`, `connect_*.py`, `spark_connect_simple.py`, `spark_advanced_options.py`).
- **Retained necessary standalone scripts:** Kept `examples/spark/spark_job.py` (for driver pod execution) and `examples/spark/test_connect_url.py` (URL connection test).
- **Updated documentation:** Updated `docs/source/examples.rst` and `examples/spark/README.md` to reference `.ipynb` notebook paths and cleaned up lingering references to deleted `.py` test files.
- **CI harness:** Added `python3` kernelspec metadata across all notebooks, introduced `hack/e2e-run-notebook.sh`, and updated `.github/workflows/test-spark-examples.yaml` to spin up KinD, execute all 8 notebooks sequentially via Papermill, and export execution artifacts.

## Verification

Formatting and repo checks:
```bash
uv run pre-commit run --all-files
uv run ruff check .
```

SDK and unit tests:
```bash
uv run pytest kubeflow/spark/ -v
```

CI runs:
- `Spark Examples E2E Test / example-validation (1.32.0)` passes completely green across all 8 notebooks.
- All documentation, linting, and matrix E2E tests pass.

## Checklist

- [x] Follows Conventional Commits
- [x] Commits signed off (`git commit -s`) per DCO
- [x] Pre-commit hooks and ruff checks pass locally
- [x] Zero references to deleted `.py` test files remain
- [x] Fixed host-level DNS resolution/timeout (#801)
- [x] E2E notebook harness validated green in CI


examples/spark/demo_existing_sparkconnect.ipynb
```diff
@@ -0,0 +1,212 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(\"Launching SparkConnect session with default configuration...\")\n",
+    "spark = client.connect(timeout=120)\n",
+    "\n",
+    "# Discover the auto-generated session\n",
+    "sessions = client.list_sessions()\n",
+    "default_session_name = sessions[-1].name if sessions else None\n",
```
RobuRishabh, line 74: `sessions[-1]` can select and delete another user’s session because Kubernetes does not guarantee list order. Preserve the exact created session name

examples/spark/spark_advanced_options.ipynb
```diff
@@ -0,0 +1,311 @@
...
+    "            }\n",
+    "        ),\n",
+    "    ],\n",
+    ")\n",
+    "\n",
+    "print(\"Spark session created with labels and annotations.\")\n",
+    "df = spark.range(100)\n",
+    "print(f\"Processed {df.count()} rows\")\n",
+    "\n",
+    "spark.stop()\n",
```
RobuRishabh, line 99: All five examples stop the local client but leave SparkConnect clusters running. Explicitly delete each session after use.

examples/spark/demo_existing_sparkconnect.ipynb
```diff
@@ -0,0 +1,212 @@
...
+    "\n",
+    "namespace = os.environ.get(\"SPARK_TEST_NAMESPACE\", \"spark-test\")\n",
+    "backend_config = KubernetesBackendConfig(namespace=namespace)\n",
+    "client = SparkClient(backend_config=backend_config)\n",
+    "\n",
+    "print(f\"SparkClient initialized for namespace: {namespace}\")\n",
+    "\n",
+    "# Check cluster access via kubectl\n",
+    "res = subprocess.run(\n",
+    "    f\"kubectl get namespace {namespace}\", shell=True, capture_output=True, text=True\n",
```
RobuRishabh, line 46: `SPARK_TEST_NAMESPACE` is interpolated into a `shell=True` command, allowing command injection. Pass arguments as a list without a shell.

Neha-Nimsatkar: Addressed. Removed shell=True and passed command arguments as a list (["kubectl", "get", "namespace", namespace]) to avoid command injection risks.

examples/spark/demo_existing_sparkconnect.ipynb
Neha-Nimsatkar: Addressed. Assigned a unique UUID-based session name upfront and directly queried/deleted that specific session instead of relying on list index ordering.

examples/spark/spark_advanced_options.ipynb
Neha-Nimsatkar: Addressed. Added explicit unique session names and appended client.delete_session(...) across all 5 example blocks to prevent resource leaks on remote clusters.

examples/spark/demo_existing_sparkconnect.ipynb
```diff
@@ -0,0 +1,220 @@
...
+    "Define a custom driver and executor topology using `Driver` and `Executor` objects."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "custom_name = \"my-spark-session\"\n",
```
RobuRishabh: still uses hardcoded custom_name = "my-spark-session" collision risk vs UUID names elsewhere

examples/spark/connect_existing_session.ipynb
```diff
@@ -0,0 +1,155 @@
...
+    "setup_client = SparkClient(backend_config=backend_config)\n",
+    "\n",
+    "print(f\"Deploying Spark Connect server: {session_name}...\")\n",
+    "setup_spark = setup_client.connect(options=[Name(session_name)], timeout=180)\n",
+    "\n",
+    "info = setup_client.get_session(session_name)\n",
+    "if os.environ.get(\"KUBERNETES_SERVICE_HOST\"):\n",
+    "    service_url = build_service_url(info)\n",
+    "else:\n",
+    "    service_url, _ = setup_client.backend.get_connect_url(info)\n",
```
RobuRishabh: Fine for short papermill runs (process dies with the kernel), but keeping/`terminate()` on cleanup would be cleaner

examples/spark/spark_connect_simple.ipynb
```diff
@@ -0,0 +1,189 @@
...
+    "    options=[Name(session_name)],\n",
+    ")\n",
+    "\n",
+    "df = spark.range(100)\n",
+    "print(f\"Generated range with {df.count()} rows across 5 executors\")\n",
+    "df.show(10)\n",
+    "\n",
+    "spark.stop()\n",
+    "client.delete_session(session_name)\n",
+    "print(\"Level 2 session stopped and cleaned up.\")"
```
RobuRishabh: Level 2 still asks for 5×(2 CPU, 4Gi) while advanced options were scaled for KinD, this works today, but higher flake risk than the other notebooks.

docs/source/examples.rst
RobuRishabh, line 117: The Spark Examples table only rewrites the existing connect-session rows from `.py` → `.ipynb`, but this PR also adds batch notebooks (and keeps `demo_existing_sparkconnect.ipynb`) that are already listed in `examples/spark/README.md`.

Please add rows here for:
- `demo_existing_sparkconnect.ipynb`
- `batch_job_lifecycle.ipynb`
- `batch_func_job_lifecycle.ipynb`
- `batch_failed_job.ipynb`
- `batch_job_options.ipynb`

Otherwise the published docs examples page stays out of sync with the README / shipped notebooks.

examples/spark/spark_connect_simple.ipynb
```diff
@@ -0,0 +1,189 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "session_name = f\"my-simple-session-{uuid.uuid4().hex[:8]}\"\n",
+    "\n",
+    "spark = client.connect(\n",
+    "    num_executors=5,\n",
+    "    resources_per_executor={\"cpu\": \"2\", \"memory\": \"4Gi\"},\n",
```
RobuRishabh: Level 2 still asks for 5×(2 CPU, 4Gi) while advanced options were scaled for KinD, this works today, but higher flake risk than the other notebooks.

examples/spark/demo_existing_sparkconnect.ipynb
Neha-Nimsatkar: I updated custom_name to add a random 6-character UUID (f"my-spark-session-{uuid.uuid4().hex[:6]}") so there is no chance of name collision.

examples/spark/connect_existing_session.ipynb
Neha-Nimsatkar: I captured pf_proc and added pf_proc.terminate() in the final cleanup cell, so the port-forward process gets killed properly.

docs/source/examples.rst
Neha-Nimsatkar: I added all 5 missing notebook rows (demo_existing_sparkconnect.ipynb and all 4 batch job notebooks) in the Sphinx table, so docs match the README now....

examples/spark/spark_connect_simple.ipynb
Neha-Nimsatkar: I scaled down Level 2 to 1 executor with 1 CPU and 512m RAM (and also scaled down Level 3), so it stays light and works fine on KinD....

.github/workflows/test-spark-examples.yaml
```diff
@@ -69,21 +70,43 @@ jobs:
...
+            PAPERMILL_TIMEOUT=600 \
+            ./hack/e2e-run-notebook.sh
+          done
         env:
           SPARK_TEST_CLUSTER: spark-test
           SPARK_TEST_NAMESPACE: spark-test
-          SPARK_E2E_DEBUG: "1"
-          SPARK_E2E_RUN_IN_CLUSTER: "1"
-          SPARK_E2E_RUNNER_IMAGE: spark-e2e-runner:local
-        timeout-minutes: 15
```
tariq-hasan: Let's also remove references to `SPARK_E2E_DEBUG` in the Spark SDK.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/backends/kubernetes/backend.py#L564-L565

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/backends/kubernetes/backend.py#L663-L664

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/kubeflow/spark/backends/kubernetes/backend.py#L66-L86

CLAUDE.md
```diff
@@ -1 +0,0 @@
-AGENTS.md
```
tariq-hasan, line 1: This change should be removed. The `CLAUDE.md` symlink should remain as is - unrelated to this PR.

hack/e2e-run-notebook.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+# Copyright 2026 The Kubeflow Authors.
+#
```
tariq-hasan, line 3: This is the new header we want to use in subprojects.

```suggestion
#!/usr/bin/env bash
# Copyright The Kubeflow Authors.
#
```

.github/workflows/test-spark-examples.yaml
tariq-hasan, line 116: Since you've removed `cluster_watcher.py` diagnostics for SparkApplications should be added. Let's remove this portion and them in `e2e-run-notebook.sh` similar to how it is done in the Trainer script: .https://github.com/kubeflow/trainer/blob/44819cff07b4d7363ed3fc0803be611d7206ddb7/hack/e2e-run-notebook.sh#L43-L63

On a longer term basis the goal would be for the dignostics to be refactored as a failure hook and called from the action planned as part of https://github.com/kubeflow/testing/issues/1088. But that is beyond immediate scope. For now let's just add the diagnostics since we are deleting `cluster_watcher.py`.

.github/workflows/test-spark-examples.yaml
```diff
@@ -69,21 +70,43 @@ jobs:
...
         timeout-minutes: 15
 
-      - name: Build and load Spark E2E runner image (in-cluster)
+      - name: Run E2E test for Spark example Notebooks
         run: |
-          docker build -f hack/Dockerfile.spark-e2e-runner -t spark-e2e-runner:local .
-          kind load docker-image spark-e2e-runner:local --name spark-test
-        timeout-minutes: 5
+          mkdir -p ./artifacts/notebooks
+          chmod +x ./hack/e2e-run-notebook.sh
```
tariq-hasan, line 76: Do we need `chmod +x ./hack/e2e-run-notebook.sh`? I think we can remove this. I don't see `.github/workflows/test-e2e.yaml` needing this.

.github/workflows/test-spark-examples.yaml
```diff
@@ -37,10 +37,11 @@ jobs:
...
         with:
           enable-cache: true
-      - name: Install project and test deps
+      - name: Install project, test, and notebook deps
         run: |
-          uv sync --extra spark
-          uv pip install pytest pytest-timeout
+          uv sync --extra spark --group notebooks
+          uv pip install papermill ipykernel
+          echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
```
tariq-hasan, line 44: Let's not add any redundancy. papermill and ipykernel are already in pyproject.toml so `uv pip install papermill ipykernel` is not needed here.

This next bit might border on cosmetic but for consistency let's just replace with the following - aligning with `.github/workflows/test-e2e.yaml`.

```
      - name: Install Python dependencies
        run: |
          echo "Installing Kubeflow SDK from source with the Notebook dependencies"
          make install-dev extras=docker groups="--no-default-groups --group notebooks"

          # SDK's hack/e2e-run-notebook.sh calls `papermill` from PATH
          echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
```

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/.github/workflows/test-e2e.yaml#L40-L46

.github/workflows/test-spark-examples.yaml
```diff
@@ -69,21 +70,43 @@ jobs:
...
+          )
+
+          for nb in "${notebooks[@]}"; do
+            nb_name=$(basename "$nb")
+            echo "Running notebook: $nb"
+            NOTEBOOK_INPUT="$nb" \
+            NOTEBOOK_OUTPUT="./artifacts/notebooks/${nb_name}" \
+            PAPERMILL_TIMEOUT=600 \
+            ./hack/e2e-run-notebook.sh
+          done
```
tariq-hasan, line 96: It is important that you rewrite this portion using the `test-e2e-notebook` Makefile target that you have added in this PR - otherwise the target is not being used.

On a side note the goal is to try to align with `.github/workflows/test-e2e.yaml` as much as we can for now to maintain consistency.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/.github/workflows/test-e2e.yaml#L56-L74

On a longer term basis a lot of the shared infra such as cluster setup will move to kubeflow/testing repo. There're active and planned items in the testing repo for this - https://github.com/kubeflow/testing/issues/1068, https://github.com/kubeflow/testing/issues/1075, https://github.com/kubeflow/testing/issues/1088. And then we would want to consolidate e2e setup for consumer subprojects/repos. But that is beyond immediate scope. For now let's try to maintain consistency with `.github/workflows/test-e2e.yaml` as much as we can.

.github/workflows/test-spark-examples.yaml
```diff
@@ -69,21 +70,43 @@ jobs:
...
-          SPARK_E2E_RUNNER_IMAGE: spark-e2e-runner:local
-        timeout-minutes: 15
+        timeout-minutes: 40
+
+      - name: Upload Notebook Artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: executed-notebooks-${{ matrix.kubernetes-version }}
+          path: ./artifacts/notebooks/*
```
tariq-hasan, line 107: More of a nit but let's rewrite this step as below

```
      - name: Upload Artifacts to GitHub
        uses: actions/upload-artifact@v6
        if: always() # Ensure artifacts are uploaded even if previous steps fail
        with:
          name: ${{ matrix.kubernetes-version }}
          path: ./artifacts/notebooks/* # Path relative to the workspace root
          retention-days: 1 #
```

so we are consistent with `.github/workflows/test-e2e.yaml`

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/.github/workflows/test-e2e.yaml#L76-L82

hack/e2e-run-notebook.sh
```diff
@@ -0,0 +1,44 @@
...
+uv run papermill \
+    "${NOTEBOOK_INPUT}" \
+    "${NOTEBOOK_OUTPUT}" \
+    --kernel python3 \
+    --execution-timeout "${PAPERMILL_TIMEOUT}" \
+    --stdout-file /dev/stdout \
+    --stderr-file /dev/stderr \
+    --log-output
+
+echo "✓ Notebook ${NOTEBOOK_INPUT} passed successfully!"
```
tariq-hasan, line 44: Do we need to deviate from how Trainer script is written?

https://github.com/kubeflow/trainer/blob/44819cff07b4d7363ed3fc0803be611d7206ddb7/hack/e2e-run-notebook.sh#L1-L66

Unless there are Spark-specific reasons to deviate I would strongly suggest to maintain a verbatim copy of the Trainer script in `hack/e2e-run-notebook.sh`. The reason again has to do with the eventual goal of moving logic to run notebook in the testing repo (https://github.com/kubeflow/testing/issues/1088). The less deviations the easier is the migration. It is important to not deviate also for consistency reasons and because the Trainer script has been heavily tested in CI already.

The only portion that would change is the `print_results()` portion. This is where the Spark-specific diagnostic would go in.

hack/e2e-setup-cluster.sh
tariq-hasan, line 324: This is not needed anymore since we are not running the notebooks in in-cluster mode.

```suggestion
```

Two points I need to make here.

First of all not testing notebooks in in-cluster mode is a genuine degradation in current e2e. And also it is in-cluster mode that tests genuine use cases since Kubeflow Notebooks run within a cluster and that's where these SDK clients should live in for e2e purposes. We should keep https://github.com/kubeflow/sdk/issues/635 open to revisit this and complete in-cluster papermill on KinD as a follow-up PR. This PR is only implementing host papermill on KinD.

The second point is that everything in `ensure_sparkconnect_rbac()` will need to go - we will do that as part of https://github.com/kubeflow/sdk/issues/611. The reason all these permissions need to go is because these permissions do not represent the typical use case. If permissions are needed for a driver to create executor they need to align with how Kubeflow Notebooks/Workspaces and Kubeflow Community Distribution would set those permissions otherwise we should use the default Helm chart installation without additional permissions to test the e2e. But this second point is beyond the scope of this PR. We will revisit this in https://github.com/kubeflow/sdk/issues/611. For now let's remove only the highlighted role and rolebinding.

tariq-hasan, line 374: Please remove all E2E_CRD_ONLY references since you deleted the smoke test that uses it. All E2E_CRD_ONLY references in `hack/e2e-setup-cluster.sh` are now redundant and should be removed.

```suggestion
```

In addition please remove the following line in pyproject.toml since it is also redundant in lieu of deleting the smoke test.

https://github.com/kubeflow/sdk/blob/3febcd7377d0cdb322bbb420207c3c26cbf7afd7/pyproject.toml#L153

I would think it is okay to remove the smoke test since it was a one-off test and we have deleted `test/e2e/spark/test_spark_examples.py` anyway - sp it won't make sense to retain the ad-hoc smoke test.

test/e2e/spark/cluster_watcher.py
```diff
@@ -1,127 +0,0 @@
-# Copyright 2025 The Kubeflow Authors.
```
tariq-hasan, line 1: Since you are removing all the modules from `test/e2e/spark` please delete the `test` directory since nothing else is left in there after deletion.

Makefile
```diff
@@ -169,3 +169,10 @@ docs-serve:  ## Build and serve docs with live reload
 	@uv sync --group docs
 	@uv pip install sphinx-autobuild
 	@uv run sphinx-autobuild docs/source docs/_build/html
+
+.PHONY: test-e2e-notebook
+test-e2e-notebook: ## Run E2E test for a Jupyter notebook using Papermill
+	@NOTEBOOK_INPUT=$(NOTEBOOK_INPUT) \
+	 NOTEBOOK_OUTPUT=$(NOTEBOOK_OUTPUT) \
+	 PAPERMILL_TIMEOUT=$(PAPERMILL_TIMEOUT) \
+	 bash ./hack/e2e-run-notebook.sh
```
tariq-hasan, line 178: Let's add default inputs like in Trainer Makefile target otherwise the usage invocation in the `e2e-setup-cluster.sh` script will fail.

https://github.com/kubeflow/trainer/blob/44819cff07b4d7363ed3fc0803be611d7206ddb7/Makefile#L263-L270

tariq-hasan (changes requested):
Hi @Neha-Nimsatkar! A few additional comments from me on the CI harness. Overall lgtm.

The goal is to ensure consistency with the more mature Trainer subproject and to align with the stream of work in the testing repo to consolidate CI harness.

